PR #1, opened 2017-02-21
Added the doze power mode support

pallavigopal:
Added the suport for the doze and doze suspend power modes.
For more details about doze and doze suspend refer the below link
https://developer.android.com/training/monitoring-device-state/doze-standby.html

Jira: IAHWC-21
Test: Not able to test the exact doze path device not entering into the idle mode


public/hwcdefs.h
```diff
@@ -63,5 +63,12 @@ enum class DisplayType : int32_t {
   kHeadless = 3
 };
 
+enum DisplayPowerMode {
+  kOff = 0,
```
kalyankondapally: Add comments as to what these modes mean.

common/display/display.cpp
```diff
@@ -164,8 +164,18 @@ bool Display::GetActiveConfig(uint32_t *config) {
   return true;
 }
 
-bool Display::SetDpmsMode(uint32_t dpms_mode) {
-  return display_queue_->SetDpmsMode(dpms_mode);
```
kalyankondapally: May be try by reverting this commit https://github.com/android-ia/frameworks-base/commit/ed03ff5defcbcb1b07a72a3c6c92a208aab19a1d

pallavigopal: Reverted still the device is not going to the idle state and the screen is still on

public/hwcdefs.h
pallavigopal: I will add the comments

common/display/display.cpp
```diff
@@ -164,8 +164,18 @@ bool Display::GetActiveConfig(uint32_t *config) {
...
+bool Display::SetDpmsMode(uint32_t power_mode) {
+
+  power_mode_ = power_mode;
+  uint32_t dpms_prop = display_queue_->GetDpmsProp();
+
+  if (power_mode == DisplayPowerMode::kOff ||
```
kalyankondapally: We should have this handled in displayqueue itself. When we go into doze mode we should also suspend the display thread. We could cache power_mode_ at this level and dont have to do set property till it has really changed.

public/hwcdefs.h
kalyankondapally: Also, can u run git diff | clang-format-diff-3.5 -p 1 -style=file to fix some indentation issues.

common/display/display.cpp
pallavigopal: Even initially i had it as part of the display queue only.
But the api expectation is such that the transition should happen when the function returns.
From the setPowerMode definition
'Sets the power mode of the given display. The transition must be complete when this function returns.'

kalyankondapally: As we discussed offline, we should move this to displayqueue and do it synchronously there, is mode is off or doze than suspend thread.

pallavigopal:
due to rebase issue closing this pull request

pallavigopal:
Initially it appear to be no changes. Now i can see the latest changes got reflected.
But not able to rebase giving some conflicts.

pallavigopal:
something went wrong

PR #4, opened 2017-02-23
Added the doze power mode support

pallavigopal:
Added the suport for the doze and doze suspend power modes.
For more details about doze and doze suspend refer the below link
https://developer.android.com/training/monitoring-device-state/doze-standby.html

Jira: IAHWC-21
Test: Not able to test the exact doze path device not entering into the idle mode

### DIFF
--- a/common/utils/hwctrace.h
+++ b/common/utils/hwctrace.h
@@ -163,7 +163,7 @@ class TraceFunc {
       "Dumping DisplayPlaneState of Current Composition planes "           \
       "-----------------------------");                                    \
   DUMPTRACE("Frame: %d", frame_);                                          \
-  DUMPTRACE("Total Layers for this Frame: %d", layers.size());             \
+  DUMPTRACE("Total Layers for this Frame: %d", queue_item.layers_.size()); \
   DUMPTRACE("Total Planes in use for this Frame: %d",                      \
             current_composition_planes.size());                            \
   int plane_index = 1;                                                     \
@@ -177,7 +177,7 @@ class TraceFunc {
         DUMPTRACE("Layers Index:");                                        \
         for (size_t primary_index : source_layers) {                       \
           DUMPTRACE("index: %d", primary_index);                           \
-          layers.at(primary_index).Dump();                                 \
+          queue_item.layers_.at(primary_index).Dump();                     \
         }                                                                  \
         break;                                                             \
       case DisplayPlaneState::State::kScanout:                             \
@@ -190,7 +190,7 @@ class TraceFunc {
         DUMPTRACE("Layers Index:");                                        \
         for (size_t overlay_index : source_layers) {                       \
           DUMPTRACE("index: %d", overlay_index);                           \
-          layers.at(overlay_index).Dump();                                 \
+          queue_item.layers_.at(overlay_index).Dump();                     \
         }                                                                  \
         break;                                                             \
       default:                                                             \


pallavigopal:
Incorporated all the review comments

common/display/displayqueue.cpp
```diff
@@ -149,27 +149,35 @@ bool DisplayQueue::ApplyPendingModeset(drmModeAtomicReqPtr property_set) {
...
+    dpms_mode_ = DRM_MODE_DPMS_OFF;
+    Flush();
+    drmModeConnectorSetProperty(gpu_fd_, connector_, dpms_prop_, dpms_mode_);
     Exit();
     return true;
   }
 
-  if (dpms_mode_ == DRM_MODE_DPMS_ON) {
+  if (power_mode == kDozeSuspend) {
+    Flush();
```
pallavigopal: do

public/hwcdefs.h
```diff
@@ -63,5 +63,14 @@ enum class DisplayType : int32_t {
   kHeadless = 3
 };
 
+enum DisplayPowerMode {
+  kOff = 0,         // Display is off
```
pallavigopal: i will do

common/display/displayqueue.cpp
```diff
@@ -316,6 +344,8 @@ void DisplayQueue::HandleExit() {
     return;
   }
 
+  Flush();
```
pallavigopal: same like above

common/display/displayqueue.cpp
```diff
@@ -296,18 +315,27 @@ void DisplayQueue::HandleRoutine() {
     ConditionalSuspend();
     return;
   }
+  display_queue_.unlock();
```
pallavigopal: yes to get the size of the queue_

common/display/displayqueue.cpp
pallavigopal: same as above

common/display/displayqueue.cpp
```diff
@@ -149,27 +149,35 @@ bool DisplayQueue::ApplyPendingModeset(drmModeAtomicReqPtr property_set) {
...
-  if (dpms_mode_ == dpms_mode)
+
+  if (dpms_mode_ == power_mode)
     return true;
 
-  dpms_mode_ = dpms_mode;
-  if (dpms_mode_ == DRM_MODE_DPMS_OFF) {
+  if ((power_mode == kOff) || (power_mode == kDoze)) {
+    dpms_mode_ = DRM_MODE_DPMS_OFF;
+    Flush();
```
pallavigopal: User experience wont be as expected if we skip those queued flips. I hope that is the expected user behavior we get the new flip if there is any geometry change in the incoming layers. please correct me if am wrong

kalyankondapally: The only worry I have is depending on the queue the power off/doze can be delayed a lot (If u think each update takes 16.6 msec and if we have two in queue than its 2* 16.6). If the action needs to be immediate than we need to do it immediately. But I think we need to test this out and see how it feels. Thats why a demo would be great to check out.

common/display/displayqueue.cpp
```diff
@@ -215,8 +223,19 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers) {
   return true;
 }
 
+void DisplayQueue::GetNextQueueItem(DisplayQueueItem& item) {
+  display_queue_.lock();
```
kalyankondapally: No need for the locks here. The callers are already holding the locks. Perhaps u want to separate this change to a separate commit and next commit can just use it. We can merge this cleanup in mean time. GetNextQueueItem being called from HandleUpdateRequest I mean.

pallavigopal: GetNextQueueItem is called from HandleRoutine not from HandleUpdateRequest.

common/display/displayqueue.cpp
pallavigopal: Ideally the size of the queue should be zero. The off/doze/doze suspend will get called only when there is no activity in the current app/the device is in idle mode. Flush is just the ideal way of handling the queue exit just added to cover any corner case for some reason the flip delayed.

PR #9, opened 2017-03-02
gl: remove extra program->Init() call

strassek:
No sense in calling Init twice as it should always give the same result.
Drop the unused call.


kalyankondapally:
Merged thanks

PR #10, opened 2017-03-03
Test: fence logic change; Since each frame are single buffered, it sh…

js0701:
…ould wait for the next frame's fence to be signaled, that is the display are displaying the new frame, then it could continue to draw it buffer;
Separate gl Context of different glRenderers

### DIFF
--- a/tests/apps/jsonlayerstest.cpp
+++ b/tests/apps/jsonlayerstest.cpp
@@ -131,7 +131,8 @@ bool init_gl() {
   return true;
 }
 
-static struct frame frames[2];
+// frames count should be larger than 3 for performance
+static struct frame frames[4];
 
 class HotPlugEventCallback : public hwcomposer::DisplayHotPlugEventCallback {
  public:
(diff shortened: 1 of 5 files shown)


tests/apps/jsonlayerstest.cpp
```diff
@@ -341,6 +342,7 @@ static void fill_hwclayer(hwcomposer::HwcLayer *pHwcLayer,
       pParameter->frame_x, pParameter->frame_y, pParameter->frame_width,
       pParameter->frame_height));
   pHwcLayer->SetNativeHandle(pRenderer->GetNativeBoHandle());
+  pHwcLayer->release_fence.Reset(-1);
 }
 
 static void init_frames(int32_t width, int32_t height) {
```
kalyankondapally: Let have one demo which is single buffered and we can continue looking at how to improve the situation there. For rest lets move to double buffered cases. Blocking on current fence even though is fixing tearing but shows lot of lag :(

PR #13, opened 2017-03-05
Review

kalyankondapally:
@pallavigopal Can you please review thanks.

### DIFF
--- a/common/display/displayqueue.cpp
+++ b/common/display/displayqueue.cpp
@@ -90,17 +90,6 @@ bool DisplayQueue::Initialize(uint32_t width, uint32_t height, uint32_t pipe,
   GetDrmObjectProperty("DPMS", connector_props, &dpms_prop_);
   GetDrmObjectProperty("CRTC_ID", connector_props, &crtc_prop_);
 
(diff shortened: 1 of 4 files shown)


common/display/displayqueue.cpp
```diff
@@ -155,26 +144,28 @@ bool DisplayQueue::ApplyPendingModeset(drmModeAtomicReqPtr property_set) {
...
-    HWCThread::ConditionalSuspend();
-    return true;
-  }
-
-  if (power_mode == kOn) {
-    dpms_mode_ = DRM_MODE_DPMS_ON;
-    needs_modeset_ = true;
-    drmModeConnectorSetProperty(gpu_fd_, connector_, dpms_prop_, dpms_mode_);
+  switch (power_mode) {
+    case kOff:
```
pallavigopal: Flush not needed in off scenario ?

kalyankondapally: Atleast didnt show anything weird in this case and we could go immediately to power off in this case, since we release all resources. When are powered on again, we dont have access to last buffer and application has to provide us with the content again, thats why got rid of flush here.

PR #20, opened 2017-03-13
Add FDHandler to watch for out fences.

rantogno:
This series adds a FDHandler that watches for both the out fence fd and the eventfd used to control the HWCThread.

### DIFF
--- a/common/utils/fdhandler.h
+++ b/common/utils/fdhandler.h
@@ -1,0 +1,49 @@
+/*
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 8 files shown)


common/utils/fdhandler.h
```diff
@@ -0,0 +1,49 @@
...
+
+namespace hwcomposer {
+
+class FDHandler {
+ public:
+  FDHandler();
+  virtual ~FDHandler();
+
+  // Add fd to the list of fds that we care about
+  bool AddFd(int fd);
```
kalyankondapally: can u add documentation like we did for HWCEvent ?

kalyankondapally:
minor cleanup tree here https://github.com/01org/IA-Hardware-Composer/commit/48f2c8f74b896229b87004ae61b033a2ce4cb343  I just didnt yet squash the top commit with previous so u can see the changes. Documentation is still missing if u want to add them along with hwcevent ? I though I added it to hwcevent but seems missing :(

PR #22, opened 2017-03-14
vk: Initial commit of the Vulkan backend

strassek:
Add an optional Vulkan backend to the compositor. For now this should be
considered very experimental as it doesn't yet work with the tests and
doesn't include support for explicit sync. Additionally, it is using an
unregistered extension function vkCreateDmaBufImageINTEL, that will need to
be replaced once a proper extension has been added to the spec.

Building the new backend is as simple as passing '--enable-vulkan' during
configuration.


kalyankondapally:
merged

PR #23, opened 2017-03-15
Confirm Validate display call before Accepting Display Changes.

pyn1:
before AcceptDisplayChanges.

Signed-off-by: Poornima <poornima.y.n@intel.com>

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -258,9 +258,9 @@ void GpuDevice::DisplayManager::HotPlugEventHandler() {
   CTRACE();
   char buffer[1024];
   int ret;
-  bool drm_event = false, hotplug_event = false;
 
   while (true) {
+    bool drm_event = false, hotplug_event = false;
     ret = read(hotplug_fd_.get(), &buffer, sizeof(buffer));
     if (ret == 0) {
       return;


os/android/drmhwctwo.h
```diff
@@ -170,6 +170,8 @@ class DrmHwcTwo : public hwc2_device_t {
     }
 
    private:
+    //True after validateDisplay
+    bool checkValidateDisplay;
```
kalyankondapally: initialize the value explicitly to false.

os/android/drmhwctwo.h
```diff
@@ -171,7 +171,7 @@ class DrmHwcTwo : public hwc2_device_t {
 
    private:
     //True after validateDisplay
-    bool checkValidateDisplay;
+    bool checkValidateDisplay = false;
```
kalyankondapally: Sorry missed this earlier, can  u have this together with rest of private member variables below.

pyn1: okay. Did you mean that the field bool checkValidateDisplay = false should be part of private variables of public hwc2_device_t ? I was thinking since, I'm going to use checkValidateDisplay in only the functions of class HwcDisplay; It is enough to keep the variable in class HWCDisplay.
As I'm also new to C++ programming, I may not have interpreted the code correctly. 
Thanks for the review comments.

pyn1: Sorry.. I got what you meant now ! :)

PR #24, opened 2017-03-16
Add support to use either of GBM or Mini GBM as needed

shobhitka:
I had earlier rebased this on commit - 11a3f4a0d5f418473705e99c6c9bfa33ff83151b

And there was no tearing whats so ever in the cube in dual display scenario. But after latest code pull, either of the display will have a tearing in centre of the cube. 

Anyways with gbm kmscube app works but only for some time after which it crashes. That investigation is pending. If you think this as experimental feature you can merge and I will push incremental fix, else you can wait for fix of the crash. But do provide early review comments.

**update**: With most recent code update, there is no crash seen. Looks like that was inherent problem with fd based changes that were done recently and are fixed now.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -27,8 +27,12 @@ SUBDIRS = \
 MAINTAINERCLEANFILES = ChangeLog INSTALL
 
 AM_CPP_INCLUDES = -I$(top_srcdir) -Ipublic -Icommon/core -Icommon/utils -Icommon/compositor -Icommon/display  -Ios/linux -Icommon/compositor/gl -Itests/common
-AM_CPPFLAGS = -std=c++11 -DUDEV_SUPPORT -DUSE_MINIGBM -fPIC -O2 -D_FORTIFY_SOURCE=2 -fstack-protector-strong -fPIE
+AM_CPPFLAGS = -std=c++11 -DUDEV_SUPPORT -fPIC -O2 -D_FORTIFY_SOURCE=2 -fstack-protector-strong -fPIE
 AM_CPPFLAGS += $(AM_CPP_INCLUDES) $(CWARNFLAGS) $(DRM_CFLAGS) $(DEBUG_CFLAGS) -Wformat -Wformat-security
+
+if !ENABLE_GBM
+AM_CPPFLAGS += -DUSE_MINIGBM
+endif
 
 libhwcomposer_la_LIBADD = \
 	$(DRM_LIBS) \
--- a/os/linux/gbmbufferhandler.cpp
+++ b/os/linux/gbmbufferhandler.cpp
@@ -26,6 +26,10 @@
 #include <hwctrace.h>
 #include <platformdefines.h>
 
+#ifndef USE_MINIGBM
+#include "gbmutils.h"
+#endif
+
 namespace hwcomposer {
 
 // static
@@ -69,12 +73,19 @@ bool GbmBufferHandler::CreateBuffer(uint32_t w, uint32_t h, int format,
   temp->import_data.width = gbm_bo_get_width(bo);
   temp->import_data.height = gbm_bo_get_height(bo);
   temp->import_data.format = gbm_bo_get_format(bo);
+#if USE_MINIGBM
   temp->import_data.fds[0] = gbm_bo_get_plane_fd(bo, 0);
   size_t total_planes = gbm_bo_get_num_planes(bo);
   for (size_t i = 0; i < total_planes; i++) {
     temp->import_data.offsets[i] = gbm_bo_get_plane_offset(bo, i);
     temp->import_data.strides[i] = gbm_bo_get_plane_stride(bo, i);
   }
+#else
+  temp->import_data.fd = gbm_bo_get_fd(bo);
+  size_t total_planes = gbm_bo_get_num_planes(temp->import_data.format);
+  temp->import_data.stride = gbm_bo_get_stride(bo);
+#endif
+
   temp->bo = bo;
   *handle = temp;
 
@@ -84,7 +95,11 @@ bool GbmBufferHandler::CreateBuffer(uint32_t w, uint32_t h, int format,
 bool GbmBufferHandler::DestroyBuffer(HWCNativeHandle handle) {
   if (handle->bo) {
     gbm_bo_destroy(handle->bo);
+#ifdef USE_MINIGBM
     close(handle->import_data.fds[0]);
+#else
+    close(handle->import_data.fd);
+#endif
     delete handle;
     handle = NULL;
   }
@@ -97,6 +112,7 @@ bool GbmBufferHandler::ImportBuffer(HWCNativeHandle handle, HwcBuffer *bo) {
   uint32_t gem_handle = 0;
   bo->format = handle->import_data.format;
   if (!handle->bo) {
+#if USE_MINIGBM
     handle->bo = gbm_bo_import(device_, GBM_BO_IMPORT_FD_PLANAR,
                                &handle->import_data, GBM_BO_USE_SCANOUT);
     if (!handle->bo) {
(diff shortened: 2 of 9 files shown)


tests/Makefile.am
```diff
@@ -26,10 +26,14 @@ bin_PROGRAMS = testlayers
 testlayers_LDFLAGS = \
 	-no-undefined
 
-AM_CPP_INCLUDES = -I$(top_srcdir) -I$(top_srcdir)/public -I../common/core -I../common/utils -I../common/compositor -I../common/display  -I../os/linux -I./common -I./third_party/json-c
-AM_CPPFLAGS = -std=c++11 -DUSE_MINIGBM -fPIC -O2 -D_FORTIFY_SOURCE=2 -fstack-protector-strong -fPIE
+AM_CPP_INCLUDES = -I$(top_srcdir) -I$(top_srcdir)/public -I../common/core -I../common/utils -I../common/compositor -I../common/display  -I../os/linux -I./common -I./third_party/json-c -I/home/skumar40/sandbox/upstream/headers/include
```
kalyankondapally: hardcoded paths

tests/common/layerrenderer.cpp
```diff
@@ -35,18 +41,28 @@ bool LayerRenderer::Init(uint32_t width, uint32_t height, uint32_t format,
...
+#if USE_MINIGBM
   planes_ = gbm_bo_get_num_planes(gbm_bo_);
   for (size_t i = 0; i < planes_; i++) {
     native_handle_.import_data.offsets[i] = gbm_bo_get_plane_offset(gbm_bo_, i);
     native_handle_.import_data.strides[i] = gbm_bo_get_plane_stride(gbm_bo_, i);
     native_handle_.import_data.fds[i] = gbm_bo_fd;
   }
+#else
+  planes_ = hwcomposer::gbm_bo_get_num_planes(gbm_bo_get_format(gbm_bo_));
+	native_handle_.import_data.stride = gbm_bo_get_stride(gbm_bo_);
```
kalyankondapally: fix indentation

tests/apps/jsonlayerstest.cpp
```diff
@@ -396,6 +396,7 @@ static void init_frames(int32_t width, int32_t height) {
         case LAYER_TYPE_GL:
           renderer = new GLCubeLayerRenderer(gbm.dev, false);
           break;
+#ifdef USE_MINIGBM
```
kalyankondapally: hmm why do we need this ? Lets avoid this and video layer will fail in this case which is fine since it tells us the current state of things.

tests/Makefile.am
```diff
@@ -50,8 +54,12 @@ testlayers_SOURCES = \
     ./common/layerrenderer.cpp \
     ./common/gllayerrenderer.cpp \
     ./common/glcubelayerrenderer.cpp \
-    ./common/videolayerrenderer.cpp \
-    ./common/imagelayerrenderer.cpp \
     ./common/esTransform.cpp \
     ./common/layerfromjson.cpp \
     ./apps/jsonlayerstest.cpp
+
+if !ENABLE_GBM
```
kalyankondapally: lets include them always

os/linux/gbmbufferhandler.h
```diff
@@ -25,6 +25,69 @@ namespace hwcomposer {
 
 class GpuDevice;
 
+#ifndef USE_MINIGBM
```
kalyankondapally: Can we move this into "gbmutils.h" ? can have it outside USE_MINIGBM and include it tests only if needed

tests/common/layerrenderer.cpp
```diff
@@ -15,6 +15,12 @@
 */
 
 #include "layerrenderer.h"
+#if !USE_MINIGBM
+#include <drm.h>
```
kalyankondapally: hmm why do we need these includes ?

Makefile.am
```diff
@@ -26,10 +26,14 @@ SUBDIRS = \
```
kalyankondapally: commit summary should be more Add support to use GBM & MiniGBM as needed ? sounds like we are removing minigbm support otherwise

Makefile.am
```diff
@@ -26,10 +26,14 @@ SUBDIRS = \
 
 MAINTAINERCLEANFILES = ChangeLog INSTALL
 
-AM_CPP_INCLUDES = -I$(top_srcdir) -Ipublic -Icommon/core -Icommon/utils -Icommon/compositor -Icommon/display  -Ios/linux -Icommon/compositor/gl -Itests/common
-AM_CPPFLAGS = -std=c++11 -DUSE_GL -DUDEV_SUPPORT -DUSE_MINIGBM -fPIC -O2 -D_FORTIFY_SOURCE=2 -fstack-protector-strong -fPIE
+AM_CPP_INCLUDES = -I$(top_srcdir) -Ipublic -Icommon/core -Icommon/utils -Icommon/compositor -Icommon/display  -Ios/linux -Icommon/compositor/gl -Itests/common -I/home/skumar40/sandbox/upstream/headers/include
```
shobhitka: Yeah, did not install the headers in $WLD. Will fix.

os/linux/gbmbufferhandler.cpp
```diff
@@ -69,12 +70,20 @@ bool GbmBufferHandler::CreateBuffer(uint32_t w, uint32_t h, int format,
...
+  temp->import_data.fds[0] = gbm_bo_get_fd(bo);
   size_t total_planes = gbm_bo_get_num_planes(bo);
+
   for (size_t i = 0; i < total_planes; i++) {
     temp->import_data.offsets[i] = gbm_bo_get_plane_offset(bo, i);
     temp->import_data.strides[i] = gbm_bo_get_plane_stride(bo, i);
   }
+#else
+  temp->import_data.fd = gbm_bo_get_fd(bo);
+	size_t total_planes = gbm_bo_get_num_planes(temp->import_data.format);
```
shobhitka: Yeah will, my vim ts=4 was active where as you have used ts=2

tests/Makefile.am
shobhitka: That would result in making changes in VideoLayer and texture layer to avoid compilation error and I would have to do more #ifdef. So think its better to enable on what is working with GBM and minimise changes else where

tests/common/layerrenderer.cpp
shobhitka: Will check. Should not be needed. Looks like some stale code form initial changes.

tests/common/layerrenderer.cpp
shobhitka: again ts=2 vs ts=4. Do we have a coding guidelines ?

tests/apps/jsonlayerstest.cpp
shobhitka: As explained my thought above, just wanted to enable only what works with GBM today to avoid making changes else which which might not yet work. But soon I will test other renderers and enable them

os/linux/gbmbufferhandler.h
shobhitka: ok

Makefile.am
shobhitka: :) Should be more specififc.

shobhitka:
Will send out an update soon

tests/common/layerrenderer.cpp
kalyankondapally: You can check rest of code base, everything should be aligned and that clang-format thing I mentioned above should do it for you. We need to add that to wiki

shobhitka: Yeah works nicely.

shobhitka:
@kalyankondapally , have a look

os/linux/gbmutils.h
```diff
@@ -0,0 +1,81 @@
+/*
+// Copyright (c) 2016 Intel Corporation
```
kalyankondapally: nitpick 2017 :)

shobhitka: Fixed and pushed

shobhitka:
@kalyankondapally , please merge if all comments are addressed. This works now without any crash.

kalyankondapally:
hmm tests dont compile without enabling gbm support.

shobhitka:
Hmm I think I tested build without gbm. Will work on fixing this.

shobhitka:
A small typo got in (USE_MIGBM) into gllayerrenderer.cpp. Fixed that and also re-based on the latest code.

Should be fine now.

kalyankondapally:
merged

PR #25, opened 2017-03-17
Confirm Validate display call before Accepting Display

pyn1:
Changes.

before AcceptDisplayChanges.

Signed-off-by: Poornima <poornima.y.n@intel.com>


kalyankondapally:
Same as pull request 25 ?

PR #26, opened 2017-03-17
Aia 12 and 28

pyn1:


os/android/drmhwctwo.cpp
```diff
@@ -263,9 +263,10 @@ HWC2::Error DrmHwcTwo::HwcDisplay::GetChangedCompositionTypes(
 
 HWC2::Error DrmHwcTwo::HwcDisplay::GetClientTargetSupport(uint32_t /*width*/,
                                                           uint32_t /*height*/,
-                                                          int32_t /*format*/,
+                                                          int32_t format,
                                                           int32_t dataspace) {
-  if (dataspace != HAL_DATASPACE_UNKNOWN &&
+  if (!(display_->CheckPlaneFormat(format)) &&
```
kalyankondapally: https://source.android.com/devices/halref/hwcomposer2_8h_source.html line 1053, "* This function must return true for a client target with width and height
 1066  * equal to the active display configuration dimensions,
 1067  * HAL_PIXEL_FORMAT_RGBA_8888, and HAL_DATASPACE_UNKNOWN."

We should have this as first check and than the plane format checks, dataspace checks like now.

os/android/drmhwctwo.cpp
```diff
@@ -211,6 +211,11 @@ HWC2::Error DrmHwcTwo::HwcDisplay::RegisterVsyncCallback(
 HWC2::Error DrmHwcTwo::HwcDisplay::AcceptDisplayChanges() {
   supported(__func__);
   uint32_t num_changes = 0;
+  if (!checkValidateDisplay) {
```
kalyankondapally: We should reset this to false in this call, otherwise we check this only the first time and if its set to true this check doesnt add any value for further AcceptDisplayChanges.

PR #27, opened 2017-03-20
Confirm Validate display call before Accepting Display

pyn1:
Changes.

before AcceptDisplayChanges.

### DIFF
--- a/os/android/drmhwctwo.cpp
+++ b/os/android/drmhwctwo.cpp
@@ -264,15 +264,20 @@ HWC2::Error DrmHwcTwo::HwcDisplay::GetChangedCompositionTypes(
   return HWC2::Error::None;
 }
 
-HWC2::Error DrmHwcTwo::HwcDisplay::GetClientTargetSupport(uint32_t /*width*/,
-                                                          uint32_t /*height*/,
-                                                          int32_t /*format*/,
+HWC2::Error DrmHwcTwo::HwcDisplay::GetClientTargetSupport(uint32_t width,
+                                                          uint32_t height,
+                                                          int32_t format,
                                                           int32_t dataspace) {
-  if (dataspace != HAL_DATASPACE_UNKNOWN &&
-      dataspace != HAL_DATASPACE_STANDARD_UNSPECIFIED)
+  if (width == display_->Width() && height == display_->Height() &&
+      format == HAL_PIXEL_FORMAT_RGBA_8888 &&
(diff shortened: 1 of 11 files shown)


public/nativedisplay.h
```diff
@@ -82,6 +82,7 @@ class NativeDisplay {
   virtual void SetOutputBuffer(HWCNativeHandle /*buffer*/,
                                int32_t /*acquire_fence*/) {
   }
+  virtual bool CheckPlaneFormat(uint32_t format) = 0;
```
kalyankondapally: Can u add comments to what is expected from this function.

os/android/drmhwctwo.cpp
```diff
@@ -264,15 +264,20 @@ HWC2::Error DrmHwcTwo::HwcDisplay::GetChangedCompositionTypes(
...
+                                                          int32_t format,
                                                           int32_t dataspace) {
-  if (dataspace != HAL_DATASPACE_UNKNOWN &&
-      dataspace != HAL_DATASPACE_STANDARD_UNSPECIFIED)
+  if (width == display_->Width() && height == display_->Height() &&
+      format == HAL_PIXEL_FORMAT_RGBA_8888 &&
+      dataspace == HAL_DATASPACE_UNKNOWN) {
+      return HWC2::Error::None;
+    }
+  else if (!(display_->CheckPlaneFormat(format)) &&
```
kalyankondapally: The check here should be ||, if format is unsupported or dataspace is not either of them. Also, this can be separate if since instead of else.
You can run git diff | clang-format-diff-3.5 -p 1 -style=file before committing, so code gets indented properly.

kalyankondapally: Also, note that you are passing Android specific format, while display expects drm format, you need to convert the Android specific formats for drm_fourcc here before passing it on. You can have something like drm_utils.h in os/android/ which does the conversion for u.

os/android/drmhwctwo.cpp
```diff
@@ -211,8 +211,16 @@ HWC2::Error DrmHwcTwo::HwcDisplay::RegisterVsyncCallback(
 HWC2::Error DrmHwcTwo::HwcDisplay::AcceptDisplayChanges() {
   supported(__func__);
   uint32_t num_changes = 0;
+  if (!checkValidateDisplay) {
```
kalyankondapally: thanks merged

PR #28, opened 2017-03-21
Hotplug event thread should resume itself

js0701:
Hotplug issues
1) Thread is always waiting for signal
2) More than one encoder could be used for one connector
3) Empty displays does not get informed

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -413,7 +413,7 @@ bool GpuDevice::DisplayManager::UpdateDisplayState() {
     headless_.release();
   }
 
-  if (callback_ && !connected_displays_.empty()) {
+  if (callback_) {
     callback_->Callback(connected_displays_);
   }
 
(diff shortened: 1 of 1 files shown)


common/core/gpudevice.cpp
```diff
@@ -313,6 +315,8 @@ void GpuDevice::DisplayManager::HandleRoutine() {
       HotPlugEventHandler();
     }
   }
+
+  Resume();
```
kalyankondapally: Instead of this u can do something like this: https://github.com/01org/IA-Hardware-Composer/blob/master/common/display/pageflipeventhandler.cpp#L115 so u dont have to read/write fds unnecessarily

js0701: OK, That is good. I did not realize the wait function is virtual

common/core/gpudevice.cpp
```diff
@@ -71,6 +71,7 @@ class GpuDevice::DisplayManager : public HWCThread {
       std::shared_ptr<DisplayHotPlugEventCallback> callback);
 
  protected:
+  void HandleWait() override;
```
kalyankondapally: Can you add details like what was the problem and in Tests, what should be fixed now ? in u commit message

PR #29, opened 2017-03-21
Review2

kalyankondapally:


kalyankondapally:
@pallavigopal @txu4 @js0701 Can u please review.

js0701:
so sync_wait is deprecated? if so, we should also use poll in test applications

kalyankondapally:
@js0701 its not deprecated, just wanting to get rid of NativeSync completely when possible.

common/display/displayqueue.h
```diff
@@ -90,13 +88,12 @@ class DisplayQueue : public HWCThread {
   bool needs_modeset_ = false;
   std::unique_ptr<PageFlipEventHandler> flip_handler_;
   std::unique_ptr<DisplayPlaneManager> display_plane_manager_;
-  std::unique_ptr<NativeSync> current_sync_;
   SpinLock spin_lock_;
   SpinLock display_queue_;
   std::queue<DisplayQueueItem> queue_;
+  std::vector<OverlayLayer> free_layers_;
```
kalyankondapally: This shouldnt be needed will remove it before pushing

common/display/displayqueue.cpp
```diff
@@ -274,33 +275,26 @@ void DisplayQueue::HandleUpdateRequest(DisplayQueueItem& queue_item) {
...
+  display_plane_manager_->EndFrameUpdate();
 #ifdef DISABLE_EXPLICIT_SYNC
   compositor_.InsertFence(fence);
 #else
   if (fence > 0) {
-    compositor_.InsertFence(dup(fence));
-    fd_handler_.AddFd(fence);
-    out_fence_.Reset(fence);
+    if (previous_layers_.size() > 0) {
+      HWCPoll(fence, -1);
```
pallavigopal: With poll the flip becomes blocking now.
This will reduce the performance and 16ms delay added for every flip.
Instead we should add this release logic based on the page flip event notified from the kernel in the event thread

common/display/displayqueue.cpp
```diff
@@ -198,8 +192,7 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers) {
     overlay_layer.SetIndex(layer_index);
     overlay_layer.SetAcquireFence(layer->acquire_fence.Release());
     queue_item.layers_rects_.emplace_back(layer->GetDisplayFrame());
-    int ret = layer->release_fence.Reset(
-        queue_item.sync_object_->CreateNextTimelineFence());
+    int ret = layer->release_fence.Reset(overlay_layer.GetReleaseFence());
```
pallavigopal: one more thing here we are resetting the release_fence with the new one. but what is the guarantee that the size of the previous layer and the new set of layers are same. Ideally we should release the previous layers release fence right.

pallavigopal:
Approved the first patch

common/display/displayqueue.cpp
kalyankondapally: These are fences associated with current set of layers. When these set of layers are on screen, we signal the fences of previous set (layers can be different etc doesn't matter any more since we track it for individual layers now)and tied to life time of that layer.

common/display/displayqueue.cpp
txu4: Agree with @pallavigopal it will introduce additional 16ms latency, I think the tearing is due to display release fence, the app can get the previous_layers buffer ahead, but the buffer update should be done after release fence signaled.
Take the Vulkan case for example, the release fence is -1, so tearing occurs.

tpalli:
Are those patches going to eventually land?

PR #30, opened 2017-03-22
Add color correction APIs

yugang:
Add three APIs for gamma, contrast and brightness to hwc display for
color correction. The gamma API accepts float value for RGB colors,
contrast and brightness APIs accept 0 ~ 255 unsigned int value, and
bigger value with stronger contrast/brightness.

Jira: None
Test: call those APIs when show colors in display in HWC display class
      to check if the display hardware change the effect.

Note:
1. applied clang-format
2. add comments which support "doxygen" for three APIs for public using
3. i am working on patch of a auto test (reusing igt test method) to remap gradient color to solid color, will submit later.

@kalyankondapally @js0701 please help to review it firstly, thanks


common/display/display.cpp
```diff
@@ -220,4 +220,39 @@ void Display::VSyncControl(bool enabled) {
   flip_handler_->VSyncControl(enabled);
 }
 
+/**
+* API for setting display color gamma value in HWC, which be used to remap
```
kalyankondapally: should be in header :)

yugang:
Updated & thanks

yugang:
just found changing one of rgb gamma will impact others, but it works normally after execute igt test, i will need time to investigate if the patch has issues or need do something  before set gamma.

kalyankondapally:
Thanks lgtm but need to wait for the test before this can be merged.

yugang:
ok.
about "changing one of rgb gamma would impact others", i have investigated and found a useful info in http://losca.blogspot.com/2013/11/workaround-for-setting-full-rgb-when.html. seems some display devices have limitation for color range(e.g. 16-235), that mean there are no "0" value for rgb colors in those display, that is why the gamma will impact all color in display device even though the impacted color is set to "0" in driver side (actually is 16 in device). 

In igt test, it set "Broadcast RGB" as "Full", so gamma of one color will not impact others which be set as "0" in driver side. @kalyankondapally, @js0701, pls correct me if i wrong. so the possible solution in HWC& tests are:
1. option-1: always set  "Broadcast RGB" as "Full" in HWC
2. open a API in HWC to enable/disable "Broadcast RGB" "Full", so that test cases can do solid remap. and API may be useful for HWC users.

which you prefer?

kalyankondapally:
option2 is fine.

kalyankondapally:
can u rebase on top of latest master TOT ?

yugang:
updated and added broadcast rgb setting API.

tests/Makefile.am
```diff
@@ -21,19 +21,15 @@
...
 
-bin_PROGRAMS = testlayers
+bin_PROGRAMS = testlayers colorcorrection
 
 testlayers_LDFLAGS = \
 	-no-undefined
 
 AM_CPP_INCLUDES = -I$(top_srcdir) -I$(top_srcdir)/public -I../common/core -I../common/utils -I../common/compositor -I../common/display  -I../os/linux -I./common -I./third_party/json-c
-AM_CPPFLAGS = -std=c++11 -fPIC -O2 -D_FORTIFY_SOURCE=2 -fstack-protector-strong -fPIE
+AM_CPPFLAGS = -std=c++11 -DUSE_MINIGBM -fPIC -O2 -D_FORTIFY_SOURCE=2 -fstack-protector-strong -fPIE
```
kalyankondapally: hmm u shoul remove this since we can compile with minigbm or without it. Also below changes related to Minigbm, shouldnt affect default builds

public/nativedisplay.h
```diff
@@ -65,6 +65,7 @@ class NativeDisplay {
 
   virtual bool GetDisplayConfigs(uint32_t *num_configs, uint32_t *configs) = 0;
   virtual bool GetDisplayName(uint32_t *size, char *name) = 0;
+  virtual bool GetDisplayPipe(uint32_t *pipe_id) = 0;
```
kalyankondapally: Can  u add comments for this API, instead of boolean u can return int, if it returns -1 than we dont have display connected.

common/display/display.cpp
```diff
@@ -176,6 +176,11 @@ bool Display::GetDisplayName(uint32_t *size, char *name) {
   return true;
 }
 
+bool Display::GetDisplayPipe(uint32_t *pipe_id) {
```
kalyankondapally: In Test section, can u add what test needs to be run and expected results. Also, can u add a Jira for this feature and use that for Jira here and in u other patch

kalyankondapally:
needs to be rebased again sorry

tests/Makefile.am
yugang: ok, thanks

common/display/display.cpp
yugang: ok, thanks

yugang:
yes, i will rebase it,  actually i just pushed the commit to my github account for backup, the github did show the commit in this pr automatically, i need clean code: e.g. add licenses, comments..., thanks, i will update soon

public/nativedisplay.h
yugang: This api combined with connected displays, the diplay should been "connected" alrady if it was used

yugang:
@kalyankondapally updated done, could you help to take a look again & thanks

yugang:
updated license on source code header

common/display/displayqueue.cpp
```diff
@@ -322,4 +339,218 @@ bool DisplayQueue::CheckPlaneFormat(uint32_t format) {
...
+  drmModeFreeObjectProperties(props);
+  return ret;
+}
+
+bool DisplayQueue::SetBroadcastRGB(const char* range_property) {
+  drmModePropertyPtr p;
+  int32_t p_value = -1;
+  uint32_t p_id;
+
+  if (!GetBroadcastRGBProperty(gpu_fd_, connector_, DRM_MODE_OBJECT_CONNECTOR,
```
kalyankondapally: No need to do this always, u can do it in constructor/initialize.

common/display/displayqueue.cpp
```diff
@@ -322,4 +339,218 @@ bool DisplayQueue::CheckPlaneFormat(uint32_t format) {
...
+  ApplyPendingLUT(lut);
+  free(lut);
+}
+
+bool DisplayQueue::GetBroadcastRGBProperty(int fd, uint32_t id, uint32_t type,
+                                           const char* name, uint32_t* p_id,
+                                           uint64_t* p_value,
+                                           drmModePropertyPtr* p) {
+  drmModePropertyPtr p_t;
+  drmModeObjectPropertiesPtr props;
```
kalyankondapally: You can use existing helper functions GetDrmObjectProperty and GetDrmObjectPropertyvalue instead of this

yugang: Done, pls check second commit & thanks

common/display/displayqueue.cpp
yugang: done, pls check second commit & thanks

yugang:
@kalyankondapally , about test got FAILED and got crc value "unitialized", i gone through code, no places init the "unitialized", maybe got from kernel. 

so possible way to check is: can you run igt test on your machine? likes "intel-gpu-tools# ./tests/kms_pipe_color --run-subtest gamma-pipe*" to check if the test can pass? to got more log, add "printf("%d--%d\n", a->crc[i], b->crc[i]);" to "intel-gpu-tools/lib/igt_debugfs.c" "igt_assert_crc_equal" function, likes:

```
void igt_assert_crc_equal(const igt_crc_t *a, const igt_crc_t *b)
         {
                int i;

                for (i = 0; i < a->n_words; i++){
                        printf("%d--%d\n", a->crc[i], b->crc[i]);
                        igt_assert_eq_u32(a->crc[i], b->crc[i]);
                }
        }
```

if also got same fail/log, maybe your kernel/hardware has some different with mine. if passed, there should some issues when i did porting.

yugang:
move colorcorrection tests to "tests/autotests" folder, named "colorcorrection_autotest"

common/display/displayqueue.cpp
```diff
@@ -490,49 +496,11 @@ void DisplayQueue::SetColorCorrection(struct gamma_colors gamma,
...
-  return ret;
-}
-
 bool DisplayQueue::SetBroadcastRGB(const char* range_property) {
-  drmModePropertyPtr p;
-  int32_t p_value = -1;
-  uint32_t p_id;
+  uint64_t p_value = -1;
+
+  drmModePropertyPtr p = drmModeGetProperty(gpu_fd_, broadcastrgb_id_);
```
kalyankondapally: do we still need this ?

common/display/displayqueue.cpp
```diff
@@ -340,17 +343,20 @@ bool DisplayQueue::CheckPlaneFormat(uint32_t format) {
...
+    const char* name, const ScopedDrmObjectPropertyPtr& props, uint64_t* value,
+    uint32_t* id) const {
   uint32_t count_props = props->count_props;
   for (uint32_t i = 0; i < count_props; i++) {
     ScopedDrmPropertyPtr property(drmModeGetProperty(gpu_fd_, props->props[i]));
     if (property && !strcmp(property->name, name)) {
-      *value = props->prop_values[i];
+      if (value)
+        *value = props->prop_values[i];
+      if (id)
```
kalyankondapally: why do we need these changes ?

common/display/displayqueue.cpp
yugang: yes, i think so, as we need map drm_mode_property_enum's name to value so that we can use name "Full" to set value

common/display/displayqueue.cpp
yugang: updated done & thanks

yugang:
update done

yugang:
moved color correction relevant APIs and test cases to two separate commits to make changes more clear for review

tests/common/igt.cpp
```diff
@@ -0,0 +1,329 @@
...
+  FILE *ctl;
+  size_t written;
+  int ret;
+
+  ctl = igt_debugfs_fopen("crtc-0/crc/control", "r+");
+  if (!ctl) {
+    ctl = igt_debugfs_fopen("i915_display_crc_ctl", "r+");
+    if (ctl)
+      printf("No display_crc_ctl found, kernel too old\n");
+    return false;
```
kalyankondapally: You are missing brackets here (if (ctl) condition), u will always return false here otherwise.

common/display/display.cpp
```diff
@@ -223,4 +229,19 @@ bool Display::CheckPlaneFormat(uint32_t format) {
...
+void Display::SetContrast(uint32_t red, uint32_t green, uint32_t blue) {
+  display_queue_->SetContrast(red, green, blue);
+}
+
+void Display::SetBrightness(uint32_t red, uint32_t green, uint32_t blue) {
+  display_queue_->SetBrightness(red, green, blue);
+}
+
+bool Display::SetBroadcastRGB(const char *range_property) {
+  display_queue_->SetBroadcastRGB(range_property);
```
kalyankondapally: missing return here.

public/nativedisplay.h
```diff
@@ -85,6 +90,43 @@ class NativeDisplay {
...
+  * @param blue valid value is 0 ~ 255, bigger value with stronger brightness
+  */
+  virtual void SetBrightness(uint32_t red, uint32_t green, uint32_t blue) {
+  }
+  /**
+  * API for setting display Broadcast RGB range property
+  * @param range_property supported property string, e.g. "Full", "Automatic"
+  */
+  virtual bool SetBroadcastRGB(const char *range_property) {
+  }
```
kalyankondapally: missing return here i.e. return false;

common/display/displayqueue.cpp
```diff
@@ -322,4 +340,181 @@ bool DisplayQueue::CheckPlaneFormat(uint32_t format) {
...
+  }
+
+  if (p_value == -1)
+    return false;
+
+  if (drmModeObjectSetProperty(gpu_fd_, connector_, DRM_MODE_OBJECT_CONNECTOR,
+                               broadcastrgb_id_, p_value) != 0)
+    return false;
+
+  drmModeFreeProperty(p);
```
kalyankondapally: u need to return true at end

common/display/displayqueue.cpp
```diff
@@ -322,4 +340,181 @@ bool DisplayQueue::CheckPlaneFormat(uint32_t format) {
...
+
+bool DisplayQueue::SetBroadcastRGB(const char* range_property) {
+  uint64_t p_value = -1;
+
+  drmModePropertyPtr p = drmModeGetProperty(gpu_fd_, broadcastrgb_id_);
+
+  if (!(p->flags & DRM_MODE_PROP_ENUM))
+    return false;
+
+  for (int i = 0; i < p->count_enums; i++) {
```
kalyankondapally: hmm is this the correct version of the patch ? we could cache all this during initialize since it shouldn't change after that

common/display/displayqueue.cpp
```diff
@@ -322,4 +340,181 @@ bool DisplayQueue::CheckPlaneFormat(uint32_t format) {
...
+    if (!strcmp(p->enums[i].name, range_property)) {
+      p_value = p->enums[i].value;
+      break;
+    }
+  }
+
+  if (p_value == -1)
+    return false;
+
+  if (drmModeObjectSetProperty(gpu_fd_, connector_, DRM_MODE_OBJECT_CONNECTOR,
```
kalyankondapally: only this needs to happen here

common/display/displayqueue.cpp
```diff
@@ -322,4 +340,181 @@ bool DisplayQueue::CheckPlaneFormat(uint32_t format) {
...
 }
 
+void DisplayQueue::GetDrmObjectPropertyValue(
+    const char* name, const ScopedDrmObjectPropertyPtr& props,
+    uint64_t* value) const {
+  uint32_t count_props = props->count_props;
+  for (uint32_t i = 0; i < count_props; i++) {
+    ScopedDrmPropertyPtr property(drmModeGetProperty(gpu_fd_, props->props[i]));
+    if (property && !strcmp(property->name, name)) {
+      if (value)
```
kalyankondapally: u can remove this check for consistency.

tests/common/igt.cpp
```diff
@@ -0,0 +1,329 @@
...
+    written = fwrite(cmd, 1, strlen(cmd), ctl);
+    ret = fflush(ctl);
+    if (written == strlen(cmd) && ret == 0)
+      printf("CRCs not supported on this platform\n");
+  }
+
+  fclose(ctl);
+}
```
kalyankondapally: missing return here

tests/autotests/colorcorrection_autotest.cpp
```diff
@@ -0,0 +1,424 @@
...
+    if (pipe_id == -1) {  // unconnected display
+      continue;
+    }
+    igt_pipe_crc_t *pipe_crc = pipe_crc_new(pipe_id);
+    igt_crc_t crc_fullcolors;
+    igt_pipe_crc_collect_crc(pipe_crc, &crc_fullcolors);
+    crc_list[i] = crc_fullcolors;
+    free(pipe_crc);
+    i++;
+  }
```
kalyankondapally: missing return value here

tests/autotests/colorcorrection_autotest.cpp
```diff
@@ -0,0 +1,424 @@
...
+  for (int i = 0; i < height; i++) {
+    if (i < height / 3) {
+      c_offiset = 0;  // blue
+    } else if (i >= height / 3 && i < (height * 2) / 3) {
+      c_offiset = 1;  // green
+    } else {
+      c_offiset = 2;  // red
+    }
+    if (gradient) {
+      c_value = 255 * ((float)(i % (height / 3)) / (height / 3));
```
kalyankondapally: we might want to have a check here that height/3 > 0 or modulo result will be undefined. i.e store height / 3 in a temp variable, ensure its > 0 and than do this calculation

tests/common/igt.cpp
yugang: yes, this function is not used actually, i will fix.

common/display/displayqueue.cpp
yugang: @kalyankondapally ,range_property is a dynamic parameter, and SetBroadcastRGB may happens anytime instead of display initialize. so i need map range_property to p_value dynamically. is that your concerns?

tests/autotests/colorcorrection_autotest.cpp
yugang: thanks a lot, i will update soon

common/display/displayqueue.cpp
kalyankondapally: I mean we could get p_value and cache it like we do with broadcastrgb id and just set the value here ?

tests/autotests/colorcorrection_autotest.cpp
yugang: done, changed to void.

PR #32, opened 2017-03-23
Add surface damage support

pallavigopal:
Added the basic support to avoid the complete validate and present
if the given layers are same as previous one.

Jira: IAHWC-36
Test: No regressions with Linux app and on Android

### DIFF
--- a/common/display/displayqueue.cpp
+++ b/common/display/displayqueue.cpp
@@ -208,9 +208,11 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
   size_t size = source_layers.size();
   std::vector<OverlayLayer> layers;
   std::vector<HwcRect<int>> layers_rects;
+  bool validate = true;
 
   for (size_t layer_index = 0; layer_index < size; layer_index++) {
     HwcLayer* layer = source_layers.at(layer_index);
+    const HwcRegion &current_surface_damage = layer->GetSurfaceDamage();
     layers.emplace_back();
     OverlayLayer& overlay_layer = layers.back();
     overlay_layer.SetNativeHandle(layer->GetNativeHandle());
@@ -219,6 +221,7 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
     overlay_layer.SetBlending(layer->GetBlending());
     overlay_layer.SetSourceCrop(layer->GetSourceCrop());
     overlay_layer.SetDisplayFrame(layer->GetDisplayFrame());
+    overlay_layer.SetSurfaceDamage(current_surface_damage);
     overlay_layer.SetIndex(layer_index);
     overlay_layer.SetAcquireFence(layer->acquire_fence.Release());
     layers_rects.emplace_back(layer->GetDisplayFrame());
@@ -228,6 +231,29 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
     int ret = layer->release_fence.Reset(overlay_layer.GetReleaseFence());
     if (ret < 0)
       ETRACE("Failed to create fence for layer, error: %s", PRINTERROR());
+
+    //check the surafce damage if its 0 for all layers then skip validate
+    // and commit
+    if (current_surface_damage.kNumRects == 1) {
+      const HwcRect<int> *rect = current_surface_damage.kRects;
+      if ((rect->top == 0) && (rect->bottom == 0) && (rect->left == 0) &&
+          (rect->right == 0)) {
+        const HwcRect<int> &previous = previous_layers_.at(layer_index).GetDisplayFrame();
+        const HwcRect<int> &current = layer->GetDisplayFrame();
+        if ((previous.left != current.left) || (previous.top != current.top)) {
+          validate = true;
+        } else {
+          validate = false;
+        }
+      }
+    } else if (current_surface_damage.kNumRects == 0) {
+      validate = true;
+    }
+  }
+
+  if (!validate) {
+    buffer_manager_->UnRegisterLayerBuffers(layers);
+    return true;
   }
 
   // Reset any DisplayQueue Manager and Compositor state.
--- a/os/android/drmhwctwo.cpp
+++ b/os/android/drmhwctwo.cpp
@@ -683,8 +683,19 @@ HWC2::Error DrmHwcTwo::HwcLayer::SetLayerSourceCrop(hwc_frect_t crop) {
 
 HWC2::Error DrmHwcTwo::HwcLayer::SetLayerSurfaceDamage(hwc_region_t damage) {
   supported(__func__);
-  // TODO: We don't use surface damage, marking as unsupported
-  unsupported(__func__, damage);
+
+  std::vector<hwcomposer::HwcRect<int>> hwc_rects;
+
+  for (size_t rect = 0; rect < damage.numRects; ++rect) {
+    hwc_rects.push_back({damage.rects[rect].left, damage.rects[rect].top,
+                         damage.rects[rect].right, damage.rects[rect].bottom});
+  }
+
+  hwcomposer::HwcRegion hwc_region = {};
+  hwc_region.kNumRects = damage.numRects;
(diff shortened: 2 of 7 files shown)


os/android/drmhwctwo.cpp
```diff
@@ -670,8 +670,19 @@ HWC2::Error DrmHwcTwo::HwcLayer::SetLayerSourceCrop(hwc_frect_t crop) {
 
 HWC2::Error DrmHwcTwo::HwcLayer::SetLayerSurfaceDamage(hwc_region_t damage) {
   supported(__func__);
-  // TODO: We don't use surface damage, marking as unsupported
-  unsupported(__func__, damage);
+
+  std::vector<hwcomposer::HwcRect<int>> hwc_rects;
+
+  for (size_t rect = 0; rect < damage.numRects; ++rect) {
+    hwc_rects.push_back({damage.rects[rect].left, damage.rects[rect].top,
```
kalyankondapally: Can we just copy this to HwcRegion directly instead of making this temporary copy ?

pallavigopal: HwcRegion and hwc_region_t is not of one to one mapping. The hwc_rect_t and the HwcRect are not matching.

kalyankondapally:
@pallavigopal  u should have everything u need now. Here, https://github.com/01org/IA-Hardware-Composer/blob/master/common/display/displayqueue.cpp#L191 u could probably check if damage is zero and u have previous layers, in that case swap ImportedBuffer from previous overlay layer to current one. This should ensure FB associated with buffer is not closed and fence is not signalled without worrying about the life time now.

pallavigopal:
Re-based @kalyankondapally  please review

common/display/displayplanemanager.cpp
```diff
@@ -458,6 +460,39 @@ void DisplayPlaneManager::ValidateCachedLayers(
...
+        const HwcRect<int> &current = layers->at(i).GetDisplayFrame();
+        if ((previous.left != current.left) || (previous.top != current.top)) {
+          validate = true;
+          break;
+        }
+        validate = false;
+        continue;
+      }
+    } else if (current_surface_damage.kNumRects == 0) {
+      validate = true;
```
kalyankondapally: hmm should we validate in this case ? shouldnt be that in this case nothing has changed compared to previous frame.

kalyankondapally: Can we combine these checks here: https://github.com/01org/IA-Hardware-Composer/blob/master/common/display/displayplanemanager.cpp#L405
You can pass bool damaged_region like we do for render_layers, for this we can just set damagaed_region =true even if one layer needs to be re-composited. This  flag along with render layers can be checked in displayqueue. If both flags are not set u can just swap https://github.com/01org/IA-Hardware-Composer/blob/master/common/display/displayqueue.cpp#L314

common/display/displayplanemanager.cpp
```diff
@@ -458,6 +460,39 @@ void DisplayPlaneManager::ValidateCachedLayers(
...
+  auto layer_begin = layers->begin();
+  auto layer_end = layers->end();
+  bool validate = true;
+  size_t num_layers = layers->size();
+
+  for (size_t i = 0; i <= num_layers; i++) {
+    const HwcRegion &current_surface_damage = layers->at(i).GetSurfaceDamage();
+
+    if (current_surface_damage.kNumRects == 1) {
+      const HwcRect<int> *rect = current_surface_damage.kRects;
```
kalyankondapally: Instead of this I think what we should do is just check if kNumRects =0, if so than assume buffer is same as previous. If it's not same for now we can fallback to 3D (later can think if using 2d block makes sense).

common/display/displayplanemanager.cpp
kalyankondapally: In cases where we fall back to 3D, we can check https://github.com/01org/IA-Hardware-Composer/blob/master/common/display/displayplanemanager.cpp#L439 here if damage region(rects) of that layer is zero. If region_changed is false and damaged(assuming thats the boolean to track the damage state of layers in this local func) we can assign the buffer associated with last_plane here to the current plane. 

In case we are just scanning out directly need to add surface damage check here https://github.com/01org/IA-Hardware-Composer/blob/master/common/display/displayplanemanager.cpp#L452 and swap previous layer buffer to current layer. This way state is always intact at layer level making it easy in future to track damage for per layer.

kalyankondapally: On a second thought we can even replace https://github.com/01org/IA-Hardware-Composer/blob/master/common/display/displayplanemanager.cpp#L440 with just kNumRects =0, checks since if geometry changed then we should have a damage rect here.

pallavigopal: This patch is the basic patch which will avoid the complete compose and flip if the all the layers are not changed. In the next final set of patches with all the optimizations including per layer damage in that scenario am planning to reuse the ValidateCachedLayer function as such and will consider your review comments also. 
Also one more think in the ValidateCachedLayer we are allocating the buffer not reusing the previous layer buffers. 

For the last comment we cant replace the left and right checks there are teh scenarios where the buffer content not changed but the src x and y co-ordinates changed in that case we need to update the src co-ordinates accordingly

common/display/displayplanemanager.cpp
pallavigopal: Please find the Surfacedamage description in the https://source.android.com/devices/halref/hwcomposer2_8h_source.html

 1697  * If damage.numRects > 0, then it may be assumed that any portion of the source
 1698  * buffer not covered by one of the rects has not been modified this frame. If
 1699  * damage.numRects == 0, then the whole source buffer must be treated as if it
 1700  * has been modified.
 1701  * 
 1702  * If the layer's contents are not modified relative to the prior frame, damage
 1703  * will contain exactly one empty rect([0, 0, 0, 0]).

KnumRects = 0 then the entire frame content is changed no damage region.

os/android/drmhwctwo.cpp
kalyankondapally: k resolved

common/display/displayqueue.cpp
```diff
@@ -219,6 +219,7 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
     overlay_layer.SetBlending(layer->GetBlending());
     overlay_layer.SetSourceCrop(layer->GetSourceCrop());
     overlay_layer.SetDisplayFrame(layer->GetDisplayFrame());
+    overlay_layer.SetSurfaceDamage(layer->GetSurfaceDamage());
```
kalyankondapally: Thanks for claryfying things. The only thing I think we should resolve here is instead of having ValidateSurfaceDamage as a separate function, lets have it here. This would have the advantage that for now we can have the minimum case where if no layer has damage than we skip commit (like u do now) but in next steps u can easily modify it to swap previous and current layer buffer here itself and not having to worry about modifying anything in the followup steps

common/display/displayqueue.cpp
```diff
@@ -230,6 +231,12 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
       ETRACE("Failed to create fence for layer, error: %s", PRINTERROR());
   }
 
+  if (!display_plane_manager_->ValidateSurfaceDamage(&layers,
```
kalyankondapally: upon changing above, this would be the if (!validate) check here

pallavigopal:
@kalyankondapally Address the review comments please review

common/display/displayplanemanager.cpp
```diff
@@ -265,6 +265,7 @@ bool DisplayPlaneManager::CommitFrame(const DisplayPlaneStateList &comp_planes,
   for (const DisplayPlaneState &comp_plane : comp_planes) {
     DisplayPlane *plane = comp_plane.plane();
     const OverlayLayer *layer = comp_plane.GetOverlayLayer();
+
```
kalyankondapally: u can remove these changes in DisplayPlaneManager

pallavigopal: ok sure

PR #34, opened 2017-03-23
Add nv12 support

js0701:
NV12 is not supported as SCANOUT in older platforms, so try to use RENDERING only;
Some logic should be change in overlaybuffer ImportImage, as NV12 is 2 plane sformat
JIRA:  IAHWC-52
Tests: testlayers -j jsonconfigs/video1layer_nv12.json

### DIFF
--- a/common/core/overlaybuffer.cpp
+++ b/common/core/overlaybuffer.cpp
@@ -65,23 +65,40 @@ GpuImage OverlayBuffer::ImportImage(GpuDisplay egl_display) {
   // Note: If eglCreateImageKHR is successful for a EGL_LINUX_DMA_BUF_EXT
   // target, the EGL will take a reference to the dma_buf.
   if (is_yuv_) {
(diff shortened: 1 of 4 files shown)


common/core/overlaybuffer.cpp
```diff
@@ -65,23 +65,40 @@ GpuImage OverlayBuffer::ImportImage(GpuDisplay egl_display) {
   // Note: If eglCreateImageKHR is successful for a EGL_LINUX_DMA_BUF_EXT
   // target, the EGL will take a reference to the dma_buf.
   if (is_yuv_) {
-    const EGLint attr_list_yv12[] = {
```
kalyankondapally: You should fix commit message *is not*

Also, add one Jira and Tests would be u video test u just added. You can list it here.

js0701: @kalyankondapally  done

kalyankondapally:
merged

PR #36, opened 2017-03-24
vk: Define missing Renderer overrides

strassek:
Since commit 78a230016b RestoreState() and MakeCurrent() are required to be
redefined in classes inheriting from Renderer. There doesn't appear to be
anything needed from Vulkan for these calls, so for now add noop overrides
to make the build green again.

Jira: IAHWC-40
Test: Vulkan backend compiles


kalyankondapally:
merged

PR #37, opened 2017-03-24
vk: Make each VKSurface track its own fb

strassek:
Only 1 framebuffer is being created as it's being tracked at global scope.
Add a member field to VKSurface so each instance has its own fb, and then
assign it to the global target during MakeCurrent().

While we're at it, remove dst_image_ and dst_image_view_ from global scope,
as they are only ever used inside vksurface.

Jira: IAHWC-40
Test: Try demo apps and check that there is no flickering


kalyankondapally:
merged thx

PR #39, opened 2017-03-29
Add color correction and broadcast rgb test to layers test app.

yugang:
Moved test parameters of color correction APIs, broadcast rgb API
and power mode APIs to json file. test parameters of color correction
APIs, broadcast rgb API and power mode APIs will use initialized value
if the json doesn't include those parameters.
Updated the original layerfromjson.cpp/.h to jsonhandlers.cpp/.h for
common using.
Updated and formatted the json files for testing.

JIRA: IAHWC-55
Test: Build hwc and update the parameters of the json to check if the
      test app testlayers runs normally and effect shows normally
      following the parameters. No differents with old json for old tests.

also fixed wrong parameter type of set gamma even though it isn't impact test result.

@kalyankondapally ， @js0701 ，@pallavigopal (i moved power mode test inputs to json as it has same situation as color correction, pls provide feedbacks), pls help to review & thanks


kalyankondapally:
@yugang probably u want to add comments to commit message like new way of enableing power mode tests. Previously it was passing --p=doze etc

yugang:
Done, @kalyankondapally, @pallavigopal, thanks

kalyankondapally:
merged

PR #40, opened 2017-03-30
Vulkan fixes

strassek:
Here are a few changes that make the Vulkan backend function better for longer periods of time.


kalyankondapally:
merged

PR #41, opened 2017-03-30
Try to leverage pipe scaler. Still need to further consideration

js0701:
Not mature yet. Just for review

### DIFF
--- a/os/linux/gbmbufferhandler.cpp
+++ b/os/linux/gbmbufferhandler.cpp
@@ -85,9 +85,9 @@ bool GbmBufferHandler::CreateBuffer(uint32_t w, uint32_t h, int format,
   temp->import_data.height = gbm_bo_get_height(bo);
   temp->import_data.format = gbm_bo_get_format(bo);
 #if USE_MINIGBM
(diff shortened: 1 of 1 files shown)


common/display/displayqueue.cpp
```diff
@@ -340,6 +356,50 @@ void DisplayQueue::HandleExit() {
...
+  }
+  
+  return true;
+}
+
+void DisplayQueue::ApplyPipeScaler(
+    DisplayPlaneStateList& current_composition_planes,
+    drmModeAtomicReqPtr pset) {
+  // here we define a simple strategy about how to use piple scaler
+  // single plane and single layer
```
kalyankondapally: We need to have this check in ValidateLayers in DisplayPlaneManager, ensure in this case test commit passes so we dont fallback to 3D composition. Otherwise this here wont make sense. Perhaps u can make a help class for pipe scaling which will be owned by DisplayQueue and pointer passed to DisplayPlaneManager in ValidateLayers call.

js0701: Right. Sounds a good idea @kalyankondapally

js0701:
Not effective anymore

PR #44, opened 2017-04-04
Fix the cursor handling if the HW cursor is not working

pallavigopal:
Cursor is not positioned correctly when we fall back to the
3D composition fixed it by moving the cursor as a normal layer
and also handle like rest of input layers.

Now cursor can be assigned to overlay plane or 3D composition based
on the available input layers which will avoid unnecessary composition
due to cursor change every time.

Jira: AIA-334/IAHWC-64
Test: Now cursor and navigation bar is visible as expected in most of the apps

### DIFF
--- a/common/display/displayplanemanager.cpp
+++ b/common/display/displayplanemanager.cpp
@@ -175,7 +175,28 @@ std::tuple<bool, DisplayPlaneStateList> DisplayPlaneManager::ValidateLayers(
   for (auto j = layers->rbegin(); j != layers->rend(); ++j) {
     if (j->GetBuffer()->GetUsage() & kLayerCursor) {
       cursor_layer = &(*(j));
-      layer_end = std::next(j).base();
+      // Handle Cursor layer.
+      DisplayPlane *cursor_plane = NULL;
+      if (cursor_layer) {
+        // Handle Cursor layer. If we have dedicated cursor plane, try using it
+        // to composite cursor layer.
+        if (cursor_plane_)
+          cursor_plane = cursor_plane_.get();
+        if (cursor_plane) {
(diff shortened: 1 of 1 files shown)


common/display/displayplanemanager.cpp
```diff
@@ -195,6 +218,7 @@ std::tuple<bool, DisplayPlaneStateList> DisplayPlaneManager::ValidateLayers(
         // it.
         if (!FallbacktoGPU(j->get(), layer, commit_planes)) {
           composition.emplace_back(j->get(), layer, index);
+          const HwcRect<int> &display_frame = layer->GetDisplayFrame();
```
kalyankondapally: probably missed removing this ?

common/display/displayplanemanager.cpp
```diff
@@ -175,7 +175,30 @@ std::tuple<bool, DisplayPlaneStateList> DisplayPlaneManager::ValidateLayers(
...
+        if (cursor_plane) {
+          commit_planes.emplace_back(OverlayPlane(cursor_plane, cursor_layer));
+          // Lets ensure we fall back to GPU composition in case
+          // cursor layer cannot be scanned out directly.
+          if (FallbacktoGPU(cursor_plane, cursor_layer, commit_planes)) {
+            cursor_plane = NULL;
+          }
+        }
+
+        // We need to do this here to avoid compositing cursor with any previous
```
kalyankondapally: u  can remove these comments now. Can u test on SKL too before commiting the changes

pallavigopal: ok tested on SKL things works fine

common/display/displayplanemanager.cpp
pallavigopal: yes

PR #45, opened 2017-04-04
[RFC] Enable runtime backend selection

strassek:
Now that there are two functional compositor backends, it makes sense to
allow for them to coexist in a single build. If they are both built in by
default it is more likely that we will catch build breakages sooner, and it
allows us to have a fallback backend in production.

We are currently exposing three classes per backend, which creates a
challenge as it makes sense for a backend to be in charge of probing for
support, but it should be all-or-nothing, or else we might end up with mixed
objects.

This change explicitly groups factory functions for each backend and binds a
grouping once one of its factories succeeds in creating an object. The
caller doesn't have an obligation to create any particular object first, so
backends will need to do be able to probe for support in any order. We
aren't currently doing any probing in the relevant contructor functions, so
for now we will always just select the backend that is listed first.

The GL backend is selected by default, with VK as a fallback. This order may
change once Vulkan becomes more viable.  Also, in the future we can add
getter/setter functions to enable forcing a particular backend at runtime.

Jira: IAHWC-40
Test: Both backends are still working with demo apps

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -90,9 +90,6 @@ LOCAL_CPPFLAGS += -DDISABLE_OVERLAY_USAGE
 endif
 
 ifeq ($(strip $(BOARD_USES_VULKAN)),)
-LOCAL_CPPFLAGS += \
-	-DUSE_GL
-
 LOCAL_SRC_FILES += \
 	common/compositor/gl/glprogram.cpp \
 	common/compositor/gl/glrenderer.cpp \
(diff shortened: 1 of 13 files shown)


strassek:
Submitting as RFC for now as this change depends on getting the Vulkan build working against Android.

strassek:
Realizing now that having constructors that can fail may be a bad idea. I'm going to look into having more explicit 'probe' or 'is_supported' methods.

PR #46, opened 2017-04-05
detect explicit sync and overlay usage in runtime

foolmm:
Jira: None
Test: pass test with AndroidIA and Android Things


os/android/drmhwctwo.cpp
```diff
@@ -71,6 +73,9 @@ HWC2::Error DrmHwcTwo::Init() {
                                           HWC2::DisplayType::Physical));
 
   displays_.at(HWC_DISPLAY_PRIMARY).Init();
+
+  DrmHwcTwo::GetExplicitSync();
```
txu4: The "DrmHwcTwo::" is not necessary, GetExplicitSync can be called directly as member function.

os/android/drmhwctwo.h
```diff
@@ -28,6 +28,8 @@
 #include <map>
 #include <utility>
 
+extern bool ExplicitSyncEnabled;
+
```
txu4: We'd better define it as static member of DrmHwcTwo

kalyankondapally:
closing this on request of @mwang30

PR #47, opened 2017-04-05
HWC: Add spinlock for scopedfd

txu4:
The fence fd will be accessed/udpated by surfaceflinger main and
binder thread, and race condition may cause binder sessin crash

Test: run 3d gles and vulkan app without tearing
JIRA: https://01.org/jira/browse/AIA-332

Signed-off-by: Randy Xu <randy.xu@intel.com>

### DIFF
--- a/public/scopedfd.h
+++ b/public/scopedfd.h
@@ -18,6 +18,7 @@
 #define PUBLIC_SCOPEDFD_H_
 
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


kalyankondapally:
LGTM @pallavigopal ping if u want to review and merge.

pallavigopal:
rebased & merged

PR #50, opened 2017-04-12
Enable runtime backend selection

strassek:
Submitting as RFC for now as the second commit depends on getting the Vulkan backend building on Android.

### DIFF
--- a/common/compositor/factory.cpp
+++ b/common/compositor/factory.cpp
@@ -17,46 +17,106 @@
 #include "factory.h"
 #include "platformdefines.h"
 
-#ifdef USE_GL
 #include "glrenderer.h"
 #include "glsurface.h"
 #include "nativeglresource.h"
-#elif USE_VK
 #include "nativevkresource.h"
 #include "vkrenderer.h"
 #include "vksurface.h"
-#endif
 
 namespace hwcomposer {
 
-NativeSurface* CreateBackBuffer(uint32_t width, uint32_t height) {
-#ifdef USE_GL
+static NativeSurface* create_gl_surface(uint32_t width, uint32_t height) {
   return new GLSurface(width, height);
-#elif USE_VK
+}
+
+static Renderer* create_gl_renderer() {
+  return new GLRenderer();
+}
+
+static NativeGpuResource* create_gl_resource() {
+  return new NativeGLResource();
+}
+
+static NativeSurface* create_vk_surface(uint32_t width, uint32_t height) {
   return new VKSurface(width, height);
+}
+
+static Renderer* create_vk_renderer() {
+  return new VKRenderer();
+}
+
+static NativeGpuResource* create_vk_resource() {
+  return new NativeVKResource();
+}
+
+struct backend {
+  const char* id;
+  bool (*is_supported)();
+  NativeSurface* (*create_surface)(uint32_t width, uint32_t height);
+  Renderer* (*create_renderer)();
+  NativeGpuResource* (*create_resource)();
+};
+
+static struct backend backends[] = {
+#ifdef USE_VK
+    {"VK", vk_is_supported, create_vk_surface, create_vk_renderer,
+     create_vk_resource},
+    {"GL", gl_is_supported, create_gl_surface, create_gl_renderer,
+     create_gl_resource},
 #else
-  return NULL;
+    {"GL", gl_is_supported, create_gl_surface, create_gl_renderer,
+     create_gl_resource},
+    {"VK", vk_is_supported, create_vk_surface, create_vk_renderer,
+     create_vk_resource},
 #endif
+    {},
+};
+
+static struct backend* active_backend = NULL;
+
+static bool select_backend() {
+  if (active_backend) {
+    return true;
+  }
+
+  for (struct backend* be = backends; be->id; be++) {
+    if (be->is_supported()) {
+      ETRACE("Selecting %s backend\n", be->id);
+      active_backend = be;
+      return true;
+    }
+  }
+
+  ETRACE("Failed to find a spported backend\n");
+  return false;
+}
+
+NativeSurface* CreateBackBuffer(uint32_t width, uint32_t height) {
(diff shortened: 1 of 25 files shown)


kalyankondapally:
@strassek ready to be pulled in ?

strassek:
I'm not sure what this will do for the Android build, it will probably break until we start compiling with all the Vulkan compositor source files.

common/compositor/vk/vkshim.cpp
```diff
@@ -18,22 +18,6 @@
...
-std::vector<VkImageView> src_image_views_;
-std::vector<VkDescriptorImageInfo> src_image_infos_;
-RingBuffer ring_buffer_;
-std::vector<RingBuffer::Allocation> ub_allocs_;
-size_t ub_offset_align_;
-std::vector<VkImageMemoryBarrier> src_barrier_before_clear_;
-VkImageMemoryBarrier dst_barrier_before_clear_;
-VkFramebuffer framebuffer_;
-
 VkFormat DrmToVkFormat(int drm_format) {
```
kalyankondapally: Do we need this ?

strassek: No, I'll remove it.

strassek:
@gaochun Could you give this series a try on Android?

common/compositor/factory.cpp
```diff
@@ -17,46 +17,106 @@
...
+  Renderer* (*create_renderer)();
+  NativeGpuResource* (*create_resource)();
+};
+
+static struct backend backends[] = {
+#ifdef USE_VK
+    {"VK", vk_is_supported, create_vk_surface, create_vk_renderer,
+     create_vk_resource},
+    {"GL", gl_is_supported, create_gl_surface, create_gl_renderer,
+     create_gl_resource},
```
gaochun: Compile failed on Android whether USE_VK is enabled or not. 

As backend is selected with different priority here, both vk and gl source codes are need to compile. 
BOARD_USES_VULKAN switch should be removed from Android.mk, and vk and gl comiple options need to merge together.

strassek: Ah, yes. good point!

strassek:
@gaochun merged gl and vk compile options in Android.mk. Keeping the BOARD_USES_VULKAN option as it is useful for selecting the default backend.

gaochun:
Build passes on Android.

strassek:
@kalyankondapally This series is ready for you to consider pulling it into master.

wsi/overlaybuffer.h
```diff
@@ -50,7 +50,8 @@ class OverlayBuffer {
 
   virtual uint32_t GetFb() const = 0;
 
-  virtual GpuImage ImportImage(GpuDisplay egl_display) = 0;
```
kalyankondapally: Can we have this as GpuDevice and typedef it to EGLDisplay and VkDevice appropriately ? same with GpuImage ?

common/compositor/factory.cpp
```diff
@@ -17,46 +17,106 @@
 #include "factory.h"
 #include "platformdefines.h"
 
-#ifdef USE_GL
 #include "glrenderer.h"
 #include "glsurface.h"
 #include "nativeglresource.h"
-#elif USE_VK
```
kalyankondapally: We might still want a way to disable vulkan on platforms which dont support vulkan. On our side we can enable USE_VK as default and give a board option to disable it if needed/configure option to disable it on Linux ? be->is_supported() would always return false for vulkan in case support is disabled

common/compositor/factory.cpp
kalyankondapally: Here if USE_VK is not set, we can just remove Vulkan from the list. This shouldn\t mean priority but rather if Vulkan support needs to be built or not.

We can have another flag USE_VK_DEFAULT, which would give an option for Vulkan to be as the preferred backend and disabled by default. Once confortable, we can get rid of this option and have vulkan as default unless USE_VK is disabled.

wsi/overlaybuffer.h
strassek: That would require each of these Gpu* types to be a union of all possible types and we would have to add a flag somewhere to declare what type is being used. If we want runtime selection to work, then ImportImage() needs to act differently depending on what backend is calling it. I chose to use function overrides because it seemed a little cleaner.

kalyankondapally: k

common/compositor/factory.cpp
strassek: What is the intent here? I really wanted Vulkan to be a hard build requirement, just like GL.

kalyankondapally: this is so that we should be able to run on old platforms like hsw etc where vulkan support is not available

kalyankondapally: by default we should have vulkan support enabled.

strassek: Even though some hardware platforms cannot support Vulkan, we **can** still build/deploy a software stack that contains Vulkan bits, right? Detecting if support is available already happens at runtime during the call to vk_is_supported().

I  think there is definitely some value in reducing the number of nonstandard build configurations, and we seem to have an opportunity here.

kalyankondapally: It wouldn't even build in this case since you will be missing vulkan headers ? as we will be depending on the fact that Mesa would be built in this case. Does Mesa export the headers alwyas irrespective of not being built with vulkan support inculded ?

strassek: Right, I'm suggesting that we add Vulkan as a hard build requirement. So Mesa does need to be built with Vulkan enabled, and the Vulkan loader library needs to be installed. Is there some scenario where that would be an impossible configuration? Again, this requirement doesn't mean that we have to change the build configuration based on the target platform.

kalyankondapally: ha k

strassek:
This is stale. Will reopen once I have a chance to rebase the patches.

PR #51, opened 2017-04-13
Native surafce release and layer allocation change

pallavigopal:
Not tested completely need to more tests. Also need to clean up the code.

### DIFF
--- a/common/compositor/nativesurface.cpp
+++ b/common/compositor/nativesurface.cpp
@@ -28,9 +28,8 @@ NativeSurface::NativeSurface(uint32_t width, uint32_t height)
       buffer_handler_(NULL),
       width_(width),
       height_(height),
-      ref_count_(0),
       framebuffer_format_(0),
-      in_flight_(false) {
+      in_use_(false) {
 }
 
 NativeSurface::~NativeSurface() {
@@ -52,7 +51,6 @@ bool NativeSurface::Init(OverlayBufferManager *buffer_manager) {
     return false;
   }
 
-  ref_count_ = 0;
   InitializeLayer(buffer_manager, native_handle_);
 
   return true;
@@ -72,13 +70,7 @@ void NativeSurface::SetNativeFence(int fd) {
 }
 
 void NativeSurface::SetInUse(bool inuse) {
-  if (inuse) {
-    ref_count_ = 3;
-  } else if (ref_count_) {
-    ref_count_--;
-  }
-
-  in_flight_ = false;
+  in_use_ = inuse;
 }
 
 void NativeSurface::SetPlaneTarget(DisplayPlaneState &plane, uint32_t gpu_fd) {
@@ -91,7 +83,7 @@ void NativeSurface::SetPlaneTarget(DisplayPlaneState &plane, uint32_t gpu_fd) {
   width_ = display_rect.right - display_rect.left;
   height_ = display_rect.bottom - display_rect.top;
   plane.SetOverlayLayer(&layer_);
-  in_flight_ = true;
+  SetInUse(true);
 
   if (framebuffer_format_ == format)
     return;
@@ -112,10 +104,6 @@ void NativeSurface::InitializeLayer(OverlayBufferManager *buffer_manager,
   layer_.SetBlending(HWCBlending::kBlendingPremult);
   layer_.SetTransform(0);
   layer_.SetBuffer(buffer);
-}
-
-void NativeSurface::ResetInFlightMode() {
-  in_flight_ = false;
 }
 
 }  // namespace hwcomposer
--- a/common/compositor/nativesurface.h
+++ b/common/compositor/nativesurface.h
@@ -68,7 +68,7 @@ class NativeSurface {
   void SetInUse(bool inuse);
 
   bool InUse() const {
-    return ref_count_ > 1 || in_flight_;
+    return in_use_;
   }
 
   void SetPlaneTarget(DisplayPlaneState& plane, uint32_t gpu_fd);
@@ -85,9 +85,8 @@ class NativeSurface {
   NativeBufferHandler* buffer_handler_;
   uint32_t width_;
   uint32_t height_;
-  uint32_t ref_count_;
+  bool in_use_;
   uint32_t framebuffer_format_;
-  bool in_flight_;
   NativeFence fd_;
 };
 
--- a/common/display/displayplanemanager.cpp
+++ b/common/display/displayplanemanager.cpp
@@ -110,8 +110,6 @@ void DisplayPlaneManager::BeginFrameUpdate() {
     (*i)->SetEnabled(false);
   }
 
-  if (!in_flight_surfaces_.empty())
-    std::vector<NativeSurface *>().swap(in_flight_surfaces_);
 }
 
 std::tuple<bool, DisplayPlaneStateList> DisplayPlaneManager::ValidateLayers(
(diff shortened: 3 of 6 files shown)


common/display/displayplanemanager.cpp
```diff
@@ -197,42 +196,71 @@ std::tuple<bool, DisplayPlaneStateList> DisplayPlaneManager::ValidateLayers(
...
     // Handle layers for overlay
+  DisplayPlaneState &last_plane = composition.back();
     uint32_t index = 0;
+    uint32_t layers_size = layers->size();
+
+    int32_t num_planes_compose = 0;
+
+    ALOGE("layers_size = %u\n", layers_size);
+    if (layers_size >= 4) {
+      num_planes_compose = layers_size - 3;
```
kalyankondapally: hmm we are just forcing to use only one plane always here?

common/compositor/nativesurface.h
```diff
@@ -68,7 +68,7 @@ class NativeSurface {
   void SetInUse(bool inuse);
 
   bool InUse() const {
-    return ref_count_ > 1 || in_flight_;
+    return in_use_;
```
kalyankondapally: We will need a spinlock here. Since InUse will be queried in application thread and SetinUse can be called from app thread and kmsfencehandler thread

common/compositor/nativesurface.cpp
```diff
@@ -72,13 +70,7 @@ void NativeSurface::SetNativeFence(int fd) {
...
 
 void NativeSurface::SetInUse(bool inuse) {
-  if (inuse) {
-    ref_count_ = 3;
-  } else if (ref_count_) {
-    ref_count_--;
-  }
-
-  in_flight_ = false;
+  in_use_ = inuse;
```
kalyankondapally: We will need a spinlock here. Since InUse will be queried in application thread and SetinUse can be called from app thread and kmsfencehandler thread

common/compositor/nativesurface.cpp
```diff
@@ -91,7 +83,7 @@ void NativeSurface::SetPlaneTarget(DisplayPlaneState &plane, uint32_t gpu_fd) {
   width_ = display_rect.right - display_rect.left;
   height_ = display_rect.bottom - display_rect.top;
   plane.SetOverlayLayer(&layer_);
-  in_flight_ = true;
+  in_use_ = true;
```
kalyankondapally: call SetInUse(true) here, so all thread management is centralized

common/display/displayplanemanager.cpp
```diff
@@ -110,8 +110,8 @@ void DisplayPlaneManager::BeginFrameUpdate() {
     (*i)->SetEnabled(false);
   }
 
-  if (!in_flight_surfaces_.empty())
-    std::vector<NativeSurface *>().swap(in_flight_surfaces_);
+/*  if (!in_flight_surfaces_.empty())
```
kalyankondapally: remove this

common/display/displayplanemanager.cpp
```diff
@@ -324,13 +324,13 @@ bool DisplayPlaneManager::TestCommit(
 }
 
 void DisplayPlaneManager::EndFrameUpdate() {
-  for (auto &fb : surfaces_) {
+/*  for (auto &fb : surfaces_) {
```
kalyankondapally: remove this and u can get rid of EndFrameUpdate call too

common/display/displayplanemanager.cpp
```diff
@@ -351,7 +351,7 @@ void DisplayPlaneManager::EnsureOffScreenTarget(DisplayPlaneState &plane) {
 
   surface->SetPlaneTarget(plane, gpu_fd_);
   plane.SetOffScreenTarget(surface);
-  in_flight_surfaces_.emplace_back(surface);
+//  in_flight_surfaces_.emplace_back(surface);
```
kalyankondapally: remove this.

common/display/displayplanemanager.h
```diff
@@ -92,7 +92,7 @@ class DisplayPlaneManager {
 
   OverlayBufferManager *buffer_manager_;
   std::vector<std::unique_ptr<NativeSurface>> surfaces_;
-  std::vector<NativeSurface *> in_flight_surfaces_;
+ // std::vector<NativeSurface *> in_flight_surfaces_;
```
kalyankondapally: you can remove this

common/display/kmsfencehandler.cpp
```diff
@@ -46,7 +46,9 @@ bool KMSFenceEventHandler::EnsureReadyForNextFrame() {
 }
 
 void KMSFenceEventHandler::WaitFence(uint64_t kms_fence,
-                                     std::vector<OverlayLayer>& layers) {
+                                     std::vector<OverlayLayer>& layers,
+                                     DisplayPlaneStateList &planes_state)
+{
```
kalyankondapally: bracket should be with previous line, can u run clang formatting ?

common/display/kmsfencehandler.cpp
```diff
@@ -59,6 +61,12 @@ void KMSFenceEventHandler::WaitFence(uint64_t kms_fence,
     }
   }
 
+  for (DisplayPlaneState &plane_state : planes_state) {
+    if (plane_state.GetCompositionState() == DisplayPlaneState::State::kRender)	{
```
kalyankondapally: indentation

kalyankondapally:
You can split the nativesurface changes to a separate CL and test it. We can merge it. For composition related patch, lets figure out why it hangs with 4 planes in use instead of forcing it to be 1 always

common/display/kmsfencehandler.cpp
```diff
@@ -82,6 +90,11 @@ void KMSFenceEventHandler::HandleRoutine() {
     close(kms_fence_);
     kms_fence_ = -1;
   }
+
+  for (NativeSurface *surface : surfaces_) {
+     surface->SetInUse(false);
```
kalyankondapally: hmm after thinking bit more, I think it would be better to do this in displayqueue itself before swapping previous and present plane composition states. This will ensure we also handle surfacedamage case where we will continue to use the same overlay buffer in some cases

kalyankondapally: In this case we also dont need to worry about thread synchronisation also.

pallavigopal: Agree but how to make sure the display is not holding the buffer anyway need to rely on the fence only right. Maybe we can add the buffer ref count and reuse the buffer accordingly.

common/display/displayplanemanager.cpp
pallavigopal: no allocating the top layers to the overlays and then compositing & displaying remaining layers into the Primary plane.
Will remove the hardcode value

common/display/kmsfencehandler.cpp
kalyankondapally: when we swap previous and current plane state, we are guaranteed that the previous plane state layers are not displayed anymore. We signal the overlaylayers in a separate thread to avoid case where previous layer are no more displayed but we cannot signal as app has not requested for next frame yet (thats we handle the wait fence in a separate thread and signal layers there). However, for nativesurface this shouldnt be an issue since app/surfaceflinger know nothing about native surface.

pallavigopal: Please clarify how its guarantee that the previous plane state layers are not displayed ? anyway we do async flip right after commit the actual mmio update of the display registers will happen at the next vblank only. Even for the nativesurface same applicable right how will know the previous native surface is no more scanout buffer

kalyankondapally: https://github.com/01org/IA-Hardware-Composer/blob/master/common/display/displayqueue.cpp#L294

we wait here till the last commit request is handled in case app tries to present content before the last has been handled.

PR #53, opened 2017-04-19
Store size of the static buffer in a variable

harishkrupo:
As the buffer is statically allocated, store the size in a variable.
Also, doing a strlen after memset would always return 0. This also fixes the
over heating issue addressed in the below jira.

Jira: IAHWC-69
Tests: Board should boot to home screen and in idle state temperature should
be between 40 to 50 degrees. check /sys/class/thermal/thermal_zone0/temp for
cpu temperature


kalyankondapally:
merged

PR #57, opened 2017-04-21
Avoid virtual display error prints

pallavigopal:
In android we dont compose the virtual display output all the
layers are marked as client and surafceflinger does the composition.
so just return from present display call otherwise we end up with
unnecessary error logs

Jira: None
Test: Able to dump the screenrecord output


kalyankondapally:
lgtm

kalyankondapally:
merged

PR #58, opened 2017-04-21
Add null pointer checks

raghs7:
Broadcast RGB properties are not defined for DSI display panel

Change-Id: I21f1c9adf4780ad6bd45b7dc85a82edcd9414010
Signed-off-by: Ragha Khandenahally <Ragha.Khandenahally@intel.com>

### DIFF
--- a/common/display/displayqueue.cpp
+++ b/common/display/displayqueue.cpp
@@ -111,6 +111,9 @@ bool DisplayQueue::Initialize(uint32_t width, uint32_t height, uint32_t pipe,
   drmModePropertyPtr broadcastrgb_props =
       drmModeGetProperty(gpu_fd_, broadcastrgb_id_);
 
+  if (!broadcastrgb_props)
+    return true;
+
   if (!(broadcastrgb_props->flags & DRM_MODE_PROP_ENUM))
     return false;
 


common/display/displayqueue.cpp
```diff
@@ -111,10 +111,10 @@ bool DisplayQueue::Initialize(uint32_t width, uint32_t height, uint32_t pipe,
   drmModePropertyPtr broadcastrgb_props =
       drmModeGetProperty(gpu_fd_, broadcastrgb_id_);
 
-  if (!(broadcastrgb_props->flags & DRM_MODE_PROP_ENUM))
+  if (broadcastrgb_props && !(broadcastrgb_props->flags & DRM_MODE_PROP_ENUM))
     return false;
 
-  for (int i = 0; i < broadcastrgb_props->count_enums; i++) {
+  for (int i = 0; broadcastrgb_props && i < broadcastrgb_props->count_enums; i++) {
```
bjbeare: order of evaluation is OK here... but this sort of construct always makes me nervous.
Am I the only one that always has to check and prefers a FPIE?

common/display/displayqueue.cpp
```diff
@@ -111,10 +111,10 @@ bool DisplayQueue::Initialize(uint32_t width, uint32_t height, uint32_t pipe,
   drmModePropertyPtr broadcastrgb_props =
       drmModeGetProperty(gpu_fd_, broadcastrgb_id_);
 
-  if (!(broadcastrgb_props->flags & DRM_MODE_PROP_ENUM))
+  if (broadcastrgb_props && !(broadcastrgb_props->flags & DRM_MODE_PROP_ENUM))
```
kalyankondapally: You can make this || condition and include check !broadcastrgb_propos this way we should just return false here, since it doesnt make sense to go any further in case its null

raghs7: returning false means failure to initialize display queue.

common/display/displayqueue.cpp
raghs7: I will replace 2 checks with a 'goto'

common/display/displayqueue.cpp
```diff
@@ -111,6 +111,9 @@ bool DisplayQueue::Initialize(uint32_t width, uint32_t height, uint32_t pipe,
   drmModePropertyPtr broadcastrgb_props =
       drmModeGetProperty(gpu_fd_, broadcastrgb_id_);
 
+  if (!broadcastrgb_props)
+	  goto out;
```
kalyankondapally: just return true here with a comment why this is still a valid case i.e. Its valid case that we might not be supporting this on certain panels. You are missing Jira: Tests: In commit message

kalyankondapally: To make it less confusing you can just wrap rest of the code under if (broadcastrgb_props) (except for final return value)

kalyankondapally:
merged after fixing commit message

PR #60, opened 2017-04-24
Hot plug event handler modified to reduce cycles consumed

amuthuk1:
HotPlugEventHandler is in runtime path for all usecases and hence matters for performance.

Moved memset and fd query out of the loop
Reduced buffer size to 256
Added sleep to avoid busy looping

Test:Device boots to home screen
     Perf bench marks with AIA

Result: Improves performance.

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -164,15 +164,15 @@ bool GpuDevice::DisplayManager::Init(uint32_t fd) {
 
 void GpuDevice::DisplayManager::HotPlugEventHandler() {
   CTRACE();
-  uint32_t buffer_size = 1024;
-  char buffer[buffer_size];
+  int fd = hotplug_fd_.get();
+  char buffer[DRM_HOTPLUG_EVENT_SIZE];
   int ret;
 
+  memset(&buffer, 0, sizeof(buffer));
   while (true) {
     bool drm_event = false, hotplug_event = false;
-    memset(&buffer, 0, sizeof(buffer));
-    size_t srclen = buffer_size - 1;
-    ret = read(hotplug_fd_.get(), &buffer, srclen);
+    size_t srclen = DRM_HOTPLUG_EVENT_SIZE - 1;
+    ret = read(fd, &buffer, srclen);
     if (ret <= 0) {
       if (ret < 0)
         ETRACE("Failed to read uevent. %s", PRINTERROR());
--- a/public/gpudevice.h
+++ b/public/gpudevice.h
@@ -27,6 +27,8 @@ namespace hwcomposer {
 
 class NativeDisplay;
 
+#define DRM_HOTPLUG_EVENT_SIZE 256
+
 class DisplayHotPlugEventCallback {
  public:
   virtual ~DisplayHotPlugEventCallback() {


common/core/gpudevice.cpp
```diff
@@ -201,6 +201,8 @@ void GpuDevice::DisplayManager::HotPlugEventHandler() {
           "UpdateDisplayState.");
       UpdateDisplayState();
     }
+    /* Adding sleep to avoid busy looping before a retry */
+    usleep(1000);
```
kalyankondapally: hmm we shouldnt busy loop here, we should return when we have nothing to read. Is sleep really needed here ?

yogesh-marathe: Kalyan, The loop is breaking / returning only when read fails. There could be a case of race / priority inversion (very corner case) where this thread will hog entire system, sleep ensures that wont happen. We should either remove the busy loop itself or have this sleep on the safer side. There is no perf improve due to this but this is more of precautionary measure. Busy loops are bad!

kalyankondapally: hmm we break when there is nothing more to read from fd. How can we ensure we are reading everything from fd without the loop ?

yogesh-marathe: Ok, in that case a loop is required so lets keep it with a sleep. We won't need sleep even if there was one blocking system call inside the loop but here that's not the case, that read() is non-blocking and will wait till it reads bytes or fails.

tpalli: I'm not familiar with this part of code but maybe check the size with lseek() and then read all that at once?

kalyankondapally: ya, probably we can split this sleep thing from this PR. In subsequent one, move to make use of lseek. I am afraid we will run into nasty race issues with sleep when u have "Real" hot plug events. Is that fine @yogesh-marathe

yogesh-marathe: I think lseek() will help when its really big buffer where you are reading in same buffer multiple times or in a file. Here I think its small number of bytes, which we read and parse until we find both strings (drm evnent and hot plug). If you add lseek here I think you are just adding one more system call.

kalyankondapally: Lets separate sleep to a separate PR, and we can check the perf impact, ensure it doesnt affect hotplug event in any way and make a call. Rest looks good to be merged.

yogesh-marathe: Yes we have checked hotplug functionality with this and that seems to be fine.

yogesh-marathe: Created PR without sleep.

tpalli: > If you add lseek here I think you are just adding one more system call.

IMO it would be more important to make the existing code more robust than worry about performance of a single system call. Adding sleeps to the code indicates that there is problem with the whole approach here.

yogesh-marathe: Tapani, Its not a single system call, HotPlugEventHandler gets called multiple times (even without actual hotplug, I didn't understand that behaviour!!), we have observed this when we put prints there. If lseek can get rid of loop itself we can get rid of that sleep at cost of 'one more sys call at runtime' (its not a single call).

kalyankondapally:
merged

PR #61, opened 2017-04-24
Avoiding busy loop in HotPlugEventHandler

yogesh-marathe:
The busy loop in HotPlugEventHandler can cause issues in race conditions
as it doesn't have any blocking system call or sleep. This patch avoids that.

Test: Android perf benches and hotplug on the device.

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -202,6 +202,11 @@ void GpuDevice::DisplayManager::HotPlugEventHandler() {
       UpdateDisplayState();
     }
   }
+  /* Adding sleep to avoid busy loop. The busy loop returns only
+   * if read() completes / fails. Race conditions can cause this
+   * busy loop hog the system. This 1 ms sleep can avoid that.
+   */
+  usleep(1000);
 }
 
 void GpuDevice::DisplayManager::HandleWait() {


yogesh-marathe:
Please refer the discussion here first https://github.com/01org/IA-Hardware-Composer/pull/60#discussion_r112874115

common/core/gpudevice.cpp
```diff
@@ -202,6 +202,11 @@ void GpuDevice::DisplayManager::HotPlugEventHandler() {
       UpdateDisplayState();
     }
   }
+  /* Adding sleep to avoid busy loop. The busy loop returns only
+   * if read() completes / fails. Race conditions can cause this
+   * busy loop hog the system. This 1 ms sleep can avoid that.
+   */
+  usleep(1000);
```
kalyankondapally: Still want this ? @tpalli  @yogesh-marathe ? do we have a data with and without this patch to make a decision ?

tpalli: I don't think we should usleep here. In case this is really required then we better look what is causing the busyloop and fix that.

yogesh-marathe: @kalyankondapally  Sorry I missed this. I had tested with and without patch, there was no functional issue.
@tpalli I'm ok to drop this if there is a solution to avoid busy loop, please suggest and I can try that. We already discussed lseek one here https://github.com/01org/IA-Hardware-Composer/pull/60#discussion_r112874115

PR #62, opened 2017-04-26
Support DRM_FORMAT_YVU420_ANDROID enum

tpalli:
Support minigbm specific format by switching it to standard
one when importing image.

Jira: None.
Test: Working video playback after minigbm rebase.

### DIFF
--- a/common/core/overlaybuffer.cpp
+++ b/common/core/overlaybuffer.cpp
@@ -26,6 +26,11 @@
 
 #include "hwctrace.h"
 
+#ifdef USE_MINIGBM
+// minigbm specific DRM_FORMAT_YVU420_ANDROID enum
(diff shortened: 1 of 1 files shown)


common/core/overlaybuffer.cpp
```diff
@@ -26,6 +26,11 @@
 
 #include "hwctrace.h"
 
+#ifdef USE_MINIGBM
+// minigbm specific DRM_FORMAT_YVU420_ANDROID enum
```
kalyankondapally: Ha didnt realise we are defining it here, in this case we can remove this ifdef ?. With that removed you can merge the patch Tapani

tpalli: ok will do

kalyankondapally: I actually cleaned it up and merged since, minigbm got rebased

kalyankondapally:
merged

PR #63, opened 2017-04-27
Add surface damage support

kalyankondapally:
Added the basic support to avoid the complete validate if the
given layers are same as previous one.

Jira: IAHWC-51
Test: No regressions with Linux app and on Android

### DIFF
--- a/common/core/overlaylayer.cpp
+++ b/common/core/overlaylayer.cpp
@@ -112,6 +112,29 @@ void OverlayLayer::ValidatePreviousFrameState(const OverlayLayer& rhs) {
   }
 }
 
+void OverlayLayer::SetSurfaceDamage(const HwcRegion& surface_damage,
+                                    const OverlayLayer& rhs) {
+  // check the surafce damage if its 0 for all layers then skip validate
+  // and commit
+  if (surface_damage.kNumRects == 1) {
+    const HwcRect<int>* rect = surface_damage.kRects;
+    if ((rect->top == 0) && (rect->bottom == 0) && (rect->left == 0) &&
+        (rect->right == 0)) {
+      layer_attributes_changed_ = false;
+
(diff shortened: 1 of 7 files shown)


kalyankondapally:
For now, I just limited to use surface damage to track geometry changes while we figure out when to assume buffer can be re-used from previous frame
@pallavigopal u want to take a look

os/android/drmhwctwo.cpp
```diff
@@ -504,11 +504,16 @@ HWC2::Error DrmHwcTwo::HwcDisplay::SetActiveConfig(hwc2_config_t config) {
...
                                                    int32_t acquire_fence,
                                                    int32_t dataspace,
-                                                   hwc_region_t /*damage*/) {
+                                                   hwc_region_t damage) {
   supported(__func__);
   client_layer_.set_buffer(target);
   client_layer_.set_acquire_fence(acquire_fence);
   client_layer_.SetLayerDataspace(dataspace);
+
+  ALOGE("SetLayerSurfaceDamage SetClientTarget %d \n", damage.numRects);
```
kalyankondapally: I will get rid of these logs

common/core/overlaylayer.cpp
```diff
@@ -112,6 +112,29 @@ void OverlayLayer::ValidatePreviousFrameState(const OverlayLayer& rhs) {
   }
 }
 
+void OverlayLayer::SetSurfaceDamage(const HwcRegion& surface_damage,
+                                    const OverlayLayer& rhs) {
+  // check the surafce damage if its 0 for all layers then skip validate
+  // and commit
```
pallavigopal: please remove this commit from the above comments

Apart from that Looks Good for me

kalyankondapally:
fixed and merged

PR #65, opened 2017-05-05
Make HWC library as a shared library

shobhitka:
Install public headers and also fixes build errors when gbm is uses instead of minigbm


shobhitka:
@kalyankondapally please review and merge

kalyankondapally:
Thanks merged, can u do a patch to even install all jsonconfig files ?

PR #67, opened 2017-05-08
Fix error when initializing virtual display.

kalyankondapally:
SurfaceFlinger expects SetActiveConfig to pass but we returned
false here.

Jira: None.
Test: Dont see any HWC failures when initializing Virtual Display.


kalyankondapally:
merged

PR #68, opened 2017-05-09
Enabling HWC standalone test for Android

kjkaranj:
Jira: None.
Tests: Build passes on Linux Enabling HWC tests for custom builds.
       Works with grallocbufferhandler.

Signed-off-by: Kedar <kedar.j.karanje@intel.com>
Signed-off-by: Yogesh <yogesh.marathe@intel.com>


kjkaranj:
Android.mk : Disabled subsequent Android.mk includes (as there are some build issues (tests folder tries to build before hwc so hence build fails )am working with Sundar to get it fixed.

yogesh-marathe:
Test on Linux was pending. We'll test on Linux and re-generate pull request.

PR #69, opened 2017-05-10
Update gbm bo definition to fix drmModeAddFB2 failure.

yugang:
Move gbm bo from native handle to GBM buffer class as there
are some impacts to refer gbm bo in native handle which caused
drmModeAddFB2 error with "No such file or directory".

Jira: None.
Test: Build the HWC with tests and run them on Linux without
      error, and all tests run normally.

### DIFF
--- a/os/linux/gbmbufferhandler.cpp
+++ b/os/linux/gbmbufferhandler.cpp
@@ -68,44 +68,43 @@ bool GbmBufferHandler::CreateBuffer(uint32_t w, uint32_t h, int format,
   if (gbm_format == 0)
     gbm_format = GBM_FORMAT_XRGB8888;
 
-  struct gbm_bo *bo = gbm_bo_create(device_, w, h, gbm_format,
-                                    GBM_BO_USE_SCANOUT | GBM_BO_USE_RENDERING);
+  gbm_bo_ = gbm_bo_create(device_, w, h, gbm_format,
+                          GBM_BO_USE_SCANOUT | GBM_BO_USE_RENDERING);
 
-  if (!bo) {
-    bo = gbm_bo_create(device_, w, h, gbm_format, GBM_BO_USE_RENDERING);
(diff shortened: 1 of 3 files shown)


yugang:
@kalyankondapally @kjkaranj, i just verified it in Linux, not sure if it will impact Android. and the PR is just for review firstly, as i am not sure if the roor cause is in other components(i just fixed in HWC side).

os/linux/gbmbufferhandler.cpp
```diff
@@ -68,44 +68,43 @@ bool GbmBufferHandler::CreateBuffer(uint32_t w, uint32_t h, int format,
   if (gbm_format == 0)
     gbm_format = GBM_FORMAT_XRGB8888;
 
-  struct gbm_bo *bo = gbm_bo_create(device_, w, h, gbm_format,
-                                    GBM_BO_USE_SCANOUT | GBM_BO_USE_RENDERING);
+  gbm_bo_ = gbm_bo_create(device_, w, h, gbm_format,
```
kalyankondapally: hm this is wrong as now if we call create buffer twice, the previous buffer pointer is lost. It should be with handle only.

yugang: can we keep: 
```
  temp->bo = gbm_bo_;
  *handle = temp;
```
here for bo tracking, in fact, currently, i don't see any place refer to handle->bo.

yugang: ao, maybe involve in double release/destory risk of bo if keep track bo on both handle and GbmBufferHandler

yugang:
found if always use gbm_bo_import in GbmBufferHandler::ImportBuffer() instead of check "if (!handle->bo)" to skip the gbm_bo_import(), the issue disappears. still not sure why bo impact the framebuffer create.

PR #70, opened 2017-05-11
Enabling HWC standalone test for Android

kjkaranj:
Signed-off-by: Kedar Karanje <kedar.j.karanje@intel.com>


kjkaranj:
Verified with changes on linux and its working fine , 
To run on Android : 1> Make full build 
                               2> Uncomment hwcomposer/Android.mk : below line
                                     include $(LOCAL_PATH)/tests/third_party/json-c/Android.mk
                               3> build only hwcomposer : mm
testlayers is in /vendor/bin along with the jsonconfig/*.json files

Android.mk
```diff
@@ -116,4 +116,5 @@ LOCAL_MODULE_CLASS := SHARED_LIBRARIES
 LOCAL_MODULE_SUFFIX := $(TARGET_SHLIB_SUFFIX)
 include $(BUILD_SHARED_LIBRARY)
 
+#include $(LOCAL_PATH)/tests/third_party/json-c/Android.mk
```
kalyankondapally: hmm I think we can leave it uncommented. If we dont want to have tests part of user builds, than can include this only for eng builds. For user builds if someone wants it than they can uncomment it.

PR #75, opened 2017-05-22
replacing gralloc.h with gralloc1.h to enable gralloc1.0

munish-b:


os/android/grallocbufferhandler.cpp
```diff
@@ -77,7 +88,7 @@ bool GrallocBufferHandler::CreateBuffer(uint32_t w, uint32_t h, int /*format*/,
                                  GRALLOC_USAGE_HW_FB | GRALLOC_USAGE_HW_RENDER |
                                      GRALLOC_USAGE_HW_COMPOSER);
   temp->handle_ = temp->buffer_->handle;
-  gralloc_->registerBuffer(gralloc_, temp->handle_);
+  (*pfn)(gralloc1_dvc, temp->handle_);
```
tpalli: please call the function in normal way, "pfn(gralloc1_dvc, temp->handle_)", it could be also renamed something else than 'pfn' which does not tell what the function does.

PR #78, opened 2017-05-26
Update bo map function for build support of gbm or mingbm.

yugang:
Mesa gbm has no plane parameter defined, so add build option
for mesa gbm.

Jira: None.
Tests: Build and run testlayers normally with or without mingbm.


kalyankondapally:
merged

PR #82, opened 2017-06-08
Fix the cursor reorder issue

pyn1:
Sometimes cursor layer comes as the lowest zorder and we end up
assigning to the primary plane.
Introduced a flag to check the above condition and hence
disabling cursor plane.

Jira: https://jira01.devtools.intel.com/browse/OAM-46186
Tests: Android boot-up and issue is not see
Signed-off-by: N, Poornima Y <poornima.y.n@intel.com>

### DIFF
--- a/os/android/drmhwctwo.cpp
+++ b/os/android/drmhwctwo.cpp
@@ -427,6 +427,9 @@ HWC2::Error DrmHwcTwo::HwcDisplay::PresentDisplay(int32_t *retire_fence) {
   // order the layers by z-order
   bool use_client_layer = false;
   uint32_t client_z_order = 0;
+  bool use_cursor_layer = false;
+  uint32_t cursor_z_order = 0;
+  DrmHwcTwo::HwcLayer *cursor_layer;
   *retire_fence = -1;
   std::map<uint32_t, DrmHwcTwo::HwcLayer *> z_map;
 
@@ -443,8 +446,12 @@ HWC2::Error DrmHwcTwo::HwcDisplay::PresentDisplay(int32_t *retire_fence) {
   for (std::pair<const hwc2_layer_t, DrmHwcTwo::HwcLayer> &l : layers_) {
     switch (l.second.validated_type()) {
       case HWC2::Composition::Device:
-      case HWC2::Composition::Cursor:
         z_map.emplace(std::make_pair(l.second.z_order(), &l.second));
+        break;
+      case HWC2::Composition::Cursor:
+        use_cursor_layer = true;
+        cursor_layer = &l.second;
+        cursor_z_order = l.second.z_order();
         break;
       case HWC2::Composition::Client:
         // Place it at the z_order of the highest client layer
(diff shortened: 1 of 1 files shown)


pyn1:
@kalyankondapally / @pallavigopal 
 Please review

Thanks

common/display/displayplanemanager.cpp
```diff
@@ -111,6 +111,9 @@ std::tuple<bool, DisplayPlaneStateList> DisplayPlaneManager::ValidateLayers(
   auto layer_begin = layers.begin();
   auto layer_end = layers.end();
   bool render_layers = false;
+  // Below flag takes care of condition,when the cursor layer is assigned the
+  // lower zorder
+  bool cursor_plane_enable = true;
```
kalyankondapally: hmm we shouldn't assign cursor plane to primary for any reason.  Is it so that in this case we just have cursor layer and nothing else ? If we have more than one layer perhaps the bug is in present call in drmhwc2 where t layers are expected to be populated in right z order. Can you check whats happening in that case and why we have cursor layer as  first layer?

kalyankondapally: https://github.com/01org/IA-Hardware-Composer/blob/master/os/android/drmhwctwo.cpp#L443

Can u check we are having the right z order here, 
https://github.com/01org/IA-Hardware-Composer/blob/master/public/nativedisplay.h#L86  expectation is that we have layers in increasing z order (0 = primary  etc)

pyn1: Hi Kalyan,
In this particular UC, we see 2 layers including cursor layer.
I have checked the Z-order of the layers in the presentDisplay call. cursor layer is coming at the lower z-order than the other layer.
I have missed the case when we get only cursor layer, that case I will assign to cursor layer.
Thanks

pallavigopal: From Gen9 onwards all the planes are windowable, we can flip the cursor layer to primary also.  But in Gen8 and below the primary pane is non windowable we cant flip the cursor layer to primary.
Please move the cursor layer to cursor plane. Always cursor is expected to have highest z-order, not sure why we get some lower zorder maybe due to some transparency in the high layer things are working

kalyankondapally: @pyn1 Better to ensure the z-order is correct for cursor when populating layers in present call in drmhwctwo.cpp rather than complicating logic here.

pyn1:
correct. I'm changing the Z-order in presentDisplay call only now.

pyn1:
Also adding a logic to skip the primary plane assignment in displayplanemanager.cpp , when there is only one layer(cursor)

common/display/displayplanemanager.cpp
kalyankondapally: Atually the other bug here is the for loop for (auto j = layers.rbegin(); j != layers.rend(); ++j) 

it should be for (auto j = layer_begin; j != layer_end; ++j)

os/android/drmhwctwo.cpp
```diff
@@ -466,6 +466,13 @@ HWC2::Error DrmHwcTwo::HwcDisplay::PresentDisplay(int32_t *retire_fence) {
   for (std::pair<const uint32_t, DrmHwcTwo::HwcLayer *> &l : z_map) {
     layers.emplace_back(l.second->GetLayer());
   }
+  // incase cursor comes at lowest z-order, move it to higest z-order
+  auto l = z_map.begin();
```
kalyankondapally: instead of this we can do something similar to what we do with client layer here i.e. if cursor layer is in use z_order is the last layer z_order + 1 ?

os/android/drmhwctwo.cpp
```diff
@@ -443,8 +446,14 @@ HWC2::Error DrmHwcTwo::HwcDisplay::PresentDisplay(int32_t *retire_fence) {
   for (std::pair<const hwc2_layer_t, DrmHwcTwo::HwcLayer> &l : layers_) {
     switch (l.second.validated_type()) {
       case HWC2::Composition::Device:
-      case HWC2::Composition::Cursor:
         z_map.emplace(std::make_pair(l.second.z_order(), &l.second));
+        cursor_z_order = std::max(cursor_z_order, l.second.z_order());
```
kalyankondapally: Instead of this ,  I think we can do  std::max(cursor_z_order, <last layer in map>.second.z_order()); ?
map should order the layers by z order, so last layer should have highest z order.

pyn1: okay. Submitted with the changes suggested. Please review. Thanks

kalyankondapally:
merged

PR #83, opened 2017-06-11
Review

kalyankondapally:


kalyankondapally:
@pallavigopal Can u check if things are any different with these changes ?

kalyankondapally:
These are needed to enable support for caching 3D composition results:
https://github.com/01org/IA-Hardware-Composer/commit/3409522e5dbcf373d8a649c8ede76f156d749e35

Still has some issues with transparent layer and when using cursor.

PR #84, opened 2017-06-12
fix compilation error when BOARD_USES_GRALLOC1 is set

tpalli:
Jira: None.
Test: Code compiles without errors on Android when
      BOARD_USES_GRALLOC1 is set


kalyankondapally:
merged thanks @tpalli

PR #86, opened 2017-06-13
Change definition place to fix assignment issue

yugang:
Move layer_parameter definition to outside of for to avoid the
assignment issue which will cause the render init failure.

Jira: None.
Test: Test app works normally without render init issue.


kalyankondapally:
Thanks

PR #89, opened 2017-07-04
REVERT [PARTIAL] : Disable invoking of hwc.info service

Kishore409:
In User debug builds, while Joule board boots via ColdBoot or ALT+CTRL+DEL,
it loops in for a while between Android TEXT and Android Animation before
reaching the home screen.
This patch disables invoking of hwc.info service which fixes the issue.

### DIFF
--- a/os/android/drmhwctwo.cpp
+++ b/os/android/drmhwctwo.cpp
@@ -95,7 +95,9 @@ HWC2::Error DrmHwcTwo::Init() {
   displays_.at(HWC_DISPLAY_PRIMARY).Init(disable_explicit_sync_);
 
   // Start the hwc service
-  hwcService_.Start(*this);
+  // FIXME: On Android, with userdebug on Joule this is causing
+  //        to hang in a loop while cold boot before going to home screen
+  // hwcService_.Start(*this);
 
   return HWC2::Error::None;
 }


Kishore409:
@pyn1 @kalyankondapally

os/android/drmhwctwo.cpp
```diff
@@ -95,7 +95,7 @@ HWC2::Error DrmHwcTwo::Init() {
   displays_.at(HWC_DISPLAY_PRIMARY).Init(disable_explicit_sync_);
 
   // Start the hwc service
-  hwcService_.Start(*this);
+  // hwcService_.Start(*this);
```
kalyankondapally: Add Jira and Test to commit message. Can u add FIXME here with comments what needs to be tested before enabling it again ? Can u create a Jira for this also, so we can track the work to enable this again

pyn1: sure.
Thanks,
Poornima

kalyankondapally:
merged

os/android/drmhwctwo.cpp
pyn1: Just synced with Kishore, looks like the issue is with latest github code base. The issue was never observed when I had pushed the patch. 
I will sync latest code to reproduce the issue.

PR #90, opened 2017-07-04
Linux: correct logic for get fd for multiple planes format

js0701:
Each plane should have it is own fd instead of only use
the first plane fd. Multiple plane format include NV12,
YUV420 etc


lathaanushareddy84:
Any changes needed in Minigbm ?

os/linux/gbmbufferhandler.cpp
```diff
@@ -115,7 +115,8 @@ bool GbmBufferHandler::ReleaseBuffer(HWCNativeHandle handle) {
       gbm_bo_destroy(handle->imported_bo);
     }
 #ifdef USE_MINIGBM
-    close(handle->import_data.fds[0]);
```
lathaanushareddy84: any change needed in Android side ?

os/linux/gbmbufferhandler.cpp
```diff
@@ -85,9 +85,9 @@ bool GbmBufferHandler::CreateBuffer(uint32_t w, uint32_t h, int format,
   temp->import_data.height = gbm_bo_get_height(bo);
   temp->import_data.format = gbm_bo_get_format(bo);
 #if USE_MINIGBM
-  temp->import_data.fds[0] = gbm_bo_get_plane_fd(bo, 0);
   size_t total_planes = gbm_bo_get_num_planes(bo);
   for (size_t i = 0; i < total_planes; i++) {
+    temp->import_data.fds[i] = gbm_bo_get_plane_fd(bo, i);
```
kalyankondapally: I dont see any other changes, what kind of changes are needed when sampling from texture ? this change by itself should not have any impact right ?

js0701: yes. Sampling from NV12 Mesa already support it

kalyankondapally: I mean within HWC, what does this change ? Place where we sample we use prime fd coming from hwcbuffer

js0701: gbm_bo_import previoulsy because some of the plane fds is 0. So this commit fix the problem

js0701: gbm_bo_import will return error due to fd of later planes is 0. This commit fix it

js0701: int drv_prime_bo_import(struct bo *bo, struct drv_import_fd_data *data)
{
  int ret;
  size_t plane;
  struct drm_prime_handle prime_handle;

  for (plane = 0; plane < bo->num_planes; plane++) {
    memset(&prime_handle, 0, sizeof(prime_handle));
    prime_handle.fd = data->fds[plane];

    ret = drmIoctl(bo->drv->fd, DRM_IOCTL_PRIME_FD_TO_HANDLE, &prime_handle);



so plane fds can't be set to zero

js0701:
DRM_IOCTL_PRIME_FD_TO_HANDLE failed (fd=0) Will get this error if we don't set fds of each plane

os/linux/gbmbufferhandler.cpp
kalyankondapally: @js0701 Can you check what needs to be done here too:
https://github.com/intel/IA-Hardware-Composer/blob/master/os/android/utils_android.h#L165

os/android/utils_android.h
```diff
@@ -114,18 +214,30 @@ static void DestroyBufferHandle(HWCNativeHandle handle) {
...
+static bool CreateGraphicsBuffer(uint32_t w, uint32_t h, int format,
+                                 HWCNativeHandle *handle, uint32_t usage) {
   struct gralloc_handle *temp = new struct gralloc_handle();
-  uint32_t usage =
+  uint32_t gralloc_usage =
       GRALLOC_USAGE_HW_FB | GRALLOC_USAGE_HW_RENDER | GRALLOC_USAGE_HW_COMPOSER;
-  if (cursor_usage) {
-    usage |= GRALLOC_USAGE_CURSOR;
+  if (usage | HW_BUFFER_USE_CURSOR) {
+    gralloc_usage |= GRALLOC_USAGE_CURSOR;
```
kalyankondapally: Wondering why we need this, can we make these implicit assumptions about Tiling/Linear within MINIGBM based on formats.

PR #91, opened 2017-07-07
Close the fence duped from surfacefligner

sophy228:
such fence is passed from SurfaceFlinger, we dup it here,
but no one closes the orignal fence which make
"fd leak" of SurfaceFlinger.

Change-Id: I7c04438fe6326e1de2e7345942afef33d582bb69
Tracked-On: https://jira01.devtools.intel.com/browse/OAM-43936
Signed-off-by: Zhongmin Wu <zhongmin.wu@intel.com>

### DIFF
--- a/os/android/iahwc2.cpp
+++ b/os/android/iahwc2.cpp
@@ -746,7 +746,7 @@ HWC2::Error IAHWC2::Hwc2Layer::SetLayerBuffer(buffer_handle_t buffer,
   native_handle_.handle_ = buffer;
   hwc_layer_.SetNativeHandle(&native_handle_);
   if (acquire_fence > 0)
-    hwc_layer_.SetAcquireFence(dup(acquire_fence));
+    hwc_layer_.SetAcquireFence(acquire_fence);
   return HWC2::Error::None;
 }
 


tpalli:
is this patch still relevant/required?

sophy228:
sorry, the patch is needed to fix the fd leaks of surfaceflinger.   But I see the file drmhwctwo.cpp is renamed to iahwc2.cpp,  so do I need to create a new patch and new pull request ?

tpalli:
Yes, please rebase the patch. BTW. is this dependent on any changes (like upcoming Mesa changes) or does this happen already with the current pipeline?

sophy228:
Well, current mesa did not implement the acquire fence actually, so the issue is not exposed.  However we are working for the acquire fence on mesa.

tpalli:
right ok, so this will be needed when mesa patches are in

kalyankondapally:
closing this, reopen if needed

kalyankondapally:
Looks like Mesa changes have landed and this change is needed

os/android/iahwc2.cpp
```diff
@@ -745,8 +745,10 @@ HWC2::Error IAHWC2::Hwc2Layer::SetLayerBuffer(buffer_handle_t buffer,
 
   native_handle_.handle_ = buffer;
   hwc_layer_.SetNativeHandle(&native_handle_);
-  if (acquire_fence > 0)
+  if (acquire_fence > 0) {
     hwc_layer_.SetAcquireFence(dup(acquire_fence));
+    close(acquire_fence);
```
kalyankondapally: We could avoid dup and close here. Just pass the acquire_fence i.e  hwc_layer_.SetAcquireFence(acquire_fence);

This should be enough. I can revise this and merge

sophy228: Done. The patch is changed as below

yogesh-marathe:
@kalyankondapally yes that's right, this is needed now. Thanks.

PR #96, opened 2017-07-17
Remove unneeded break if encoder is found

harishkrupo:
When an encoder is found for the first connector, the
variable found_encoder is set to true which then causes
the loop to terminate unnecessarily. This causes only one
connector to be detected at boot time.

Jira: None
Tests: On boot, more than one connected displays are identified.


kalyankondapally:
thx merged

PR #97, opened 2017-07-18
Fix tearing issue in multi display mode

harishkrupo:
Jira: None
Tests: No tearing is observed on all the displays


kalyankondapally:
merged thanks

PR #99, opened 2017-07-19
Update the HWC ebuild file to support .so and binary installation for Chrome OS.

yugang:
Latest HWC will install shared lib .so files and testlayers binary
will be generated after the "make install", to align with latest
HWC source, this commit will support .so files and testlayers
installation(not install test json example files).

Jira: None
Tests: Build HWC in Chrome OS SDK with this ebuild file without
       error, and deploy the HWC package to Chrome OS device
       without error, the testlayers can be executed normally
       with the shared the libhwcomposer.so.* installed.

### DIFF
--- a/os/chromeos/IA-Hardware-Composer-9999.ebuild
+++ b/os/chromeos/IA-Hardware-Composer-9999.ebuild
@@ -3,7 +3,8 @@
 # $Header: $
 
 EAPI=5
-inherit cros-workon eutils autotools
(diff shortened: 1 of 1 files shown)


yugang:
@kalyankondapally @js0701 pls help review

kalyankondapally:
lgtm

kalyankondapally:
merged

PR #104, opened 2017-08-01
Fix hang during boot up

harishkrupo:
Jira: None
Tests: Boots to home screen.


kalyankondapally:
merged

PR #105, opened 2017-08-02
Android: Build with vulkan source files

gaochun:
Build with vulkan backend source files and add corlor format mapping from gbm to vulkan

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -73,7 +73,8 @@ LOCAL_SRC_FILES := \
 	wsi/drm/drmdisplaymanager.cpp \
 	wsi/drm/drmscopedtypes.cpp \
         os/android/iahwc2.cpp \
-        os/android/hwcservice.cpp
+	os/android/hwcservice.cpp \
+	os/android/platformdefines.cpp
 
 ifeq ($(strip $(BOARD_USES_GRALLOC1)), true)
 LOCAL_SRC_FILES += os/android/gralloc1bufferhandler.cpp
--- a/common/compositor/vk/vkshim.cpp
+++ b/common/compositor/vk/vkshim.cpp
@@ -14,7 +14,6 @@
 
 #include "vkshim.h"
 
-#include <gbm.h>
 #include "hwctrace.h"
 
 namespace hwcomposer {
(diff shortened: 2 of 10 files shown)


gaochun:
@strassek Pls help review the patches.

Android.mk
```diff
@@ -98,7 +98,27 @@ ifeq ($(strip $(BOARD_DISABLE_NATIVE_COLOR_MODES)),true)
 LOCAL_CPPFLAGS += -DDISABLE_NATIVE_COLOR_MODES
 endif
 
-ifeq ($(strip $(BOARD_USES_VULKAN)),)
+ifeq ($(strip $(BOARD_USES_VULKAN)),true)
+LOCAL_SHARED_LIBRARIES += \
+        libvulkan
```
strassek: I think this file has been using tabs instead of spaces, might as well be consistent...

common/compositor/vk/vkshim.h
```diff
@@ -16,6 +16,7 @@
 #define VKSHIM_H_
 
 #include <drm/drm_fourcc.h>
+#include <gbm.h>
```
strassek: Is this needed outside of vkshim.cpp? Either way, It seems like this change belongs in the other patch...

kalyankondapally: Any gbm dependencies should be moved to os layer

kalyankondapally: Any gbm dependencies should be moved to os layer

gaochun: Compile failed without including gbm.h, I'm moving it to a better place.

kalyankondapally: You can add it to platformdefines.h and include that here also I see we are including gbm conversions here we need to figure out a way to isolate them

strassek: @gaochun Just had a chat about this with Kalyan, it seems like the most appropriate thing to do for now would be to rename this Vulkan format conversion function to something more generic (maybe NativeToVkFormat), and allow each platform to define their own version. So there would be version in os/linux/platformdefines.h that would do this GBM->VK conversion and a version in os/android/platfomrdefines.h that would do the appropriate thing for Android.

kalyankondapally:
@gaochun Any update ?

common/compositor/vk/vkshim.h
gaochun: I've renamed GbmToVkFormat to NativeToVkFormat and moved to platfomrdefines.h. But I found DrmToVkFormat was not used at any place , I think GbmToVkFormat should be replaced with DrmToVkFormat at https://github.com/intel/IA-Hardware-Composer/blob/master/wsi/drm/drmbuffer.cpp#L151,  right?

kalyankondapally: yes, makes sense

strassek: > GbmToVkFormat should be replaced with DrmToVkFormat

Almost, we should replace that with a call to NativeToVkFormat. There shouldn't be any need for DrmToVkFormat or GbmToVkFormat after this conversion.

common/compositor/vk/vkshim.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include "vkshim.h"
 
+#include <gbm.h>
```
kalyankondapally: remove this, add platformdefines.h

os/android/platformdefines.cpp
```diff
@@ -0,0 +1,67 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+*/
+
+#include "platformdefines.h"
+#include <drm/drm_fourcc.h>
+
+#ifdef USE_VK
+VkFormat NativeToVkFormat(int drm_format) {
```
kalyankondapally: this can be generic under os/

gaochun:
@kalyankondapally @strassek I deleted my forked repo by mistake, so could not update the commit here, I've update the patch with a new PR https://github.com/intel/IA-Hardware-Composer/pull/108 , please review again.

PR #109, opened 2017-08-09
Disable SetCursorPosition usage

harishkrupo:
According to the spec setting the cursor composition to
device disables the use SetCursorPosition by SurfaceFlinger.
For now we are not using this so inform SurfaceFlinger
about the same

Jira: None
Test: System doesn't crash when moving the mouse


kalyankondapally:
merged

PR #110, opened 2017-08-10
 Change the init_frames place to load test json firstly

yugang:
to fix the bug which can't load parameters before setting, e.g. color correction parameters, power mode parameters.


yugang:
only the last commit is for this PR

kalyankondapally:
can u rebase on tot ?

PR #120, opened 2017-09-08
Add implementation for grallochandler Map, Unmap functions

JeevakaPrabu:
Change-Id: Ic2377b02413f7536784cd2560e16bedfc4426992
Tracked-On: None
Signed-off-by: Jeevaka Prabu Badrappan <jeevaka.badrappan@intel.com>

### DIFF
--- a/common/compositor/compositorthread.cpp
+++ b/common/compositor/compositorthread.cpp
@@ -140,6 +140,7 @@ void CompositorThread::HandleReleaseRequest() {
   }
 
   if (release_all_resources_) {
+    gpu_resource_handler_->ReleaseGPUResources();
     plane_manager_->ReleaseAllOffScreenTargets();
     gpu_resource_handler_.reset(nullptr);
   } else {
--- a/os/android/iahwc2.cpp
+++ b/os/android/iahwc2.cpp
@@ -72,8 +72,9 @@ class IARefreshCallback : public hwcomposer::RefreshCallback {
 class IAHotPlugEventCallback : public hwcomposer::HotPlugCallback {
  public:
   IAHotPlugEventCallback(hwc2_callback_data_t data,
-                         hwc2_function_pointer_t hook)
-      : data_(data), hook_(hook) {
+                         hwc2_function_pointer_t hook,
+                         IAHWC2::HwcDisplay *display)
+      : data_(data), hook_(hook), display_(display) {
   }
 
   void Callback(uint32_t display, bool connected) {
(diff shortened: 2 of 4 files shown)


JeevakaPrabu:
@kalyankondapally @harishkrupo

os/android/gralloc1bufferhandler.cpp
```diff
@@ -129,15 +133,43 @@ void Gralloc1BufferHandler::CopyHandle(HWCNativeHandle source,
...
+  rect.top = y;
+  rect.width = width;
+  rect.height = height;
+
+  gralloc1_device_t *gralloc1_dvc =
+      reinterpret_cast<gralloc1_device_t *>(device_);
+  uint32_t status = lock_(gralloc1_dvc, handle->imported_handle_, GRALLOC_USAGE_SW_READ_RARELY,
+          GRALLOC_USAGE_SW_READ_RARELY, &rect, map_data, acquireFence);
+  ALOGI("%s - lock returned: %u", __FUNCTION__, status);
   return NULL;
```
kalyankondapally: retrun NULL here is confusing, probably we should share the API ?

kalyankondapally:
@yugang @js0701 @Kishore409 @harishkrupo  please review

os/android/gralloc1bufferhandler.cpp
harishkrupo: yeah, we can just return the status instead.

os/android/gralloc1bufferhandler.cpp
```diff
@@ -129,15 +133,43 @@ void Gralloc1BufferHandler::CopyHandle(HWCNativeHandle source,
...
+  auto gr_handle = (struct cros_gralloc_handle *)handle->imported_handle_;
+  if (!gr_handle) {
+    ETRACE("could not find gralloc drm handle");
+    return;
+  }
+
+  int releaseFence = 0;
+  gralloc1_device_t *gralloc1_dvc =
+      reinterpret_cast<gralloc1_device_t *>(device_);
+  unlock_(gralloc1_dvc, handle->imported_handle_, &releaseFence);
```
harishkrupo: Similarly we can return the status here.

os/android/gralloc1bufferhandler.cpp
js0701: I think the map api is to return the vritual interface CPU can access? That is at leat the story in Linux side. Linux is directly this API to gbm_bo_map

os/android/gralloc1bufferhandler.cpp
js0701: this API is inheriting from NativeBufferHandler. gbm_bo_umap is a void function too

kalyankondapally:
@JeevakaPrabu Are you going to update the full feature of dumping the buffers being sent to display ?

JeevakaPrabu:
@kalyankondapally : Dumping the buffers will be delivered as a separate patch. New pull request for this changes is https://github.com/intel/IA-Hardware-Composer/pull/132

PR #122, opened 2017-09-10
Track CursorLayer in hwc2.

kalyankondapally:
We change composition type of cursor to HWC2::Composition::Device
in ValidateDisplay. However, expectation in Present call is that
this is of type HWC2::Composition::Cursor. Instead of relying on
composition type, we keep track of cursor layer and ensure this is
having highest z_order.

Jira: None.
Test: Cursor is still visible when playing Video content.


kalyankondapally:
@harishkrupo ping

PR #124, opened 2017-09-12
Fix the external display not init when start Android UI.

yugang:
Jira: GSE-35
Test: Connect more than one display of Android platform and
      boot the device, the all display will show same cloned
      content.


kalyankondapally:
merged

PR #125, opened 2017-09-15
Fix connected boot in android.

harishkrupo:
Don't unset kNotifyClient state for the external display as we are not
going to notify Surface Flinger about it during RegisterHotPlugCallback.


harishkrupo:
@kalyankondapally please review.

kalyankondapally:
merged thanks

PR #126, opened 2017-09-18
Fixed the display pipe power off

pallavigopal:
DPMS off is enough to turn off the display.
with this patch Hotplug works fine.

Jira: None
Test: Hotplug works fine in Android. Tried 10-15 times.

### DIFF
--- a/wsi/drm/drmdisplay.cpp
+++ b/wsi/drm/drmdisplay.cpp
@@ -540,27 +540,10 @@ bool DrmDisplay::GetFence(drmModeAtomicReqPtr property_set,
 void DrmDisplay::Disable(const DisplayPlaneStateList &composition_planes) {
   IHOTPLUGEVENTTRACE("Disable: Disabling Display: %p", this);
 
-  ScopedDrmAtomicReqPtr pset(drmModeAtomicAlloc());
-  if (pset) {
-    bool active = false;
-    int ret = drmModeAtomicAddProperty(pset.get(), crtc_id_, active_prop_,
-                                       active) < 0;
-    if (ret) {
-      ETRACE("Failed to set display to inactive");
-    }
-
-    for (const DisplayPlaneState &comp_plane : composition_planes) {
-      DrmPlane *plane = static_cast<DrmPlane *>(comp_plane.plane());
-      plane->SetEnabled(false);
-      plane->Disable(pset.get());
-    }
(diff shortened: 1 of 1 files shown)


pallavigopal:
@kalyankondapally  Please review

wsi/drm/drmdisplay.cpp
```diff
@@ -540,27 +540,10 @@ bool DrmDisplay::GetFence(drmModeAtomicReqPtr property_set,
...
-    }
-
-    ret = drmModeAtomicCommit(gpu_fd_, pset.get(),
-                              DRM_MODE_ATOMIC_ALLOW_MODESET, NULL);
-    if (ret)
-      ETRACE("Failed to disable display:%s\n", PRINTERROR());
-  } else {
-    ETRACE("Failed to allocate property set %d", -ENOMEM);
+  for (const DisplayPlaneState &comp_plane : composition_planes) {
+    DrmPlane *plane = static_cast<DrmPlane *>(comp_plane.plane());
```
kalyankondapally: Which Scenario is this fixing hotplug ? External monitor works fine for me on APL and handles hotplugs correctly. While the patch makes sense just wondering what issue is it fixing.

pallavigopal: without this fix Hotplug is not working in KBL-RVP. The ddb is set to 0 for the secondary display.
dpms off is enough to disable the display no need to call atomic call with all the properties set to 0.

pallavigopal: Even display suspend/resume works fine with this patch.

kalyankondapally: I think we can probably skip this whole "disable" part in drmdisplay when monitor is physically disconnected. So, for powermode changes, ya we should still call this and setting DPMS_OFF should be enough. Incase of physical disconnection we dont need to even call this. Probably in Physical Display https://github.com/intel/IA-Hardware-Composer/blob/master/wsi/physicaldisplay.cpp#L115 we can change this to a new entry point in DisplayQueue. That function in DisplayQueue would call HandleExit. HandleExit could take in a boolean to represent if we want to call the disable or not ?

kalyankondapally: Merged this patch, can you send a follow up patch with above mentioned change

PR #128, opened 2017-09-20
Fix Display frame change check

js0701:
Include top/bottom to detect display frame change

Jira: OAM-49818
Tests: WebView Browser Tester with number keyboard


kalyankondapally:
merged

PR #130, opened 2017-09-20
Android: Add setColorTransform to HWC2 to enable color correction

gaochun:
Get 4x4 color transform matrix (CTM) from SurfaceFlinger, and convert
the matrix to 3x3 DRM CSC matrix.

Jira: OAM-49721
Test: Go to Settings-Accessibility-Color correction, set it to On
and choose the Correction mode.

### DIFF
--- a/common/display/displayqueue.cpp
+++ b/common/display/displayqueue.cpp
@@ -51,6 +51,8 @@ DisplayQueue::DisplayQueue(uint32_t gpu_fd, bool disable_overlay,
 
   /* use 0x80 as default brightness for all colors */
   brightness_ = 0x808080;
+  /* use HWCColorTransform::kIdentical as default color transform hint */
+  color_transform_hint_ = HWCColorTransform::kIdentical;
   /* use 0x80 as default brightness for all colors */
   contrast_ = 0x808080;
   /* use 1 as default gamma value */
@@ -438,6 +440,7 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
     render_layers = display_plane_manager_->ValidateLayers(
         layers, cursor_layers, state_ & kConfigurationChanged,
         idle_frame || disable_ovelays, current_composition_planes);
+
     state_ &= ~kConfigurationChanged;
     bool gpu_rendered = true;
     for (uint32_t i = 0; i < total_cursor_layers_; i++) {
@@ -491,7 +494,8 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
   }
 #endif
   if (state_ & kNeedsColorCorrection) {
-    display_->SetColorCorrection(gamma_, contrast_, brightness_);
+    display_->SetColorCorrection(gamma_, contrast_, brightness_,
+                                 color_transform_matrix_, color_transform_hint_);
     state_ &= ~kNeedsColorCorrection;
   }
 
@@ -712,6 +716,16 @@ void DisplayQueue::SetGamma(float red, float green, float blue) {
   gamma_.red = red;
   gamma_.green = green;
   gamma_.blue = blue;
+  state_ |= kNeedsColorCorrection;
+}
+
+void DisplayQueue::SetColorTransform(const float *matrix, HWCColorTransform hint) {
+  color_transform_hint_ = hint;
+
+  if (hint == HWCColorTransform::kArbitraryMatrix) {
+    memcpy(color_transform_matrix_, matrix, sizeof(color_transform_matrix_));
+  }
+
   state_ |= kNeedsColorCorrection;
 }
 
--- a/public/hwcdefs.h
+++ b/public/hwcdefs.h
@@ -71,5 +71,10 @@ enum DisplayPowerMode {
                     // updates from the client
 };
 
+enum HWCColorTransform {
+  kIdentical = 0,       // Applies no transform to the output color
+  kArbitraryMatrix = 1  // Applies an arbitrary transform defined by a 4x4 affine matrix
+};
+
 }  // namespace hwcomposer
 #endif  // PUBLIC_HWCDEFS_H_
--- a/public/nativedisplay.h
+++ b/public/nativedisplay.h
@@ -141,6 +141,33 @@ class NativeDisplay {
   */
   virtual void SetGamma(float /*red*/, float /*green*/, float /*blue*/) {
   }
+
+  /**
+  * API for setting a color transform which will be applied after composition.
+  *
+  * The matrix provided is an affine color transformation of the following form:
+  *
+  * |r.r r.g r.b 0|
+  * |g.r g.g g.b 0|
+  * |b.r b.g b.b 0|
+  * |Tr  Tg  Tb  1|
+  *
+  * This matrix will be provided in row-major form: {r.r, r.g, r.b, 0, g.r, ...}.
+  *
+  * Given a matrix of this form and an input color [R_in, G_in, B_in], the output
+  * color [R_out, G_out, B_out] will be:
+  *
+  * R_out = R_in * r.r + G_in * g.r + B_in * b.r + Tr
+  * G_out = R_in * r.g + G_in * g.g + B_in * b.g + Tg
+  * B_out = R_in * r.b + G_in * g.b + B_in * b.b + Tb
+  *
+  * @param matrix a 4x4 transform matrix (16 floats) as described above
+  * @param hint a hint value to specify the transform type, applying no transform
+  *        or applying transform defined by given matrix
+  */
+  virtual void SetColorTransform(const float * /*matrix*/, HWCColorTransform /*hint*/) {
+  }
+
   /**
   * API for setting display color contrast in HWC
   * @param red valid value is 0 ~ 255, bigger value with stronger contrast
--- a/wsi/drm/drmdisplay.cpp
+++ b/wsi/drm/drmdisplay.cpp
@@ -60,6 +60,8 @@ bool DrmDisplay::InitializeDisplay() {
       drmModeObjectGetProperties(gpu_fd_, crtc_id_, DRM_MODE_OBJECT_CRTC));
   GetDrmObjectProperty("ACTIVE", crtc_props, &active_prop_);
   GetDrmObjectProperty("MODE_ID", crtc_props, &mode_id_prop_);
+  GetDrmObjectProperty("CTM", crtc_props, &ctm_id_prop_);
+  GetDrmObjectProperty("CTM_POST_OFFSET", crtc_props, &ctm_post_offset_id_prop_);
   GetDrmObjectProperty("GAMMA_LUT", crtc_props, &lut_id_prop_);
   GetDrmObjectPropertyValue("GAMMA_LUT_SIZE", crtc_props, &lut_size_);
   GetDrmObjectProperty("OUT_FENCE_PTR", crtc_props, &out_fence_ptr_prop_);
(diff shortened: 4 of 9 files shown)


wsi/drm/drmdisplay.cpp
```diff
@@ -425,19 +452,68 @@ float DrmDisplay::TransformGamma(float value, float gamma) const {
...
+    case 0: { // HAL_COLOR_TRANSFORM_IDENTITY
+      memset(ctm->matrix, 0, sizeof(ctm->matrix));
+      for (int i = 0; i < 3; i++) {
+        ctm->matrix[i * 3 + i] = (1ll << 32);
+      }
+      ApplyPendingCTM(ctm);
+      break;
+    }
+    case 1: { // HAL_COLOR_TRANSFORM_ARBITRARY_MATRIX
+      // Extract the coefficients from 4x4 Android HWC2 CTM to the drm 3x3 CTM.
```
kalyankondapally: Probably define our own color hint enums and in iahwc2, translate them to our enum and pass it here, instead of case 1: 

Since our API also expects the same 4x4 matrix,  probably you can generalize this instead of saying "Android HWC2"

common/display/displayqueue.cpp
```diff
@@ -51,6 +51,8 @@ DisplayQueue::DisplayQueue(uint32_t gpu_fd, bool disable_overlay,
 
   /* use 0x80 as default brightness for all colors */
   brightness_ = 0x808080;
+  /* use HAL_COLOR_TRANSFORM_IDENTITY as default color transform hint */
```
kalyankondapally: HWCColorTransform::kIdentical  here

public/hwcdefs.h
```diff
@@ -76,5 +76,10 @@ enum DisplayPowerMode {
                     // updates from the client
 };
 
+enum HWCColorTransform {
+  kIdentical = 0,
```
kalyankondapally: Add documentation what do these different values mean, do these cover all cases we support ?

public/nativedisplay.h
```diff
@@ -142,6 +142,33 @@ class NativeDisplay {
...
+  *
+  * Given a matrix of this form and an input color [R_in, G_in, B_in], the output
+  * color [R_out, G_out, B_out] will be:
+  *
+  * R_out = R_in * r.r + G_in * g.r + B_in * b.r + Tr
+  * G_out = R_in * r.g + G_in * g.g + B_in * b.g + Tg
+  * B_out = R_in * r.b + G_in * g.b + B_in * b.b + Tb
+  *
+  * @param matrix a 4x4 transform matrix (16 floats) as described above
+  * @param hint a hint value which may be used instead of the given matrix unless
```
kalyankondapally: enum HWCColorTransform should cover all hints we can support. Instead of in32_t here we should take in HWCColorTransform hint

wsi/drm/drmdisplay.cpp
```diff
@@ -425,14 +452,61 @@ float DrmDisplay::TransformGamma(float value, float gamma) const {
...
+    case HWCColorTransform::kIdentical: {
+      memset(ctm->matrix, 0, sizeof(ctm->matrix));
+      for (int i = 0; i < 3; i++) {
+        ctm->matrix[i * 3 + i] = (1ll << 32);
+      }
+      ApplyPendingCTM(ctm);
+      break;
+    }
+    case HWCColorTransform::kArbitraryMatrix: {
+      // Extract the coefficients from HWC2 4x4 CTM to the DRM 3x3 CTM. |Tr Tg Tb|
```
kalyankondapally: You can remove HWC2 term here.

public/hwcdefs.h
gaochun: The SurfaceFlinger only uses HAL_COLOR_TRANSFORM_IDENTITY and HAL_COLOR_TRANSFORM_ARBITRARY_MATRIX for color transform,  I think we can only support these 2 hints.

kalyankondapally: ya lets just do that than

wsi/physicaldisplay.h
```diff
@@ -116,8 +117,11 @@ class PhysicalDisplay : public NativeDisplay, public DisplayPlaneHandler {
   /**
   * API for setting color correction for display.
   */
-  virtual void SetColorCorrection(struct gamma_colors gamma, uint32_t contrast,
-                                  uint32_t brightness) const = 0;
+  virtual void SetColorCorrection(struct gamma_colors gamma,
+                                  uint32_t contrast,
```
kalyankondapally: hmm this will break Linux tests. I think we need to adapt Linux tests to use the new API and that way also test this actually

gaochun: Did not find SetColorCorrection used in Linux tests

kalyankondapally: hmm ya I thought we had some tests

kalyankondapally:
merged thanks

kalyankondapally:
@gaochun  Sorry, this broke Android O for some reason,. Can you check ?

gaochun:
@kalyankondapally I'm looking at this on Android O, did you see it's a compile issue or running issue?

kalyankondapally:
@gaochun it had wrong colors

kalyankondapally:
@yugang  @Kishore409 @harishkrupo @tpalli please review

gaochun:
@kalyankondapally I cherry-pick this patch to AndroidO default hwcomposer, it works right.

gaochun:
@kalyankondapally I see the color issue in my last AndroidO build with the latest HWC code, it seems a kernel/drm issue. After I update AndroidO to m0 r12, the issue could not be reproduced, the color correction and 3 options works correct. Please verify it with the latest AndroidO again.

kalyankondapally:
@gaochun I tested on couple of panels. On one seems results are good, while on other the colors where all wrong in settings app for example. Not sure what is going on,

wsi/physicaldisplay.h
yugang: @gaochun , public/nativedisplay.h, SetGamma(), SetContrast(), SetBrightness() are the APIs of Linux side, tests/jsonconfigs/colorcorrection.json is the test json. 

about SetColorCorrection(), ./wsi/drm/drmdisplay.cpp already had, seems some different with yours, let me confirm with you about details offline.

gaochun:
@kalyankondapally Patch update:
commit 1: Keep unchanged.
commit 2: New commit  for color inversion, fallback to gpu rendering
commit 3: Workaround for the color issue, it should be reverted once the drm ctm color issue get fixed.

common/display/displayqueue.cpp
```diff
@@ -713,22 +713,13 @@ void DisplayQueue::SetColorTransform(const float *matrix, HWCColorTransform hint
 
   if (hint == HWCColorTransform::kArbitraryMatrix) {
     memcpy(color_transform_matrix_, matrix, sizeof(color_transform_matrix_));
-    if (*(uint32_t*)(matrix + 12) != 0 ||
-        *(uint32_t*)(matrix + 13) != 0 ||
-        *(uint32_t*)(matrix + 14) != 0) {
       compositor_.SetColorTransformMatrix(color_transform_matrix_);
       state_ |= kDisableOverlayUsage;
```
kalyankondapally: hmm I dont think we need to disable overlay usage in this case. What we can do is force a GPU composition pass for layers in this case. You can do it by adding another state, POST_PROCESSING_FOR_COLOR_MATRIX and here https://github.com/intel/IA-Hardware-Composer/blob/master/common/display/displayqueue.cpp#L454, we can check if that state is set and if so change the state of all composition planes to be gpu rendered i.e. calling this API for all current composition planes https://github.com/intel/IA-Hardware-Composer/blob/master/common/display/displayplanestate.h#L122

Note that when this Matrix changes i.e in case Display can handle it straight away and dont need compositor to get involved, than we will need to force re-validation of layers. This can be done by setting state_ |= FrameStateTracker::kRevalidateLayers;

gaochun: @kalyankondapally Thanks, I'll enhance the implementation according to your comment, and I'll continue followup the feature in PRC holidays.

gaochun:
@kalyankondapally Commits updated, pls review again.

common/Android.mk
```diff
@@ -83,6 +83,7 @@ endif
 LOCAL_CPPFLAGS += \
         -DHWC2_INCLUDE_STRINGIFICATION \
         -DHWC2_USE_CPP11 \
+        -DUSE_DRM_CTM \
```
kalyankondapally: Should we have a BOARD Config to disable this ?
i.e ifneq ($(strip $(BOARD_DISABLE_DRM_CTM)), true)
LOCAL_CPPFLAGS += -DUSE_DRM_CTM
endif

common/compositor/gl/glprogram.cpp
```diff
@@ -231,6 +237,9 @@ void GLProgram::UseProgram(const RenderState &state, GLuint viewport_width,
               (state.width_) / (float)viewport_width,
               (state.height_) / (float)viewport_height);
 
+  glUniformMatrix4fv(ctm_loc_, 1, GL_FALSE, ctm_);
```
kalyankondapally: Do we need to set this always ? Can we have a flag to track any change to these values and if so do this.

wsi/Android.mk
```diff
@@ -64,6 +64,7 @@ endif
 LOCAL_CPPFLAGS += \
         -DHWC2_INCLUDE_STRINGIFICATION \
         -DHWC2_USE_CPP11 \
+        -DUSE_DRM_CTM \
```
kalyankondapally: Should we have a BOARD Config to disable this ?
i.e ifneq ($(strip $(BOARD_DISABLE_DRM_CTM)), true)
LOCAL_CPPFLAGS += -DUSE_DRM_CTM
endif

common/Android.mk
kalyankondapally: I see u revert this in next patch, instead let's have this board config, and selectively disable it on devices which show issues.

gaochun:
@kalyankondapally Add BOARD_DISABLE_DRM_CTM board config, and improved the invoking of glUniformMatrix4fv.

common/display/displayqueue.cpp
```diff
@@ -693,6 +704,41 @@ void DisplayQueue::SetGamma(float red, float green, float blue) {
...
+    compositor_.SetColorTransformMatrix(color_transform_matrix_);
+    state_ |= kApplyColorMatrix;
+#endif
+  } else {
+    compositor_.SetColorTransformMatrix(NULL);
+    state_ &= ~kApplyColorMatrix;
+  }
+
+  if (!(old_state & kApplyColorMatrix) && (state_ & kApplyColorMatrix) ||
+      (old_state & kApplyColorMatrix) && !(state_ & kApplyColorMatrix)) {
```
kalyankondapally: warning: '&&' within '||' [-Wlogical-op-parentheses]

kalyankondapally: Needs rebasing sorry :(

gaochun:
@kalyankondapally Rebased and fixed the warning.

gaochun:
@kalyankondapally Also add 2 lines to avoid settings getting invalid when device wake up from lock screen:
https://github.com/gaochun/IA-Hardware-Composer/blob/dd16ebd9a8c226d8f21b97b0d5b127411df8c0da/common/compositor/compositorthread.cpp#L259
https://github.com/gaochun/IA-Hardware-Composer/blob/dd16ebd9a8c226d8f21b97b0d5b127411df8c0da/common/display/displayqueue.cpp#L704

common/display/displayqueue.cpp
```diff
@@ -689,7 +701,7 @@ void DisplayQueue::HandleExit() {
     cloned_mode = true;
   }
 
-  state_ = kConfigurationChanged;
+  state_ = kConfigurationChanged | (state_ & kApplyColorMatrix);
```
kalyankondapally: We shouldn't always set this. Like for  disable_overlay below, we should check if this is part of state_ already before setting it again.

common/display/displayqueue.cpp
```diff
@@ -436,7 +438,16 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
     RecyclePreviousPlaneSurfaces();
     render_layers = display_plane_manager_->ValidateLayers(
         layers, cursor_layers, state_ & kConfigurationChanged,
-        idle_frame || disable_ovelays, current_composition_planes);
+        idle_frame || disable_ovelays || (state_ & kApplyColorMatrix),
```
kalyankondapally: We don't need this (state_ & kApplyColorMatrix) here as below we are forcing GPU Composition for different planes.

common/Android.mk
```diff
@@ -94,6 +94,10 @@ LOCAL_CPPFLAGS += \
         -D_GNU_SOURCE=1 -D_FILE_OFFSET_BITS=64 \
         -O3
 
+ifneq ($(strip $(BOARD_DISABLE_DRM_CTM)), true)
+LOCAL_CPPFLAGS += -DUSE_DRM_CTM
```
kalyankondapally: @Kishore409 Take note, We need to enable this flag BOARD_DISABLE_DRM_CTM in some of our Board Configs for APL. We don't need this flag for KBL or other boards.

common/display/displayqueue.cpp
gaochun: Android launcher wallpaper could not be correctly rendered after remove this, seems there is something wrong in layer calculation.

kalyankondapally: @gaochun  if some layer is getting messed up, let's try to understand which one. This should be fixed in hwc anyway.

common/Android.mk
kalyankondapally: @Kishore409 ping

Kishore409: Sure @kalyankondapally

common/display/displayqueue.cpp
gaochun: The wallpaper layer data is HAL_PIXEL_FORMAT_RGBX_8888 format, which will get wrong color after the ctm transform, I'm checking whether there is issue of processing alpha data.

gaochun: @kalyankondapally I took some time to enable gpu rendering of each plane while keeping overlay enabled, but there are still some issues and regressions to fix. As the color correction and inversion need to be enabled by  ww42 and I'm OOO util 10/19, can I  keep gpu compositing here and add a TODO tag first? I'll submit a new path to enable overlays after all issues get fixed.

kalyankondapally: @gaochun This broke APL platforms completely.

wsi/physicaldisplay.h
js0701: @yugang  @kalyankondapally 
What is the problem about this PR
We need to move forward

common/display/displayqueue.cpp
gaochun: @kalyankondapally Verified with @js0701 's Broxton device, the patch works, did you add "BOARD_DISABLE_DRM_CTM := true"  to BoardConfig.mk?

gaochun:
The update patch uses a new drm interface to set the ctm post offset. So it's depends on the kernel patch and the libdrm patch https://github.com/intel/external-libdrm/pull/6

common/display/displayqueue.cpp
```diff
@@ -438,6 +440,7 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
     render_layers = display_plane_manager_->ValidateLayers(
         layers, cursor_layers, state_ & kConfigurationChanged,
         idle_frame || disable_ovelays, current_composition_planes);
+
```
kalyankondapally: unnecessary

gaochun: Ya

PR #131, opened 2017-09-21
Added Exit function to handle the display Disconnect

pallavigopal:
Add Exit function in the displayqueue to handle the Display Disconnect and also
did small code cleanup in the display disable

Jira: None
Tests: Hotplug and suspend/resume works fine

### DIFF
--- a/common/display/displayqueue.cpp
+++ b/common/display/displayqueue.cpp
@@ -92,6 +92,16 @@ bool DisplayQueue::Initialize(uint32_t pipe, uint32_t width, uint32_t height,
   return true;
 }
 
+bool DisplayQueue::Exit(uint32_t power_mode) {
+  if (power_mode != kOff) {
+    HandleExit();
+  } else {
+    return false;
+  }
+
+  return true;
+}
+
 bool DisplayQueue::SetPowerMode(uint32_t power_mode) {
   switch (power_mode) {
     case kOff:
@@ -656,9 +666,7 @@ void DisplayQueue::HandleExit() {
   state_ |= kIgnoreIdleRefresh;
   power_mode_lock_.unlock();
   vblank_handler_->SetPowerMode(kOff);
-  if (!previous_plane_state_.empty()) {
-    display_->Disable(previous_plane_state_);
-  }
+  display_->Disable();
 
   std::vector<OverlayLayer>().swap(in_flight_layers_);
   DisplayPlaneStateList().swap(previous_plane_state_);
--- a/common/display/displayqueue.h
+++ b/common/display/displayqueue.h
@@ -54,6 +54,7 @@ class DisplayQueue {
 
   bool Initialize(uint32_t pipe, uint32_t width, uint32_t height,
                   DisplayPlaneHandler* plane_manager);
+  bool Exit(uint32_t power_mode);
 
   bool QueueUpdate(std::vector<HwcLayer*>& source_layers, int32_t* retire_fence,
                    bool idle_update);
--- a/wsi/drm/drmdisplay.cpp
+++ b/wsi/drm/drmdisplay.cpp
@@ -537,14 +537,8 @@ bool DrmDisplay::GetFence(drmModeAtomicReqPtr property_set,
   return true;
 }
 
-void DrmDisplay::Disable(const DisplayPlaneStateList &composition_planes) {
+void DrmDisplay::Disable() {
   IHOTPLUGEVENTTRACE("Disable: Disabling Display: %p", this);
-
-  for (const DisplayPlaneState &comp_plane : composition_planes) {
-    DrmPlane *plane = static_cast<DrmPlane *>(comp_plane.plane());
-    plane->SetEnabled(false);
-    plane->SetNativeFence(-1);
-  }
 
   drmModeConnectorSetProperty(gpu_fd_, connector_, dpms_prop_,
                               DRM_MODE_DPMS_OFF);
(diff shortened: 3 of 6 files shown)


pallavigopal:
@kalyankondapally  Please review

wsi/physicaldisplay.cpp
```diff
@@ -110,21 +110,16 @@ void PhysicalDisplay::DisConnect() {
 
   display_state_ &= ~kDisconnectionInProgress;
 
-  if (!(display_state_ & kConnected)) {
+  if (!display_queue_->Exit(power_mode_)) {
```
kalyankondapally: We can just leave like now.

wsi/physicaldisplay.cpp
```diff
@@ -110,21 +110,16 @@ void PhysicalDisplay::DisConnect() {
...
   }
+
   IHOTPLUGEVENTTRACE(
       "PhysicalDisplay DisConnect called for Display: %p hotplugdisplayid: %d "
       "\n",
       this, hot_plug_display_id_);
   display_state_ |= kNotifyClient;
-
-  if (power_mode_ != kOff) {
-    display_queue_->SetPowerMode(kOff);
```
kalyankondapally: Here we can call Exit API. It doesn't need to take in powermode, it would just mean display has been physically disconnected

wsi/drm/drmdisplay.cpp
```diff
@@ -537,15 +537,9 @@ bool DrmDisplay::GetFence(drmModeAtomicReqPtr property_set,
   return true;
 }
 
-void DrmDisplay::Disable(const DisplayPlaneStateList &composition_planes) {
+void DrmDisplay::Disable() {
   IHOTPLUGEVENTTRACE("Disable: Disabling Display: %p", this);
 
-  for (const DisplayPlaneState &comp_plane : composition_planes) {
-    DrmPlane *plane = static_cast<DrmPlane *>(comp_plane.plane());
```
kalyankondapally: Good to reset the state neverthless, so I think we want to do this.

common/display/displayqueue.h
```diff
@@ -54,6 +54,7 @@ class DisplayQueue {
 
   bool Initialize(uint32_t pipe, uint32_t width, uint32_t height,
                   DisplayPlaneHandler* plane_manager);
+  bool Exit(uint32_t power_mode);
```
kalyankondapally: No need of power_mode, probably make it more descriptive of what it means when this is called. HandleDisconnect ?

common/display/displayqueue.cpp
```diff
@@ -656,9 +666,7 @@ void DisplayQueue::HandleExit() {
   state_ |= kIgnoreIdleRefresh;
   power_mode_lock_.unlock();
   vblank_handler_->SetPowerMode(kOff);
-  if (!previous_plane_state_.empty()) {
-    display_->Disable(previous_plane_state_);
-  }
+  display_->Disable();
```
kalyankondapally: Goal was that we dont even need to call this when display is disconnected right ?

wsi/physicaldisplay.cpp
pallavigopal: But you only asked me to add the entry point in display_queue to do the disable. I hope my understanding is correct.

pallavigopal: Then I will add this cleanup in the disable path. Is that fine?

wsi/physicaldisplay.cpp
pallavigopal: Ok I will add this exit here and rename into HandleDisconnect()

common/display/displayqueue.h
pallavigopal: do

wsi/drm/drmdisplay.cpp
pallavigopal: After this disable call the previous_plane_state_ list is swapped so removed this state reset.

common/display/displayqueue.cpp
pallavigopal: I mainly called to keep all the drm related calls in the DrmDisplay otherwise we need to move the dpms off property set to displayqueue.

kalyankondapally: I mean you dont need to call it in this particular case. HandleExit can take a boolean based on which you can call this. If display is disconnected than we dont need to call it i.e. send false, in other cases call when if (!previous_plane_state_.empty())  is true

wsi/drm/drmdisplay.cpp
kalyankondapally: Ya the planes will be left with previous state which will not be valid anymore, hence I think it's good to reset the state here.

common/display/displayqueue.cpp
pallavigopal: you mean no need to call dpms  off during disconnect.

wsi/drm/drmdisplay.cpp
pallavigopal: no more used then why to reset the state not going to impact anything. I tested also

common/display/displayqueue.cpp
kalyankondapally: ya

pallavigopal: then how display will be turned off if we don't call dpms off.

kalyankondapally: When display is disconnected, it's already handled by kernel. We only need to handle it when power mode is off

pallavigopal: No kernel is not turning off the display. If I comment the disable part, after unplug also I see pipe/plane is on.

kalyankondapally: hmm k than this change doesn't make much sense

PR #133, opened 2017-09-27
resolve HWC2_VSYNC_DISABLE correctly

kalyankondapally:
Jira: None.
Test: adb shell stop/start no regression.

Change-Id: Ife0eae217b648d859637c38e419361df24ed8fcc

### DIFF
--- a/os/android/iahwc2.cpp
+++ b/os/android/iahwc2.cpp
@@ -726,7 +726,17 @@ HWC2::Error IAHWC2::HwcDisplay::SetPowerMode(int32_t mode_in) {
 
 HWC2::Error IAHWC2::HwcDisplay::SetVsyncEnabled(int32_t enabled) {
   supported(__func__);
-  display_->VSyncControl(enabled);
+  switch (enabled) {
+    case HWC2_VSYNC_ENABLE:
+      display_->VSyncControl(true);
+      break;
+    case HWC2_VSYNC_DISABLE:
+      display_->VSyncControl(false);
+      break;
+    default:
+      ALOGE("SetVsyncEnabled called with invalid parameter");
+      return HWC2::Error::BadParameter;
+  }
   return HWC2::Error::None;
 }
 


kalyankondapally:
@Kishore409 ping

os/android/iahwc2.cpp
```diff
@@ -726,7 +726,17 @@ HWC2::Error IAHWC2::HwcDisplay::SetPowerMode(int32_t mode_in) {
 
 HWC2::Error IAHWC2::HwcDisplay::SetVsyncEnabled(int32_t enabled) {
   supported(__func__);
-  display_->VSyncControl(enabled);
+  switch (enabled) {
+    case HWC2_VSYNC_ENABLE:
+      display_->VSyncControl(true);
```
kalyankondapally: hmm Now I can reproduce the bug.

kalyankondapally:
@mwang30  I see all services "died" but no body seems to inform hwc to shut down

mwang30:
@kalyankondapally Please see more details in file:
hardware/interfaces/graphics/composer/2.1/default/ComposerClient.cpp

143ComposerClient::~ComposerClient()
144{
145    // We want to call hwc2_close here (and move hwc2_open to the
146    // constructor), with the assumption that hwc2_close would
147    //
148    //  - clean up all resources owned by the client
149    //  - make sure all displays are blank (since there is no layer)
150    //
151    // But since SF used to crash at this point, different hwcomposer2
152    // implementations behave differently on hwc2_close.  Our only portable
153    // choice really is to abort().  But that is not an option anymore
154    // because we might also have VTS or VR as clients that can come and go.
155    //
156    // Below we manually clean all resources (layers and virtual
157    // displays), and perform a presentDisplay afterwards.

kalyankondapally:
@mwang30 so what we should be doing ?

mwang30:
Since all resources are cleaned, we don't need to do anything.

kalyankondapally:
Merged this to our next branch. Please make sure this is working as expected on Android O. Otherwise we will have to revert this patch.

PR #134, opened 2017-10-01
Added function to get Graffic Buffer Info

munish-b:


os/android/utils_android.h
```diff
@@ -118,6 +118,7 @@ static uint32_t DrmFormatToHALFormat(int format) {
     case DRM_FORMAT_XBGR8888:
     case DRM_FORMAT_RGBX8888:
     case DRM_FORMAT_ABGR8888:
+      return HAL_PIXEL_FORMAT_RGBA_8888;
```
tpalli: this seems wrong, for all of these formats we return same format?

munish-b: I will update. Thanks

munish-b:
I will add public function to get Buffer Information.

tpalli:
This sounds like API for gralloc (minigbm) and not hwcomposer (?) Which component will use the API?

munish-b:
You are correct this API is needed to get information of Allocated Graphics Buffers.
Graphics buffers are being allocated using "new android::GraphicBuffer(width, height, format, usage)" and later buffer information is used to allocate additional buffers for internal validation purposes. I need to check if I can directly use CreateBuffer and ImportBuffer to avoid this change.

kalyankondapally:
@tpalli This is for hwc-val test suite and we want to keep the gralloc/gbm backend transparent in the test suite.
@munish-b Ya, you want to check existing API, instead of adding new ones.

PR #136, opened 2017-10-09
Ensure VA processing the buffer with correct crop/frame

XiaosongWei:
The surface region for VA input must be from layer's SourceCrop
and the output region for VA output must be from layer's DisplayFrame


XiaosongWei:
@kalyankondapally 
It now works with the vaapi on Android-IA, please take a look...

kalyankondapally:
merged thanks. Can you look into Android O, so we can enable this by default.

PR #137, opened 2017-10-10
Enable build testlayers for Android with hwc's static lib

XiaosongWei:
As Android build system doesn't recognize shared library name like
hwcomposer.xx.so, previously in order to build testlayers we have to
modify Android.mk to rename hwcomposer.xx.so to libhwcomposer.xx.so.
This patch provides an easy way to build testlayers without renaming
the library name back and forth.

### DIFF
--- a/Android.static.mk
+++ b/Android.static.mk
@@ -1,0 +1,161 @@
+# Copyright (c) 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+ifeq ($(strip $(BOARD_USES_IA_HWCOMPOSER)), true)
+# Obtain root HWC source path
+HWC_PATH := $(call my-dir)
+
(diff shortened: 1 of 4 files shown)


XiaosongWei:
@kalyankondapally PTAL

Android.static.mk
```diff
@@ -0,0 +1,161 @@
...
+HWC_VERSION_GIT_BRANCH := $(shell pushd $(HWC_PATH) > /dev/null; git rev-parse --abbrev-ref HEAD; popd > /dev/null)
+HWC_VERSION_GIT_SHA := $(shell pushd $(HWC_PATH) > /dev/null; git rev-parse HEAD; popd > /dev/null)
+
+LOCAL_PATH := $(call my-dir)
+include $(CLEAR_VARS)
+
+LOCAL_WHOLE_STATIC_LIBRARIES := \
+        libhwcomposer_common \
+        libhwcomposer_wsi
+
```
kalyankondapally: @XiaosongWei  nice, wondering if we should move all common stuff to https://github.com/intel/IA-Hardware-Composer/blob/master/Android.common.mk and only have this and other Android.mk build the appropriate libs ?

XiaosongWei: @kalyankondapally Android.common.mk is NOT used by any Android.mk so far.

XiaosongWei: Let's have this landed firstly and improve the needed *.mk later on.

kalyankondapally: It's better to squash the un-needed ones first.

XiaosongWei: @kalyankondapally It makes me confused.. My change doesn't involve any un-needed ones, this patch is intended to build testlayers easily. The Android.common.mk you mentioned above is not related to this patch.

kalyankondapally: @XiaosongWei  Understood. Not your fault :) What I was saying was we will now have three files duplicating the same. Instead of adding a new file yet again, let's try to use the available files (i.e. Android.common.mk) to include the common things.

PR #141, opened 2017-10-13
Add config handling for logical and mosiac.

yugang:
The HWC will read the display config for logical and mosiac, the
config file will be pointed out by ENV var HWC_DISPLAY_CONFIG, and
use default path /etc/hwc_display.ini if no HWC_DISPLAY_CONFIG found.

Pls see comments of the config for detailed config format and usages.

Jira: None.
Test: Change the config file and connect some displays, the cube test
      can show logical and mosiac normally according to the config
      parameters.

Signed-off-by: Fan, Yugang <yugang.fan@intel.com>

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -41,126 +41,179 @@ bool GpuDevice::Initialize() {
 
   std::vector<NativeDisplay *> displays = display_manager_->GetAllDisplays();
   size_t size = displays.size();
-  NativeDisplay *primary_display = NULL;
-  for (size_t i = 0; i < size; i++) {
-    hwcomposer::NativeDisplay *display = displays.at(i);
-    if (display->IsConnected()) {
-      primary_display = display;
-      break;
-    }
+
+  // Handle config file reading
+  const char *hwc_dp_cfg_path = std::getenv("HWC_DISPLAY_CONFIG");
+  if (!hwc_dp_cfg_path) {
+    hwc_dp_cfg_path = "/etc/hwc_display.ini";
   }
 
-  if (!primary_display) {
-    primary_display = displays.at(0);
-  }
-
-  // TODO: How do we determine when to use Logical Manager ?
   bool use_logical = false;
   bool use_mosiac = false;
-  std::vector<NativeDisplay *> physical_displays;
-  std::vector<std::vector<NativeDisplay *>> mdisplays;
+  std::vector<uint32_t> logical_displays;
+  std::vector<std::vector<uint32_t>> mosiac_displays;
+  std::ifstream fin(hwc_dp_cfg_path);
+  std::string cfg_line;
+  std::string key_logical("LOGICAL");
+  std::string key_mosiac("MOSIAC");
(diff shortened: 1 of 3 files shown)


yugang:
@kalyankondapally , this PR is a temp version for linux for review, so i didn't delete comments and print log. i need time to check it on Android IA too, and pls help review firstly to check if the features meet your requirements. thanks

common/core/gpudevice.cpp
```diff
@@ -41,126 +41,171 @@ bool GpuDevice::Initialize() {
...
-            mosiac_index_added = i;
-          }
-
-          break;
-        }
-      }
+  // Save save no mosiac displays to final total_displays_
+  for (size_t i = 0; i < available_displays.size(); i++) {
+    if (available_displays.at(i)) {
+      total_displays_.emplace_back(temp_displays.at(i));
```
kalyankondapally: these should be sorted by order. i.e. Mosiac Display having display 1 should be first in list and so on. It could be that Logical display could have 1 and be part of Mosiac display etc.

kalyankondapally: or could be that physical display 1 is just an independent display

hwc_display.ini
```diff
@@ -0,0 +1,34 @@
+# The switches of logical and mosiac mode.
+
+LOGICAL="true"
```
kalyankondapally: by default these should be false

yugang:
Updated, keep the order, e.g. if a no mosac-ed logical display has a smaller display index comparing with mosac display which includes multiple logical displys( and the index of them are bigger than the one "no mosac-ed "), the "no mosac-ed logical display" will be the front of the mosac("no mosac-ed logical display" has smaller index in the total_displays_).

yugang:
when two physical displays are connected:

LOGICAL_DISPLAY="0:3"
LOGICAL_DISPLAY="1:2"

How about the differents between:
MOSIAC_DISPLAY="2+5"
MOSIAC_DISPLAY="0+3+4"

and

MOSIAC_DISPLAY="2+5"
MOSIAC_DISPLAY="3+0+4"

in current implementation, i found there are wrong rendering of "3+0+4", not sure it is a test app limitation or logical/mosiac issue, i checked the config reading functions, should be ok there.

kalyankondapally:
@yugang We should rely on the displays being connected in the order they are mentioned in the settings. So, it should be fine

yugang:
Add disconnected/connected checking if user config use the disconnected display and skip the disconnected displays( and their logical displays).

yugang:
Verified in android ia, the config reading is ok:
1. adb shell mount -o rw,remount /system
2.  set logical and mosiac as true in the hwc_display.ini, and "adb push hwc_display.ini /etc/"
3. stop and start the UI again

but still met some disordered mosiac screens if restart the ui  several times, seems there are some error in logical/mosiac mode, or some disordered usages from android ia high level.

yugang:
remove the disconnected/connected checking as the logical can be created successfully even though the physical display is disconnected, the old issues is due to wrong index, fixed.

PR #144, opened 2017-10-19
Fix build for Linux

XiaosongWei:
Fix va support for Linux


XiaosongWei:
@kalyankondapally

PR #145, opened 2017-10-19
Add Color transform Matrix

js0701:
This is based on gaochun's PR
https://github.com/intel/IA-Hardware-Composer/pull/130
A few changes:
1) dismangle with SetColorCorrection
2) remove macro DRM_CTM, but use a function to check display
support CTM or not
3) use a separate flag if GPU CTM is forced
4) FloatToFixPoint should mutiple (1<<31)

Tests: Settings-Accessibility-Color Correction test pass on GP
Jira: OAM-49721

### DIFF
--- a/common/display/displayqueue.cpp
+++ b/common/display/displayqueue.cpp
@@ -51,6 +51,8 @@ DisplayQueue::DisplayQueue(uint32_t gpu_fd, bool disable_overlay,
 
   /* use 0x80 as default brightness for all colors */
   brightness_ = 0x808080;
+  /* use HWCColorTransform::kIdentical as default color transform hint */
+  color_transform_hint_ = HWCColorTransform::kIdentical;
   /* use 0x80 as default brightness for all colors */
   contrast_ = 0x808080;
   /* use 1 as default gamma value */
@@ -399,6 +401,10 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
   bool composition_passed = true;
   bool disable_ovelays = state_ & kDisableOverlayUsage;
 
+  //FIXME? disalbe overlay when need apply CTM but is not supported in
+  // display so need to fallback to GPU
+  disable_ovelays = disable_ovelays | ( state_ & kApplyGPUColorMatrix);
+
   // Validate Overlays and Layers usage.
   if (!validate_layers) {
     bool can_ignore_commit = false;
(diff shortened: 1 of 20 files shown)


common/display/displayqueue.cpp
```diff
@@ -399,6 +401,10 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
   bool composition_passed = true;
   bool disable_ovelays = state_ & kDisableOverlayUsage;
 
+  //FIXME? disalbe overlay when need apply CTM but is not supported in
+  // display so need to fallback to GPU
+  disable_ovelays = disable_ovelays | ( state_ & kApplyGPUColorMatrix);
```
kalyankondapally: Can we avoid this ?

wsi/drm/drmbuffer.cpp
```diff
@@ -64,7 +64,6 @@ void DrmBuffer::Initialize(const HwcBuffer& bo) {
   switch (format_) {
     case DRM_FORMAT_NV12:
     case DRM_FORMAT_NV16:
-    case DRM_FORMAT_P010:
       total_planes_ = 2;
```
kalyankondapally: u want to avoid this change

os/android/iahwc2.cpp
```diff
@@ -681,7 +682,20 @@ HWC2::Error IAHWC2::HwcDisplay::SetColorTransform(const float *matrix,
...
+  if (hint != HAL_COLOR_TRANSFORM_IDENTITY &&
+      hint != HAL_COLOR_TRANSFORM_ARBITRARY_MATRIX &&
+      hint != HAL_COLOR_TRANSFORM_VALUE_INVERSE &&
+      hint != HAL_COLOR_TRANSFORM_GRAYSCALE &&
+      hint != HAL_COLOR_TRANSFORM_CORRECT_PROTANOPIA &&
+      hint != HAL_COLOR_TRANSFORM_CORRECT_DEUTERANOPIA &&
+      hint != HAL_COLOR_TRANSFORM_CORRECT_TRITANOPIA)
+    return HWC2::Error::BadParameter;
+
+  ETRACE("Calling SetColorTransform!\n");
```
kalyankondapally: you can remove this.

js0701: right

wsi/drm/drmbuffer.cpp
js0701: OK, yes

common/display/displayqueue.cpp
js0701: @kalyankondapally  But how can we make use it goes to the GPU composition when CTM is not supported directly in display

kalyankondapally: Once Validate layers is done, we can set force gpu rendering for each display plane state:
https://github.com/intel/IA-Hardware-Composer/blob/master/common/display/displayplanestate.h#L157

Once that is done, we need to ensure each plane has offscreen target i.e. 
Check if  NativeSurface* surface = last_plane.GetOffScreenTarget(); is available, if this is null ensure offscreen target is set by callng:
https://github.com/intel/IA-Hardware-Composer/blob/master/common/display/displayqueue.cpp#L241

PR #149, opened 2017-10-27
Suppress compile warnings on Android

renchenglei:
Jira: None.
Test: Build passes on Android and no warnings reported.


renchenglei:
@kalyankondapally @js0701 
I'm clean compile warnings on Android. Many unused parameter is caused by unfinished function in os/android/hwcservice.cpp. Currently, we suppress warning reporting in Android.mk file. PTAL.

js0701:
@kalyankondapally  OneAndroid has complained about those warnings. This PR is to remove most Android warnings

PR #150, opened 2017-10-27
Android: Enable color correction and inversion

gaochun:
Add setColorTransform to HWC2. Get 4x4 color transform matrix (CTM)
from SurfaceFlinger, and convert the matrix to 3x3 DRM CSC matrix
and DRM CSC post offset.

Jira: OAM-49721,OAM-50246
Test: Go to Settings-Accessibility-Color correction, set it to On
and choose the Correction mode; Go to Settings-Accessibility-Color
inversion, set it to On.

Signed-off-by: Gao Chun <chun.gao@intel.com>

### DIFF
--- a/common/display/displayqueue.cpp
+++ b/common/display/displayqueue.cpp
@@ -51,6 +51,8 @@ DisplayQueue::DisplayQueue(uint32_t gpu_fd, bool disable_overlay,
 
   /* use 0x80 as default brightness for all colors */
   brightness_ = 0x808080;
+  /* use HWCColorTransform::kIdentical as default color transform hint */
+  color_transform_hint_ = HWCColorTransform::kIdentical;
   /* use 0x80 as default brightness for all colors */
   contrast_ = 0x808080;
   /* use 1 as default gamma value */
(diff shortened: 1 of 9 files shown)


gaochun:
The update patch uses a new drm interface to set the ctm post offset. So it's depends on the kernel patch and the libdrm patch intel/external-libdrm#6

gaochun:
@kalyankondapally @js0701 Please review. For the GL fallback, as it will bring more complexity,  I think we can add it after the feature is enabled on own target device.

js0701:
Can see leave SetColorCorrection unchanged?
Just add a new function SetColorTransfrom in drmdisplay.cpp
The benefit is it will not break Linux test problem and it is more clear

kalyankondapally:
hmm ya makes sense to split them up.

gaochun:
I'll make SetColorCorrection unchanged, thanks.

gaochun:
Patch updated. @js0701 Add a API SetColorTransformMatrix to PhysicalDisplay, you can add your linux CTM test for it in future.

PR #151, opened 2017-10-30
Fix the issue that the HWCLock thread will not exit.

sophy228:
The displaymanager thread is blocked in "read" call,
so it will never get chance to disable HWClock.

As we already use "select" to receive the "POLL IN" event of
the fd, we had better to set the "Non Blocking" flag on.

Change-Id: I01bfd52e1c822421b76a1ad7e8b06f1b87ee5371
Signed-off-by: Zhongmin Wu <zhongmin.wu@intel.com>


kalyankondapally:
merged

PR #152, opened 2017-10-30
removing not used file Android.common.mk

munish-b:
Jira:None
Tests:None
Signed-off-by:Munish Bhardwaj<munishx.bhardwaj@intel.com>


kalyankondapally:
We dont need this as we now use it.

PR #155, opened 2017-11-02
Fix for crash during rotation

pallavigopal:
The cursor z_order is not the highest in one of the corner case,
so we end up with segmentation fault in cursor layer getbuffer becoz the
hwc implementation is based on the assumption that the cursor is always on the top.

Jira: None
Test: Able to rotate the screen without system reboot


pallavigopal:
@kalyankondapally  please review

kalyankondapally:
thx merged

PR #156, opened 2017-11-03
Enable VA support by default

XiaosongWei:
It's verified on 1Android tree with VPG's driver.

Change-Id: I486fa16843b6958ecbd2da6706d76d2be414a26c
Tracked-On:


kalyankondapally:
Removed the boardconfig and just removed the compiler checks making this a hard dependency

PR #162, opened 2017-11-10
Fix CPU busy caused by HWCLock thread

js0701:
It will not exit due to release_lock false
This fix is to ensure HWCLock will exit

Jira: https://jira01.devtools.intel.com/browse/OAM-51903
Test: After install hwcompoper to DUT
use top to see the process CPU usage

### DIFF
--- a/wsi/drm/drmdisplaymanager.cpp
+++ b/wsi/drm/drmdisplaymanager.cpp
@@ -396,14 +396,17 @@ void DrmDisplayManager::ForceRefresh() {
   spin_lock_.unlock();
 }
 
-void DrmDisplayManager::HandleLazyInitialization() {
+bool DrmDisplayManager::HandleLazyInitialization() {
+  bool dispatched = false;
   spin_lock_.lock();
   if (release_lock_ && hwc_lock_.get()) {
     hwc_lock_->DisableWatch();
     hwc_lock_.reset(nullptr);
     release_lock_ = false;
+    dispatched = true;
   }
   spin_lock_.unlock();
+  return dispatched;
 }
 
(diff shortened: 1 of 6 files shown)


wsi/drm/drmdisplaymanager.cpp
```diff
@@ -396,14 +396,17 @@ void DrmDisplayManager::ForceRefresh() {
   spin_lock_.unlock();
 }
 
-void DrmDisplayManager::HandleLazyInitialization() {
+bool DrmDisplayManager::HandleLazyInitialization() {
+  bool dispatched = false;
   spin_lock_.lock();
   if (release_lock_ && hwc_lock_.get()) {
```
kalyankondapally: Do you know why release_lock_ is not true here. We shouldn't even be here if it's not true.

js0701: release_lock_ is initialized to be false... So HWCLock is always on

kalyankondapally: Sequence is this:
1) DisplayQueue should not update anything till lock is held.
2) Once lock is released and we get notification from HWClock thread we call Refresh.
3)In Refresh, ignore update flag is removed.
4)next update DisplayQueue calls HandleLazyInitialization

kalyankondapally: https://github.com/intel/IA-Hardware-Composer/commit/bbba01b3d30dc51272143ee826457cbb0108960b can u check if this resolves the issue being seen. Atleast this makes it explicit.

kalyankondapally:
As discussed offline closing this.

PR #163, opened 2017-11-14
fix the screencap black screen issue

YangShi-Intel:
When there is no display to connect device, the original HWC module
will create only 1*1 framebuffer.
So screencap will not works properly, it only can get a black screen.
This will block many test case.

In order to fix this issue, we make the default width and height to
1080p format for most HDMI device use this format.

JIRA: none
TEST: screencap can works well when don't connect any display device
Signed-off-by: Yang Shi <yang.a.shi@intel.com>


kalyankondapally:
thx merged

PR #164, opened 2017-11-14
Fix for Flicker when unlock by pin/pattern

shwethabolumbu:
Observed screen corruption and flicker when we unlock and go to homescreen.

When we try to do surafce damage optimization for the given layer with alpha
equal to 255(status bar in this case),there is a corruption & flicker.

Made clear_surface true if the layer alpha equal to 255 then no issues.

Jira: None
Test: Tested unlocking directly to the home screen using pin (and pattern).
       No flicker & corruption observed.
Signed-off-by: bshwetha <shwetha.b@intel.com>

### DIFF
--- a/common/display/displayqueue.cpp
+++ b/common/display/displayqueue.cpp
@@ -164,7 +164,6 @@ void DisplayQueue::GetCachedLayers(const std::vector<OverlayLayer>& layers,
 
     if (plane_state_render || plane.SurfaceRecycled()) {
       bool clear_surface = false;
-      bool alpha_damaged = false;
       bool content_changed = false;
       const std::vector<size_t>& source_layers = last_plane.source_layers();
       HwcRect<int> surface_damage = HwcRect<int>(0, 0, 0, 0);
@@ -197,11 +196,7 @@ void DisplayQueue::GetCachedLayers(const std::vector<OverlayLayer>& layers,
           // one layer having alpha = 255 and damaged.
           if (layer.GetBlending() == HWCBlending::kBlendingPremult &&
               layer.GetAlpha() == 255) {
-            if (alpha_damaged) {
               clear_surface = true;
-            } else {
-              alpha_damaged = true;
-            }
           }
         }
       }


shwethabolumbu:
@kalyankondapally @pallavigopal  please review

common/display/displayqueue.cpp
```diff
@@ -197,11 +196,7 @@ void DisplayQueue::GetCachedLayers(const std::vector<OverlayLayer>& layers,
           // one layer having alpha = 255 and damaged.
           if (layer.GetBlending() == HWCBlending::kBlendingPremult &&
               layer.GetAlpha() == 255) {
-            if (alpha_damaged) {
```
kalyankondapally: We dont actually use surface damage. This has been disabled as we found some regressions. Now, it seems for some reason we don't clear surface/swap surfaces in TransferSurface below. Can u check in this case if display frame rect has changed or not ?

shwethabolumbu: Rectangle/position are not changed but content is changed, so we need to clear the surface

shwethabolumbu: please let me know where it disabled in HWC i dont find that or it disabled in mesa/frameworks ?

kalyankondapally: You can see here https://github.com/intel/IA-Hardware-Composer/blob/master/common/core/overlaylayer.cpp#L220 we always use display_frame_rect as surface damage. 

So, content is changed but we didn't re-draw it ?Can u check if content changed is set to true in this case here ?

shwethabolumbu: yes. content is changed. But we are not redrawing it.

kalyankondapally: k, can you debug why https://github.com/intel/IA-Hardware-Composer/blob/master/common/core/overlaylayer.cpp#L357 this flag is getting removed or why content_changed = true is not being set here. That probably will give us the right fix.

shwethabolumbu: I meant content_changed flag is true and also content is changed.
I checked again redrawing also happening.
But clear_surface flag is false so end up corruption and flicker.
if i make that clear_surface = true when the content_changed and alpha =255 then no flicker & corruption.

I hope making clear_surface=true is the right fix from my understanding maybe am wrong please correct me

shwethabolumbu: otherway to fix it is make the clearsurface true in the state_ variable if the layer content changed.
ie . In the line https://github.com/intel/IA-Hardware-Composer/blob/master/common/core/overlaylayer.cpp#L222 
if i add one more check   if (layer->HasLayerContentChanged() || with other checks also the clear_surface is true and i dont see the issue

kalyankondapally: Can you share logs for displayframe, source rect of the frame which has flicker and the before one, along with surface damage values from https://github.com/intel/IA-Hardware-Composer/blob/master/common/core/hwclayer.cpp#L112 and if visible region has changed https://github.com/intel/IA-Hardware-Composer/blob/master/common/core/hwclayer.cpp#L164 I want to check what state is really changing between both frames and see if we can figure out the case which is missing in our checks today.

shwethabolumbu: When we move from unlock pattern screen to home screen, 
the status bar buffer size is changed from 2560x1440 to 2560x24. Accordingly visible region is also changed but the surface damage remains 2560x1440 for first two frames and after that it is 2560x24. In that case e need to clear the surfaces

Log attached -
[flicker_log.txt](https://github.com/intel/IA-Hardware-Composer/files/1478229/flicker_log.txt)

kalyankondapally: Thx this is nice finding. const std::unique_ptr<OverlayBuffer>& buffer = imported_buffer_->buffer_;

In case we have previous layer check true i.e previous_layer https://github.com/intel/IA-Hardware-Composer/blob/master/common/core/overlaylayer.cpp#L215, you can have boolean to check if buffer size has changed between frames i.e. Current layer buffer is const std::unique_ptr<OverlayBuffer>& buffer = imported_buffer_->buffer_;  previous layer buffer is 
const std::unique_ptr<OverlayBuffer>& buffer = previous_layer->imported_buffer_->buffer_; You should be able to query buffer width and height from  the buffer. If they dont match the boolean can be true. You can add it checks here: https://github.com/intel/IA-Hardware-Composer/blob/master/common/core/overlaylayer.cpp#L223

kalyankondapally: Ignore the buffer change one.

Couple of things to check:
1) Can you check if layer->HasVisibleRegionChanged() is returning true in this case, which should set the flag that we need to clear the surface. If this is not being set correctly, than the bug might be here https://github.com/intel/IA-Hardware-Composer/blob/master/common/core/hwclayer.cpp#L143  
2) When regions have changed, we set full surface damage to the offscreen surface https://github.com/intel/IA-Hardware-Composer/blob/master/common/display/displayqueue.cpp#L219. and we clear the current offscreen surface. I think we need to do the here:
After line 213 
i.e. plane.TransferSurfaces(last_plane, content_changed);
        std::vector<NativeSurface*>& surfaces = last_plane.GetSurfaces();
        size_t size = surfaces.size();
        bool force_reset = false;
      if (region_changed) {
        for (size_t i = 0; i < size; i++) {
          surfaces.at(i)->EnsureSufaceCleared() -> or whatever API u can think off.
      }

     if (size > 0){
       force_reset = surfaces.at(0)->ClearSurface(); -> Getter to check the state in case surface needs to be cleared.
      }
    
   if (force_reset){
    clear_surface = true;
  }
  https://github.com/intel/IA-Hardware-Composer/blob/master/common/display/displayqueue.cpp#L214 this would change to: 
      if (region_changed || clear_surface) {
        const HwcRect<int>& current_rect = last_plane.GetDisplayFrame();
        for (size_t i = 0; i < size; i++) {
          surfaces.at(i)->UpdateSurfaceDamage(current_rect, current_rect);
          surfaces.at(i)->UpdateDisplayFrame(current_rect);
        }
      } else {
        const std::vector<CompositionRegion>& comp_regions =
            plane.GetCompositionRegion();
        last_plane.GetCompositionRegion().assign(comp_regions.begin(),
                                                 comp_regions.end());
}

And you would reset the flag to clear Surface here:
https://github.com/intel/IA-Hardware-Composer/blob/master/common/compositor/gl/glrenderer.cpp#L87

3) While writing 2 above, realized we have a bug here:
https://github.com/intel/IA-Hardware-Composer/blob/master/common/display/displayqueue.cpp#L214 this should have been       if (region_changed || clear_surface) {

Try 2 though, I think we should be doing that and see if it fixes u flicker issues.

pallavigopal: The layer->HasVisibleRegionChanged()  is true when the region changed ie from 2560x1440 to 2560x24. so in the GetcachedLayer also region_changed and clear_surface is also true. No issue in that case glclear will happen. 
Issue is with the layer buffer size change & surface damage change which is kind of random with the same scenario. so far we observed two different behaviors when status bar source crop change from 2560x1440 to 2560x24
1. surface damage is 2560x1440 for 3 frames then changed to 2560x24
2. surface damage is 2560x1440 for 1 frame then changed to 2560x24
But the visible region and display frame is 2560x24. Also the VisibleRegionChanged is true only for the first change after that no change in the visible region.

pallavigopal: The above suggestions are not fixing the issue still flicker is observed.
if we clear_surface only things works

kalyankondapally: can u share the patch with proposal 2 above ?

pallavigopal: Shwetha will share the patch I discussed with her she had tried the changes you suggested expect changing it from plane level to surface level everywhere. she will add that also and share the change.
I have some doubts we do the above surface changes only when region_change || clear_surface but issue is seen in the scenario only when conten_changed rest are false. Then how this will fix that issue. Maybe am not getting the point

kalyankondapally: Let's say the plane has three surfaces now and this frame caused things which force us to clear the surface:
1) The current offscreen target is cleared and re-drawrn. However, we don't clear the second and triple buffer. 
2)When the surface is swapped in next frame second buffer becomes current and it's damage is sum of current damage + previous frame. If previous frame damage caused us to clear the surface we didn't account for it here. This just makes sure in this case we clear all surfaces eventually.

pallavigopal: thanks for the clarification. I hope by changing from plane to surface things will work

shwethabolumbu: Applied proposal 2 when content is changed as well. i.e 
within this condition https://github.com/intel/IA-Hardware-Composer/blob/3e253028e74d9c09f4be3b631de94f5637083bdd/common/display/displayqueue.cpp#L230 
force_reset = surfaces.at(0)->ClearSurface();
if(force_reset)
{
clear_surface = true;
}
This works fine

kalyankondapally: hmm no if we do this when content is changed always there is no point in doing this. I understand this is fixing the issue for u, but it doesn't add much value.

Sorry, u need to try one more. We need to understand precisely what is causing this clear
What about 2+ "In case we have previous layer check true i.e previous_layer https://github.com/intel/IA-Hardware-Composer/blob/master/common/core/overlaylayer.cpp#L215, you can have boolean to check if buffer size has changed between frames i.e. Current layer buffer is const std::unique_ptr& buffer = imported_buffer_->buffer_; previous layer buffer is
const std::unique_ptr& buffer = previous_layer->imported_buffer_->buffer_; You should be able to query buffer width and height from the buffer. If they dont match the boolean can be true. You can add it checks here: https://github.com/intel/IA-Hardware-Composer/blob/master/common/core/overlaylayer.cpp#L223"

kalyankondapally: Also, can u upload your patch so I can see if we are missing some checks somewhere

kalyankondapally: This is what I think we should do https://github.com/intel/IA-Hardware-Composer/commit/73c4df4a49d2d3e53d1f627b31f3a6c5cbf97c6e . Can you test with this, if not lets see if we need to take into account buffer size changes also.

kalyankondapally: @shwethabolumbu  Any updates here ?

shwethabolumbu:
Sorry for the delay
Tried with the branch which you sent recently. Flicker is seen in that also.

shwethabolumbu:
closed by mistake

kalyankondapally:
try with master as all changes are in there. We didnt need original WA for some of the issues we had. So, now this is something new. What we want to check is to check display frame, visible region,  buffer size couple of frames before and after flicker. Also check if we are clearing surface in any of these frames.

shwethabolumbu:
Checked with master branch. 
Patch on ensure clearing surfaces fixes this.

PR #173, opened 2017-11-23
Revert patches

Kishore409:
Jira: None
Test: Cold/Warm reboot with displays connected, no blank screen
              in either of displays


kalyankondapally:
@foolmm ping FYI this broke multi display support so need to revert this. You can work with @Kishore409  to know about the issue.

kalyankondapally:
merged thanks

PR #176, opened 2017-11-28
Add external layer buffer cache management

js0701:
to avoid import and generate eglImage and texture
each time layer is input

Change-Id: I39d7dd2e99b647d48723227582e250319fcc4916
Jira: None
Tests: UI, Video, Camera

### DIFF
--- a/common/compositor/gl/glrenderer.cpp
+++ b/common/compositor/gl/glrenderer.cpp
@@ -21,16 +21,25 @@
 #include "nativesurface.h"
 #include "renderstate.h"
 #include "shim.h"
+#include "hwclayerbuffermanager.h"
 
 namespace hwcomposer {
 
 GLRenderer::~GLRenderer() {
+  if (!context_.MakeCurrent()) {
+    ETRACE("Failed make current context.");
+    return;
+  }
+
+  if (buffer_manager_)
+    buffer_manager_->PurgeGraphicsResources();
   if (vertex_array_)
     glDeleteVertexArraysOES(1, &vertex_array_);
 }
 
-bool GLRenderer::Init() {
+bool GLRenderer::Init(HwcLayerBufferManager *buffer_manager) {
   // clang-format off
+  buffer_manager_ = buffer_manager;
   const GLfloat verts[] = {0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 2.0f,
                            0.0f, 2.0f, 2.0f, 0.0f, 2.0f, 0.0f};
   // clang-format on
--- a/common/core/hwclayerbuffermanager.cpp
+++ b/common/core/hwclayerbuffermanager.cpp
@@ -1,0 +1,88 @@
+/*
+// Copyright (c) 2016 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+*/
+
+#include "hwclayerbuffermanager.h"
+
+namespace hwcomposer {
+
+HwcLayerBufferManager::HwcLayerBufferManager() {
+  for(size_t i = 0; i < BUFFER_CACHE_LENGTH; i++)
+    cached_buffers_.emplace_back();
+}
+
+HwcLayerBufferManager::~HwcLayerBufferManager() {
+    PurgeBuffer();
+}
+
+void HwcLayerBufferManager::PurgeBuffer() {
+  for(auto &map : cached_buffers_) {
+  	map.clear();
+  }
+}
+
+void HwcLayerBufferManager::PurgeGraphicsResources() {
+  for (auto& map : cached_buffers_) {
+    for (auto& pair : map) {
+      pair.second->DeleteImage();
+      pair.second->DeleteTexture();
+    }
+  }
+}
+
+void HwcLayerBufferManager::Dump() {
+}
+
+std::shared_ptr<OverlayBuffer>& HwcLayerBufferManager::FindCachedBuffer(const HWCNativeBuffer& native_buffer) {
+
+  BUFFER_MAP&  first_map = cached_buffers_[0];
+  static std::shared_ptr<OverlayBuffer> pBufNull = nullptr;
+  for (auto& map : cached_buffers_) {
+    if (map.count(native_buffer)) {
+      std::shared_ptr<OverlayBuffer>& pBuf = map[native_buffer];
+      if (&map != &first_map) {
+        first_map.emplace(std::make_pair(native_buffer, pBuf));
+        map.erase(native_buffer);
+      }
+#ifdef CACHE_TRACING
+      hit_count_++;
+#endif
+      return pBuf;
+    }
+  }
+
+#ifdef CACHE_TRACING
+  miss_count_++;
+  if (miss_count_ % 100 == 0)
+    ICACHETRACE("cache miss count is %llu, while hit count is %llu",
+                miss_count_, hit_count_);
+#endif
+
+  return pBufNull;
+}
+
+void HwcLayerBufferManager::RegisterBuffer(const HWCNativeBuffer& native_buffer, std::shared_ptr<OverlayBuffer>& pBuffer) {
+
+  BUFFER_MAP&  first_map = cached_buffers_[0];
+  first_map.emplace(std::make_pair(native_buffer, pBuffer));
+}
+
+void HwcLayerBufferManager::RefreshBufferCache() {
+  auto begin = cached_buffers_.begin();
+  cached_buffers_.emplace(begin);
+  cached_buffers_.pop_back();
+}
+
+}  // namespace hwcomposer
--- a/common/core/hwclayerbuffermanager.h
+++ b/common/core/hwclayerbuffermanager.h
@@ -1,0 +1,77 @@
+/*
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+*/
+
+/*
+Design of hwclayerbuffermanager:
+The purpose is to add cache magagement to external buffer owned by hwcLayer
+to avoid import buffer and glimage/texture generation overhead
+
+1: the hwclayerbuffermanager is owned per display, as each display has a
+separate
+GL context
+2: hwclayerbuffermanager stores a refernce of external buffers in a vector
+   cached_buffers, each vector member is hash map.
+   The vector stores history buffer in this way, vector[0] is for the current
+   frame buffers, vector[1] is for last frame (-1) buffers, vector[2] is -2
+   frames buffers etc. A constant (currently 4) frames of buffers is stored.
+   When a buffer refernce is stored in vector[3] and it is not used in the
+   current frame present, it will go out of scope and be released.
+   If a buffer is fetched from map, it will always re-registered in vector[0]
+map.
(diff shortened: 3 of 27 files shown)


common/core/hwclayerbuffermanager.cpp
```diff
@@ -0,0 +1,74 @@
...
+// limitations under the License.
+*/
+
+#include "hwclayerbuffermanager.h"
+#include <hwclayer.h>
+
+#include <cmath>
+
+namespace hwcomposer {
+
```
kalyankondapally: Can you add documentation as how this manager would work i.e. how it would release the cached buffers etc

common/display/virtualdisplay.cpp
```diff
@@ -39,6 +39,10 @@ VirtualDisplay::VirtualDisplay(uint32_t /*gpu_fd*/,
       buffer_handler_(buffer_handler),
       width_(0),
       height_(0) {
+  buffer_manganer_.reset(new HwcLayerBufferManager());
+  if (!buffer_manganer_) {
```
kalyankondapally: buffer_manager_

os/android/platformdefines.h
```diff
@@ -51,6 +49,33 @@ struct gralloc_handle {
...
 typedef struct gralloc_handle* HWCNativeHandle;
+typedef native_handle_t HWCNativeBuffer;
+#define GETNATIVEBUFFER(handle) (*(handle->handle_))
+
+struct BufferHash {
+  size_t operator()(native_handle_t const& p) const {
+    std::size_t seed = 0;
+    hash_combine_hwc(seed, p.numFds);
+    for (int i = 0; i < p.numFds; i++) {
+      hash_combine_hwc(seed, (const size_t)p.data[i]);
```
kalyankondapally: shouldn't we have width, height, format etc like linux ?

kalyankondapally:
@yugang @harishkrupo @Kishore409 @XiaosongWei  Please take a look and review the proposal.

common/core/hwclayerbuffermanager.cpp
```diff
@@ -0,0 +1,74 @@
...
+void HwcLayerBufferManager::RegisterBuffer(const HWCNativeBuffer& native_buffer, std::shared_ptr<OverlayBuffer>& pBuffer) {
+
+  BUFFER_MAP&  first_map = cached_buffers_[0];
+  first_map.emplace(std::make_pair(native_buffer, pBuffer));
+}
+
+void HwcLayerBufferManager::RefreshBufferCache() {
+     auto begin = cached_buffers_.begin();
+	 cached_buffers_.emplace(begin);
```
kalyankondapally: fix style format

kalyankondapally:
Also any stats on how many times we have the cache hits on Android ?

common/core/hwclayerbuffermanager.cpp
js0701: OK. I think the whole design is not complicate. So I list a few key points here:
1: the hwclayerbuffermanager is owned per display, as each display has a separate GL context 
2: hwclayerbuffermanager stores refernces of external buffers in a vector, each vector member is hash map. The vector stores history buffer in this way, vector[0] is for the current frame buffers, vector[1] is for last frame (-1) buffers, vector[2] is -2 frames buffers etc. A constant (currently 4) frames of buffers is stored. When a buffer refernce  is stored in vector[3] and it is not used in the current present, it will go out of scope and be released. If a buffer is fetched from map, it will always re-registered in vector[0] map.
3; egl-image and texture can be owned by drmbuffer and reused

os/android/platformdefines.h
js0701: on Android, the only input is buffer_handle_t
See
HWC2::Error IAHWC2::Hwc2Layer::SetLayerBuffer(buffer_handle_t buffer,
                                              int32_t acquire_fence) 

numFds and data are the only members of buffer_handle_t

js0701:
I have printed some logs which shows the cache hit.
But I have no stat about miss/hit rate.
You are right, I may need to give some data

common/core/hwclayerbuffermanager.cpp
kalyankondapally: Adding it as comments in Code would help for future.

js0701:
the cache hit rate is around 95% and have added log machanism for it

common/core/hwclayerbuffermanager.cpp
kalyankondapally: can u run git diff | clang-format-diff-3.5 -p 1 -style=file -i

common/core/hwclayerbuffermanager.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+void HwcLayerBufferManager::RegisterBuffer(const HWCNativeBuffer& native_buffer, std::shared_ptr<OverlayBuffer>& pBuffer) {
+
+  BUFFER_MAP&  first_map = cached_buffers_[0];
+  first_map.emplace(std::make_pair(native_buffer, pBuffer));
+}
+
+void HwcLayerBufferManager::RefreshBufferCache() {
+     auto begin = cached_buffers_.begin();
```
kalyankondapally: What are we actually doing here ?

common/core/hwclayerbuffermanager.h
```diff
@@ -0,0 +1,74 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+*/
+
+/*
+Design of hwclayerbuffermanager:
+The purpose is to add cache magagement to external buffer owned buy hwcLayer
```
kalyankondapally: buy -> by

common/core/hwclayerbuffermanager.h
```diff
@@ -0,0 +1,74 @@
...
+to avoid import buffer and glimage/texture generation overhead
+
+1: the hwclayerbuffermanager is owned per display, as each display has a separate
+GL context
+2: hwclayerbuffermanager stores refernces of external buffers in a vector
```
kalyankondapally: a reference

common/core/hwclayerbuffermanager.h
js0701: this is to make the oldest reference of buffers release
and make the other buffers grow age

common/core/hwclayerbuffermanager.cpp
js0701: this is to make the oldest reference of buffers release
and make the other buffers grow age

kalyankondapally:
@strassek  Sorry forgot to ping u earlier, If you want to give your feedback.

common/core/hwclayerbuffermanager.cpp
```diff
@@ -0,0 +1,80 @@
...
+  for(size_t i = 0; i < BUFFER_CACHE_LENGTH; i++)
+    cached_buffers_.emplace_back();
+}
+
+HwcLayerBufferManager::~HwcLayerBufferManager() {
+    PurgeBuffer();
+}
+
+void HwcLayerBufferManager::PurgeBuffer() {
+  for(auto &map : cached_buffers_) {
```
kalyankondapally: This should be called in CompositorThread only. I think CompositorThread owing the ownership of texture, eglimage of buffer makes sense. We need to make sure that context is current in this thread before releasing these resources.

common/display/displayqueue.cpp
```diff
@@ -300,6 +306,8 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
     cursor_state_changed = true;
   }
 
+  buffer_manager_->RefreshBufferCache();
```
kalyankondapally: This is needed every time ?

common/core/hwclayerbuffermanager.h
```diff
@@ -0,0 +1,76 @@
...
+struct HwcLayer;
+class OverlayBuffer;
+class NativeBufferHandler;
+
+class HwcLayerBufferManager {
+
+public:
+  HwcLayerBufferManager();
+  ~HwcLayerBufferManager();
+  void PurgeBuffer();
```
kalyankondapally: No body outside this class is calling this, can we make it private ?

js0701: OK, right

common/display/displayqueue.cpp
js0701: yes, it should be called each frame

common/core/hwclayerbuffermanager.cpp
js0701: Will make it private

kalyankondapally: This is where we release gl resources https://github.com/intel/IA-Hardware-Composer/blob/master/common/compositor/compositorthread.cpp#L116  we need to make sure all this is sync. Only way is to move destruction of buffers to compositor thread.

js0701: You mean we should call purge in here?https://github.com/intel/IA-Hardware-Composer/blob/master/common/compositor/compositorthread.cpp#L116 ?
It makes sense

kalyankondapally: ya

kalyankondapally:
@js0701 You might need to rebase this.

common/core/hwclayerbuffermanager.cpp
js0701: We can't do it as ReleaseGpuResources is called each Draw frame, and it will make cache for GL resources not meaningful?

js0701:
@kalyankondapally  the new PR is to make sure remove GPU resources before EGL context is freed

common/compositor/gl/glrenderer.cpp
```diff
@@ -21,16 +21,25 @@
...
 namespace hwcomposer {
 
 GLRenderer::~GLRenderer() {
+  if (!context_.MakeCurrent()) {
+    ETRACE("Failed make current context.");
+    return;
+  }
+
+  if (buffer_manager_)
+    buffer_manager_->PurgeGraphicsResources();
```
kalyankondapally: Hmm we do this only when renderer is destroyed, we need to do this every time a buffer is destroyed

common/core/hwclayerbuffermanager.cpp
kalyankondapally: I think what we want to do is after every frame post  a task to clean up any buffers which can be destroyed. So, this PurgeBuffer should be called by posting a task to CompositorThread from DisplayQueue after you have updated commit for the current frame. This task should only posted when previous frame had involved gpu composition otherwise you could call PergeBuffer from DisplayQueue itself

common/core/hwclayerbuffermanager.cpp
```diff
@@ -0,0 +1,88 @@
...
+void HwcLayerBufferManager::RegisterBuffer(const HWCNativeBuffer& native_buffer, std::shared_ptr<OverlayBuffer>& pBuffer) {
+
+  BUFFER_MAP&  first_map = cached_buffers_[0];
+  first_map.emplace(std::make_pair(native_buffer, pBuffer));
+}
+
+void HwcLayerBufferManager::RefreshBufferCache() {
+  auto begin = cached_buffers_.begin();
+  cached_buffers_.emplace(begin);
+  cached_buffers_.pop_back();
```
kalyankondapally: This line:  cached_buffers_.pop_back();  would destroy buffers right ? In this case you want to move it to another list and delete these buffers from Compositor thread.

kalyankondapally: Thinking more on this case, simple way would be store these buffers which go out of scope in a separate vector, in DisplayQueue after u do the current commit, query for these buffers from BufferManager. If vector is not empty, pass the ownership of these buffers to DisplayQueue. DisplayQueue would than post a task to CompositorThread. Compositor thread in its side would delete all the buffers which dont need gl context first (bu checking if buffer has valid gl resources). If it has buffers which have gl resources, than it will ensure context is initialized and delete these buffers.

kalyankondapally: Actually I am bit confused what this function is also doing. Can you explain ?

kalyankondapally: @js0701 On top of this work I refactored the internals so all resources are only released in compositorthread:
https://github.com/intel/IA-Hardware-Composer/commits/next

Can you test this on u side and see if surfacedamage still works like before ? If things look good I can send it as PR for review.

kalyankondapally: Also, I see lot of these fb creation failures during test commit:
"drmModeAddFB2 error (1920x1280, AB24, handle 9 pitch 7680) (No such file or directory)"

I think we want to understand this before moving next to master branch. This happens with Android IA, when u use launcher to switch between all opened applications. @js0701  Do you have env setup or @harishkrupo Can you help ?

js0701: really?  I will try in my env and investigate why. I have 2 days trainning and may have some delays

PR #185, opened 2017-12-15
Don't create VA context for each drawing

XiaosongWei:
vaCreateContext takes long time to initialize video driver
so we don't create the context in each drawing.

Jira: OAM-54014
Test: Build passes on both Android & Linux and video playback works

Change-Id: I7814874e0414960116c91839d71d1af902077755
Tracked-On: https://jira01.devtools.intel.com/browse/OAM-54014

### DIFF
--- a/common/compositor/va/varenderer.cpp
+++ b/common/compositor/va/varenderer.cpp
@@ -285,6 +285,11 @@ bool VARenderer::Draw(const MediaState& state, NativeSurface* surface) {
   param.filters = nullptr;
   param.filter_flags = VA_FRAME_PICTURE;
 
+  // currently rotation is only supported by VA on Android.
+#ifdef ANDROID
+  param.rotation_state = HWCTransformToVA(state.layer_->GetTransform());
+#endif
+
   VAProcFilterCapColorBalance vacaps[VAProcColorBalanceCount];
   uint32_t vacaps_num = VAProcColorBalanceCount;
 
@@ -424,6 +429,20 @@ void VARenderer::DestroyContext() {
     vaDestroyConfig(va_display_, va_config_);
     va_config_ = VA_INVALID_ID;
   }
+}
+
+uint32_t VARenderer::HWCTransformToVA(uint32_t transform) {
+  switch (transform) {
+    case kTransform270:
+      return VA_ROTATION_270;
+    case kTransform180:
+      return VA_ROTATION_180;
+    case kTransform90:
+      return VA_ROTATION_90;
+    default:
+      break;
+  }
+  return VA_ROTATION_NONE;
 }
 
 }  // namespace hwcomposer
--- a/common/compositor/va/varenderer.cpp
+++ b/common/compositor/va/varenderer.cpp
@@ -19,8 +19,6 @@
 
 #include <xf86drm.h>
 #include <drm_fourcc.h>
-#include <va/va.h>
-#include <va/va_vpp.h>
 #include <va/va_drmcommon.h>
 #ifdef ANDROID
 #include <va/va_android.h>
@@ -50,6 +48,7 @@ class ScopedVASurfaceID {
   }
 
   bool CreateSurface(int format, VASurfaceAttribExternalBuffers& external) {
+    CTRACE();
     VASurfaceAttrib attribs[2];
     attribs[0].flags = VA_SURFACE_ATTRIB_SETTABLE;
     attribs[0].type = VASurfaceAttribMemoryType;
@@ -79,69 +78,6 @@ class ScopedVASurfaceID {
   VASurfaceID surface_ = VA_INVALID_ID;
 };
 
-class ScopedVAConfigID {
- public:
-  ScopedVAConfigID(VADisplay display) : display_(display) {
-  }
-  ~ScopedVAConfigID() {
-    if (config_ != VA_INVALID_ID) {
-      vaDestroyConfig(display_, config_);
-    }
-  }
-
-  bool CreateConfig(int format) {
-    VAConfigAttrib config_attrib;
-    config_attrib.type = VAConfigAttribRTFormat;
-    config_attrib.value = format;
-    VAStatus ret =
-        vaCreateConfig(display_, VAProfileNone, VAEntrypointVideoProc,
-                       &config_attrib, 1, &config_);
-    return ret == VA_STATUS_SUCCESS ? true : false;
-  }
-
-  operator VAConfigID() const {
-    return config_;
-  }
-
-  VAConfigID config() const {
-    return config_;
-  }
-
- private:
-  VADisplay display_;
-  VAConfigID config_ = VA_INVALID_ID;
-};
-
-class ScopedVAContextID {
- public:
-  ScopedVAContextID(VADisplay display) : display_(display) {
-  }
-  ~ScopedVAContextID() {
-    if (context_ != VA_INVALID_ID) {
-      vaDestroyContext(display_, context_);
-    }
-  }
-
-  bool CreateContext(VAConfigID config, int width, int height,
-                     VASurfaceID render_target) {
-    VAStatus ret = vaCreateContext(display_, config, width, height, 0x00,
-                                   &render_target, 1, &context_);
-    return ret == VA_STATUS_SUCCESS ? true : false;
-  }
-
-  operator VAContextID() const {
-    return context_;
-  }
-
-  VAContextID context() const {
-    return context_;
-  }
-
- private:
-  VADisplay display_;
-  VAContextID context_ = VA_INVALID_ID;
-};
-
 class ScopedVABufferID {
  public:
   ScopedVABufferID(VADisplay display) : display_(display) {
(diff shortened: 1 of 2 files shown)


XiaosongWei:
@kalyankondapally 
@xuguangxin PTAL.
The performance drop is caused by vaCreateContext taking long time to initialize video driver.

common/compositor/va/varenderer.cpp
```diff
@@ -357,4 +296,39 @@ int VARenderer::DrmFormatToRTFormat(int format) {
   return 0;
 }
 
+bool VARenderer::CreateContext() {
+  DestroyContext();
+
```
kalyankondapally: Can we do something similar to  this: ScopedVASurfaceID surface_in(va_display_);

vasurface can allocate this. We anyway have tripple buffer support and recycle them, so we should easily avoid re-creating va surfaces.

common/compositor/va/varenderer.cpp
```diff
@@ -357,4 +296,39 @@ int VARenderer::DrmFormatToRTFormat(int format) {
...
+      vaCreateConfig(va_display_, VAProfileNone, VAEntrypointVideoProc,
+                     &config_attrib, 1, &va_config_);
+  if (ret != VA_STATUS_SUCCESS) {
+    ETRACE("Create VA Config failed\n");
+    return false;
+  }
+
+  // These parameters are not used in vaCreateContext so just set them to dummy
+  // values
+  int width = 1920;
```
kalyankondapally: How about 1 and 1.

kalyankondapally:
Another issue I just found with this backend is that we dont take layer transform into consideration at all. We need to take that into account along with blending.

kalyankondapally:
merged this for now but please look into rotation bug

common/compositor/va/varenderer.cpp
```diff
@@ -251,15 +191,13 @@ bool VARenderer::Draw(const MediaState& state, NativeSurface* surface) {
...
+  if (va_context_ == VA_INVALID_ID || render_target_format_ != rt_format) {
+    DTRACE("to create VA context\n");
+    render_target_format_ = rt_format;
+    if (!CreateContext()) {
+      ETRACE("Create VA context failed\n");
+      return false;
+    }
   }
 
   VAProcPipelineParameterBuffer param;
```
kalyankondapally: You should also be able to set this capabilities only when context is created.

kalyankondapally:
Here are total things, I think we need to do before pushing any further changes to this backend:
1) Fix rotation, so it unblocks our camera team.
2) Don't create VA context for each drawing
3) Create vasurface (out one) in our vasurface and not for every draw.
4)Set capabilities, filters etc only when they have really changed or context needs to be re-initialized.

common/compositor/va/varenderer.cpp
XiaosongWei: vaCreateSurface takes very limited time so we don't need to make thing complicated.

common/compositor/va/varenderer.cpp
XiaosongWei: as the comment said, just set dummy values as the driver impl actually doesn't use these values.

common/compositor/va/varenderer.cpp
XiaosongWei: it won't run to here if context is not created.

XiaosongWei:
the change sounds like WA as it breaks the original VA api behavior. the original behavior is that vaCreateContext should be called for each draw with different w/h & render target. @xuguangxin please comment that why vaCreateContext doesn't use that values in the latest driver and why it doesn't work on Android-IA

XiaosongWei:
rotation should be a new feature, @xuguangxin, please comment that if VA can support it now. We should not mix that in performance fix.

kalyankondapally:
@XiaosongWei Rotation was something we missed when the backend was implemented. The above things I mentioned are the issues I have seen so far and ya agree with you we need to separate them out. Rotation is just top of the list as it's causing functional regressions now. Not sure how context works in Vaapi :( but thats bad as context needs whole bunch of set up.

XiaosongWei:
@kalyankondapally do you have a JIRA for the rotation issue? I will look into that for how VA handle rotation.

kalyankondapally:
@XiaosongWei I pinged you in our internal database, should be camera preview related.

xuguangxin:
@XiaosongWei , the original design for vaCreateContext need register render target. But in practice, no driver care about the param. so the libva upstream changed for api description. https://github.com/01org/libva/commit/492b692005ccd0d8da190209d5b3ae7b7825f4b8#diff-ce40de7930eb5bb0c9f5801e5101ce29L1129

xuguangxin:
the AIA should be work, if not work I can review the code with you. The 1a driver supports rotation too. but AIA does not.

xuguangxin:
@kalyankondapally , why camera team needs rotation. Can they implement them in camera HAL(call libva)?

kalyankondapally:
@xuguangxin This is now a bug in HWC. One can even have the display rotated which would mean we need to rotate all the layers using compositor backend. GL Backend handles all this now but Media layers will not be rotated in this case. If libva cannot support it than when we have transforms applied on Media/camera related layers we would need to fallback to gl/manipulate src/dst co-ordinates. Do you mean when we move AIA to new media driver it can support rotation ? Ir-respective of this for now can you help with adding rotation to this backend, so on 1A we can support rotation and on AIA, we can disable it for now and fallback to 3DCompositor.

Can you please check this code and see if we need to update libva, vaapi for this change to work.

xuguangxin:
Hi @kalyankondapally , yes, when AIA move to the new driver, it will support rotation. No need update libva, The 1a driver already support rotation for a long time. 
@XiaosongWei ,it's very easy to add roration support in current 1a. just one flag.
https://github.com/01org/libva/blob/master/va/va_vpp.h#L476

common/compositor/va/varenderer.cpp
kalyankondapally: Yes usually dummy values are 1 :)

common/compositor/va/varenderer.cpp
kalyankondapally: We already have surface management in place, nothing new which needs to be done specifically for this. vasurface in our side needs to allocate and de-allocate surface appropriately. Nothing else should be needed.

kalyankondapally:
@xuguangxin Can you help check why this patch fails creating context on Android IA.

common/compositor/va/varenderer.cpp
```diff
@@ -453,4 +392,39 @@ int VARenderer::DrmFormatToRTFormat(int format) {
...
+    return false;
+  }
+
+  // These parameters are not used in vaCreateContext so just set them to dummy
+  // values
+  int width = 1920;
+  int height = 1080;
+  VASurfaceID render_target = 0;
+  ret = vaCreateContext(va_display_, va_config_, width, height, 0x00,
+                        &render_target, 1, &va_context_);
```
xuguangxin: need pass NULL and 0 for render target.

XiaosongWei:
added the rotation fix in the 2nd commit, @kalyankondapally @xuguangxin please take a look.

kalyankondapally:
Seems like latest libva has the rotation related changes needed. @xuguangxin We should really align Android IA and 1A libva version. Other wise this becomes painful. I am going to cherry pick rotation fix only to stable so it can land for unblocking camera. For this to land in our master, we should really update Android IA version of libva to include these changes.

kalyankondapally:
patches merged. We should update libva, vappi in Android IA first to enable rotation support by default.

Btwn https://github.com/intel/IA-Hardware-Composer/pull/187/commits/1d69b5d5d1791fb54fa21633f24c04ef996c1ca5

Support to re-use vasurface. @XiaosongWei If you want to review @xuguangxin  you too. One question I have is once surface is created in thread A, can we delete it in thread B or it needs to be in thread A only ?

Caching other redundant state:
https://github.com/intel/IA-Hardware-Composer/pull/187/commits/9530a1f828d9f81f6fd167800b23ce94ea5451fb

@XiaosongWei  @xuguangxin  please review it too

xuguangxin:
The driver problem here is driver not libva version. only 1a driver support rotation.

kalyankondapally:
@xuguangxin That should be k as atleast the interface is their and we dont need compiler flags. Any idea when we are moving to our  new driver in Android IA

xuguangxin:
@kalyankondapally , upgrade libva do not fix the problem. The root cause is the driver, if AIA driver finds the rotation flags not supported, it will report an error. So we still need compile flag.

xuguangxin:
>One question I have is once surface is created in thread A, can we delete it in thread B or it needs to be in thread A only ?
Should support, many projects use a reference count to track the VaSurface. Any thread may decrease the reference count to 0 and destroy the surface. Never heard any issue on this.

kalyankondapally:
@xuguangxin sorry missed your comments. 

Reg compiler flags:
We wouldn't need, it would just show up as a bug. As of today we need as otherwise you cannot even compile the code.

xuguangxin:
@kalyankondapally , compile flag I mean the marco protect the rotation logical, Without AIA driver will report an error at runtime.

PR #186, opened 2017-12-15
Fix for external display hot plug/unplug

pallavigopal:
with current code the nested display is also having the display id of 0
and we are wrongly notifying the plug/unplug of the external display to
the primary display.

Looke like the nested display support is not completed,
So just moved the nested display hotplug notification under the
NESTED_DISPLAY_SUPPORT flag which is not defined

Jira: None
Test: External display Hot plug and unplug works fine

### DIFF
--- a/os/android/iahwc2.cpp
+++ b/os/android/iahwc2.cpp
@@ -230,7 +230,9 @@ HWC2::Error IAHWC2::RegisterCallback(int32_t descriptor,
   switch (callback) {
     case HWC2::Callback::Hotplug: {
       primary_display_.RegisterHotPlugCallback(data, function);
+#ifdef NESTED_DISPLAY_SUPPORT
       nested_display_.RegisterHotPlugCallback(data, function);
+#endif
       for (size_t i = 0; i < size; ++i) {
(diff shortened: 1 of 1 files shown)


pallavigopal:
@mwang30 @kalyankondapally  please review

wsi/drm/drmdisplaymanager.cpp
```diff
@@ -346,11 +346,13 @@ bool DrmDisplayManager::UpdateDisplayState() {
     NotifyClientsOfDisplayChangeStatus();
   }
 
+#ifdef NESTED_DISPLAY_SUPPORT
```
kalyankondapally: This is not the right fix. We need to add this in iahwc2 side and not register hot plug callback in their.

pallavigopal: OK But the NESTED_DISPLAY flag is not added in all the places its purposefully done like that.

pallavigopal:
@kalyankondapally  Changed as per the suggestion and pushed the same. Please review

kalyankondapally:
merged thanks

PR #187, opened 2017-12-17
Scalar

kalyankondapally:
Adds support to take advantage of scalars available at plane level and layers are composited using libva or Mesa


kalyankondapally:
For review and sanity checking:
@Kishore409 If you can also do some basic checks. 
@yugang If you want to also check it from Mosaic, Logical display perspective.
@js0701 If you want to review it.

kalyankondapally:
@js0701 I can confirm that if I disable cache than I dont see any failures related to AB24 buffer which I shared before. Seems like we might have a bug in the key or how we retrieve buffers from cache. This needs to be fixed before master can move to stable/go for full validation.

js0701:
@kalyankondapally  Did you share AB24 buffer failures?  Can you resent how to reproduce it. Thanks

kalyankondapally:
@js0701 If you run Android IA, just play around and you should seem them in logcat

kalyankondapally:
@js0701 
"12-18 05:25:06.862  2147  2320 E iahwcomposer: CreateFrameBuffer: drmModeAddFB2 error (1280x24, AB24, handle 5 pitch 5120) (Invalid argument)
12-18 05:25:06.989  2147  2320 E iahwcomposer: CommitFrame: Failed to commit pset ret=Unknown error -22
"

js0701:
@kalyankondapally 
I read the rebased code, fb is not realsed if owns_gpu_resources_ is false for drm_buffer.
But it should be released each time, I think

js0701:
@kalyankondapally I checked in PR on lastet build
Camera and video can work
no invalid fb log.

I guess ~drmbuffer need to release FB and texture even if owns_gpu_resources_ is false

kalyankondapally:
@js0701 You tested these changes on APL and everything seem to work ?

In general just play around opening/closing views etc which recycles the buffers

js0701:
@kalyankondapally 
Not, I mean in my github https://github.com/js0701/IA-Hardware-Composer.
As I said, the ~drmbuffer() logic may have some problems here

kalyankondapally:
@js0701  How do we make sure the buffer is deleted from cache once its deleted ?

js0701:
@kalyankondapally 
we are using shared_ptr, I think the map just remove the refernce to the buffer, if the buffer is not referenced anymore, it will get deleted

kalyankondapally:
@js0701 That's my understanding too but not sure why I am seeing this error. Can you try with master + this series and see if we are missing something obvious. If you have Android IA env setup, it should be pretty easy to reproduce.

js0701:
@kalyankondapally OK, I will try to debug it

kalyankondapally:
Moving to patches to master.

js0701:
@kalyankondapally 
The code logic is good
but there is a initalization problem
std:memset(&image_, 0, sizeof(image_)); should be added to drm_buffer init
After it, I did not see any error about AddFB error or releasefB error.

kalyankondapally:
@js0701 Thanks

kalyankondapally:
@js0701 I force a full re-validation in case fb creation fails during re-cycling the commit results. This is bad but I guess this was also a bug in our logic as we should have a safe fallback in case things dont go as expected. However, we need to find out why sometimes this fb creation fails :(

kalyankondapally:
@js0701 k Fixed most of the issues caused by surface damage, caching and plane scaling.

Remaining though:
1) When playing Video in full screen, sometimes va reports errors as unable to render things. I see following errors in logcat:
"12-21 00:43:40.406  2138  3632 W GrallocMapperPassthrough: buffer descriptor with invalid usage bits 0x2000
12-21 00:43:40.461  2138  2205 W GrallocMapperPassthrough: buffer descriptor with invalid usage bits 0x2000
"

So, this is a separate issue which needs tackling.

2) drmModeAddFB2 error  still exists.
3) Error while destroying gem_handle still exists.

While I am fine to move these changes (the list has grown a bit) to master, would atleast need to fix 2 before master can move to stable. 2 is something we should also look into before going to stable. 3 is something we can deal later.

@Kishore409 If you want to pick the changes for your build

kalyankondapally:
@js0701 k Fixed most of the issues caused by surface damage, caching and plane scaling.

Remaining though:

    When playing Video in full screen, sometimes va reports errors as unable to render things. I see following errors in logcat:
    "12-21 00:43:40.406 2138 3632 W GrallocMapperPassthrough: buffer descriptor with invalid usage bits 0x2000
    12-21 00:43:40.461 2138 2205 W GrallocMapperPassthrough: buffer descriptor with invalid usage bits 0x2000
    "

So, this is a separate issue which needs tackling.

    drmModeAddFB2 error still exists.
    Error while destroying gem_handle still exists.

While I am fine to move these changes (the list has grown a bit) to master, would atleast need to fix 2 before master can move to stable. 2 is something we should also look into before going to stable. 3 is something we can deal later.

@Kishore409 If you want to pick the changes for your build

common/compositor/va/varenderer.cpp
```diff
@@ -203,15 +203,21 @@ bool VARenderer::Draw(const MediaState& state, NativeSurface* surface) {
...
     if (handle.surface_ != VA_INVALID_ID) {
-      vaDestroySurfaces(va_display_, &handle.surface_, 1);
+      surfaces.emplace_back(handle.surface_);
     }
   }
 
+  if (surfaces.empty())
+    return true;
+
+  vaDestroySurfaces(va_display_, surfaces.data(), surfaces.size());
```
kalyankondapally: @xuguangxin Does this look correct to you, destroying all va surfaces once.

xuguangxin: no, from API level, if you create surface one by one , you need destroy them one by one

kalyankondapally: hmm k

kalyankondapally: thanks will remove this commit

kalyankondapally:
@xuguangxin I think I am able to reproduce u hangs even on SKL with vaapi. Try this branch https://github.com/intel/IA-Hardware-Composer/tree/scalar, for some reason sometimes video hangs, sometime va fails to render when surface size changes dynamically. Ping me on Android IA IRC and I can tell u how to test this.

xuguangxin:
I tried this branch ok KBL , it's very unstable. hard to play any file

kalyankondapally:
@xuguangxin  Can you try with latest code and see. Most of the stability issues are fixed except case where you have media playing and than u navigate back to a video which doesn't have video layers. In this case for some reason we still try to  allocate buffer with video usage flags and gralloc just fails the allocation as some of the usage flags are not applicable to non video formats.

@js0701 Not sure why we are still returning an old buffer.

I guess we just got lucky so far as we always re-created everything, but now with caching enabled it's bringing out bugs like the tracking of layer state correctly which caused lot of regressions on Android IA side atleast.

js0701:
@kalyankondapally 
I am using intel master with my fixes (don't release twice)...
I see cache miss count is 100 while hitcount is 5308.. etc
So I guess cache is good

kalyankondapally:
@js0701 I think we are still returning old buffers in some cases for some reason. Now, with latest code layer get's the usage from Buffer. When we play video and move away from video so that no layer has video, still we get buffers with video usage. I am still debugging that, in case you also want to check it out. With latest code I get createfb allocation failures and gem handle failures only when switching apps using launcher.

I think caching is good, it's just bringing out bugs in rest of the code which just relied on everything being re-created. This just got highlighted when we started adding the scalar support.

js0701:
OK.
The cache has will only get cleaned after 4 frames because we define a constant length 4

kalyankondapally:
I mean we are returning wrong buffer or we are getting a layer from Android with wrong handle, as the buffer says it's still using a format needed for Media when u dont have any video buffers in the frame.

kalyankondapally:
@js0701 If you pull latest branch all crashes should be fixed now. 

However Media usage is having problems:
1) When playing fullscreen after certain frames it complains about va rendering issues @xuguangxin Can you help with this.
2) Sometimes Media shows content from previous frame. Wondering why @js0701 If you want to check this. Try playing videos in youtube.

kalyankondapally:
Moved everything to master @js0701 I think still we have issues getting wrong buffers as I see flickers (showing old content) @xuguangxin If you want to check video issues.

kalyankondapally:
@js0701 Flickers I was seeing was related to not clearing the surface correctly. With that fixed, it's looking good. @xuguangxin Is their anyway to tell va to clear the surface in case it's being recycled ?

PR #188, opened 2017-12-19
Add sharp adjustment on VA module.

feiwan1:
Jira: None.

### DIFF
--- a/common/compositor/compositor.cpp
+++ b/common/compositor/compositor.cpp
@@ -78,6 +78,7 @@ bool Compositor::Draw(DisplayPlaneStateList &comp_planes,
       MediaState &media_state = state.media_state_;
       lock_.lock();
       media_state.colors_ = colors_;
+      media_state.sharp_ = sharp_;
       lock_.unlock();
       const OverlayLayer &layer = layers[plane.source_layers().at(0)];
       media_state.layer_ = &layer;
@@ -209,6 +210,21 @@ void Compositor::GetVideoColor(HWCColorControl /*color*/, float * /*value*/,
 }
 
 void Compositor::RestoreVideoDefaultColor(HWCColorControl /*color*/) {
+  // TODO
+}
+
+void Compositor::SetVideoSharp(float value) {
+  lock_.lock();
+  sharp_ = value;
+  lock_.unlock();
+}
+
+void Compositor::GetVideoSharp(float * /*value*/, float * /*start*/,
+                               float * /*end*/) {
+  // TODO
+}
+
+void Compositor::RestoreVideoDefaultSharp() {
   // TODO
 }
 
--- a/common/display/displayqueue.cpp
+++ b/common/display/displayqueue.cpp
@@ -808,6 +808,24 @@ void DisplayQueue::RestoreVideoDefaultColor(HWCColorControl color) {
   video_lock_.unlock();
 }
 
+void DisplayQueue::SetVideoSharp(float value) {
+  video_lock_.lock();
+  requested_video_effect_ = true;
+  compositor_.SetVideoSharp(value);
+  video_lock_.unlock();
+}
+
+void DisplayQueue::GetVideoSharp(float* value, float* start, float* end) {
+  compositor_.GetVideoSharp(value, start, end);
+}
+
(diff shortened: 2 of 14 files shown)


kalyankondapally:
Can u rebase on top of https://github.com/intel/IA-Hardware-Composer/tree/scalar this branch pls

feiwan1:
@kalyankondapally , what's this branch used for? Do I also need to develop my next deinterlace patch on this branch ?

kalyankondapally:
@feiwan1 These patches are waiting to be merged to master based on test results. I hope to get them merged today. I would have rebased u changes myself but these changes seem to touch similar things which are changed in the above branch. If you can pull latest code and rebase u changes would be great. Sorry for this.

common/display/displayqueue.cpp
```diff
@@ -808,6 +808,24 @@ void DisplayQueue::RestoreVideoDefaultColor(HWCColorControl color) {
...
 
+void DisplayQueue::SetVideoSharp(float value) {
+  video_lock_.lock();
+  requested_video_effect_ = true;
+  compositor_.SetVideoSharp(value);
+  video_lock_.unlock();
+}
+
+void DisplayQueue::GetVideoSharp(float* value, float* start, float* end) {
+  compositor_.GetVideoSharp(value, start, end);
```
kalyankondapally: We want to have lock here too ?

public/nativedisplay.h
```diff
@@ -203,8 +203,27 @@ class NativeDisplay {
...
+  }
+
+  /**
+   * API for setting video sharp in HWC
+   */
+  virtual void SetVideoSharp(float /*value*/) {
+  }
+
+  /**
+       * API for getting video sharp in HWC
```
kalyankondapally: indentation

public/nativedisplay.h
```diff
@@ -203,8 +203,27 @@ class NativeDisplay {
...
+       */
+  virtual void GetVideoSharp(float * /*value*/, float * /*start*/,
+                             float * /*end*/) {
+  }
+
+  /**
+       * API for restoring video default sharp in HWC
```
kalyankondapally: indentation

common/compositor/compositor.cpp
```diff
@@ -212,6 +213,21 @@ void Compositor::RestoreVideoDefaultColor(HWCColorControl /*color*/) {
   // TODO
 }
 
+void Compositor::SetVideoSharp(float value) {
+  lock_.lock();
+  sharp_ = value;
```
kalyankondapally: I am wondering if instead of storing it here, should we push it all the way to CompositorThread and store it their ? Than when rendering media layers we can set to MediaRenderer. We can avoid passing this vai mediastate. Same with colors if possible

common/display/displayqueue.cpp
feiwan1: I copy this from DisplayQueue::GetVideoColor, @XiaosongWei Do we need add lock both of here? If so, I will correct both of them.

common/compositor/compositor.cpp
feiwan1: Er, I think colors actually pass value via MediaState .

kalyankondapally: k, if it varies per media layer than ya. But if not I dont think its needed via Media state

common/display/displayqueue.cpp
kalyankondapally: hmm If setter needs a look than this needs to. If not we should be able to get rid of it from setter.

XiaosongWei:
@feiwan1 Isn't sharpness kind of video color? It's better to just extend HWCColorControl to have kColorSharpness defined.

feiwan1:
@XiaosongWei , For Libva, only HSBC included in colorbalance, and it's provided different api/param for colorbalance and sharp. If we define sharp into color, I can implement it by using if(HSBC)....else if(sharp).... in VARenderer. Indeed, by using this way, it will save tones of hwc api to set/get/restore sharp from top level to VARenderer. How do you think of this?

XiaosongWei:
@feiwan1 From HWC perspective, there's no much different, please keep the api the same.

PR #189, opened 2017-12-20
Fix float value read/write in IControls

XiaosongWei:
Fix the float value read/write in Parcel for IControls

Jira: None
Test: SetColorParam works as expected.

Change-Id: I94625eccc2d001b0ff5d4fe237df7f4f46e8f273
Tracked-On:


kalyankondapally:
merged to master

PR #190, opened 2017-12-20
Should initialize image_ in drmbuffer Android does not automically in…

js0701:
…itialize it to be zeros

Jira: None
Tests: None

### DIFF
--- a/wsi/drm/drmbuffer.cpp
+++ b/wsi/drm/drmbuffer.cpp
@@ -86,6 +86,8 @@ void DrmBuffer::Initialize(const HwcBuffer& bo) {
       is_yuv_ = false;
       total_planes_ = 1;
   }
+
+  std:memset(&image_, 0, sizeof(image_));
 }
 
 void DrmBuffer::InitializeFromNativeHandle(HWCNativeHandle handle,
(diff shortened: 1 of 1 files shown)


js0701:
@kalyankondapally with this change, I did not see release buffer error, or release fb or add fb error

kalyankondapally:
hmm logic is wrong. You are setting handle to null after importing the buffer.

js0701:
right...

wsi/drm/drmbuffer.cpp
```diff
@@ -240,9 +242,7 @@ void DrmBuffer::SetRecommendedFormat(uint32_t format) {
 
 bool DrmBuffer::CreateFrameBuffer(uint32_t gpu_fd) {
   if (image_.drm_fd_) {
-    ResourceHandle temp;
-    temp.drm_fd_ = image_.drm_fd_;
-    resource_manager_->MarkResourceForDeletion(temp, false);
+    return true;
```
kalyankondapally: We can have cases where native surface being recycled to a different plane might want to allocate fb ignoring alpha channel i.e. Primary plane

js0701:
@kalyankondapally 
I guess the problem is we are generate FB and free handles in diffrent thread now
But as we free twice, an live handle maybe closed causing unable to createFB from handles

Avoid to free twice does not have any side effects from my observation
and all the fb and free handle messges is gone.

We may use this and keep eye on it

js0701:
Maybe also we can try to add lock to avoid racing

wsi/drm/drmbuffer.cpp
kalyankondapally: hmm or this is not relevant anymore as we choose prefered format in drmplane and query that when creating the surface. This might be correct thing to do. I checked quickly but still gem handle -1 and frame buffer creation failures exist :(

kalyankondapally:
Merged it to this branch https://github.com/intel/IA-Hardware-Composer/tree/scalar , will push it master once I have test results for this branch.

PR #191, opened 2017-12-21
Add sharp adjustment on VA module.

feiwan1:
Jira: None.


kalyankondapally:
Thanks merged to the branch and will land in master along with other patches.

PR #194, opened 2017-12-22
Flicker

kalyankondapally:
Fix flicker issues seen when navigating between views


kalyankondapally:
@js0701 If you want to review

PR #199, opened 2017-12-28
Add deinterlace on VA module.

feiwan1:
Jira: None.

### DIFF
--- a/common/compositor/va/varenderer.cpp
+++ b/common/compositor/va/varenderer.cpp
@@ -105,6 +105,12 @@ bool VARenderer::SetVAProcFilterColorDefaultValue(
   return true;
 }
 
+bool VARenderer::SetVAProcFilterDeinterlaceDefaultValue() {
+  deinterlace_caps_.mode_ = VAProcDeinterlacingNone;
+  update_caps_ = true;
+  return true;
+}
+
 bool VARenderer::SetVAProcFilterColorValue(HWCColorControl mode, float value) {
   if (mode == HWCColorControl::kColorHue ||
       mode == HWCColorControl::kColorSaturation ||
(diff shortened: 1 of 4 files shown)


kalyankondapally:
@XiaosongWei @xuguangxin ping

common/compositor/va/varenderer.cpp
```diff
@@ -132,6 +138,34 @@ bool VARenderer::SetVAProcFilterColorValue(HWCColorControl mode, float value) {
     return false;
   }
 }
+
+bool VARenderer::SetVAProcFilterDeinterlaceValue(HWCDeinterlaceControl mode) {
```
xuguangxin: this function called for every frame, so you set update_caps_ = true for every frame, then you need to create new deinterlace_ for every frame. this is not necessary.  could we just create/chang deinterlace_ when the value changed? I mean share same deinterlace_ for multiple frames

xuguangxin: I saw the same issue on color balance too

kalyankondapally:
This will need rebasing, also we need to figure out a way to test this outside Android now. Can we expand linux tests to do this ?

xuguangxin:
Hi @kalyankondapally , @XiaosongWei to deliver end2end usage, we need to 
1. Ddd interlace information to gralloc, let media middleware set it
2. Expose deinterlace mode from hwc service. When we pause the video, user may turn on/off the deinterlace mode, they want to see the result instantly.
1 is more urgent since the interface need use by media middleware.

kalyankondapally:
@xuguangxin Why gralloc, does this setting have any influence on how the buffers should be allocated ?

xuguangxin:
I mean minigmb, media stack will set the interlace flag for intelaced frame(not all frame are interlaced). then the hwc need does deinterlace base on the flag. They are one exception if the user sets the force deinterlace mode , hwc will ignore the flag and always do deinterlace.  We need a private function or data structure because google did not define the deinterlace flag in graphic buffer.

kalyankondapally:
hmm we dont need to do that via minigbm, we should propogate that via hwcservice.gralloc is not the place for it.

xuguangxin:
the interlace flag is buffer based, one frame is interlaced, next frame may not interlaced.

kalyankondapally:
hmm unless u need a change in buffer allocation, doesn't make any sense to have it through buffer manager. We need to send it via hwcservice only.

xuguangxin:
No need change in allocation, but media middleware will set the interlace flag after we decode it. You mean we need communicate with hwc in media stack using binder? It may be too heavy. we may need set interlace flag for every frame.

kalyankondapally:
hmm k, how about media initiates the possibility of setting this every frame and passes an fd. hwc will read from this fd, while media framework will write into it. Once done media will inform hwc that their is no need for this and we will stop polling/reading from this fd. I think we can extend this to any other thing i.e. protected content session as desired. Or HWC can create an fd and poll on it. HWC can share that fd with all clients who can write to that fd. Basically a standard server, client communication ?

kalyankondapally:
but anyway this PR can go ahead without that. I will try to come up with a way for this.

feiwan1:
@kalyankondapally, Will you use another PR to implement guangxin's suggestion? If so, I guess I needn't change anything for this patch right ?

xuguangxin:
@kalyankondapally , could you share me related API? I hope we do not need extra IPC for every frame, it will hurt the performance.

kalyankondapally:
@feiwan1  Please go ahead with this PR while we figure out how do we set per buffer configurations separately.

kalyankondapally:
@feiwan1 You need rebasing this patch

feiwan1:
Close this PR, I will use another PR for rebase.

PR #203, opened 2018-01-08
Apply video effect immediately once the effect setting changed

XiaosongWei:
Once the video effect is changed we can't reuse the last offscreen
target and instead let the video layer be rendered immediately to apply
the effect.

Jira: None
Test: Video effect works as expected.

Change-Id: Ic9b52368309e69eae5e14b43390930ec66cb30af
Tracked-On:


XiaosongWei:
@kalyankondapally @feiwan1 PTAL

common/display/displayqueue.cpp
```diff
@@ -563,7 +569,7 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
     bool needs_plane_validation = false;
     GetCachedLayers(layers, remove_index, &current_composition_planes,
                     &render_layers, &can_ignore_commit, &needs_plane_validation,
-                    &validate_layers);
+                    &validate_layers, force_media_composition);
```
kalyankondapally: This should be all managed at https://github.com/intel/IA-Hardware-Composer/blob/master/common/display/displayqueue.cpp#L591 

If this is not doing what it's supposed do, we should fix it their.

XiaosongWei:
@kalyankondapally  updated.

common/display/displayqueue.cpp
```diff
@@ -814,14 +812,20 @@ void DisplayQueue::SetMediaEffectsState(
...
     plane.SetApplyEffects(apply_effects);
     const std::vector<NativeSurface*>& surfaces = plane.GetSurfaces();
     // Handle case where we enable effects but video plane is currently
     // scanned out directly. In this case we will need to ensure we
     // have a offscreen surface to render to.
     if (apply_effects && surfaces.empty()) {
       display_plane_manager_->SetOffScreenPlaneTarget(plane);
-    } else if (!apply_effects && !surfaces.empty() && plane.Scanout()) {
+    } else if (!apply_effects && !surfaces.empty() &&
+               !plane.NeedsOffScreenComposition()) {
```
kalyankondapally: Isnt it same as plane.Scanout()

common/display/displayplanestate.cpp
```diff
@@ -236,6 +236,12 @@ void DisplayPlaneState::ReUseOffScreenTarget() {
   recycled_surface_ = true;
 }
 
+void DisplayPlaneState::DisableReuseOffScreenTarget() {
+  if (recycled_surface_) {
+    recycled_surface_ = false;
```
kalyankondapally: Instead of adding this API,  can we reset this when ever effects state change ?

XiaosongWei:
@kalyankondapally done

XiaosongWei:
DO NOT MERGE, this no longer work after rebase to the latest master. the composition logic changed a lot and I almost got lost with the recent change. The purpose of this patch is to to ensure that video layer got immediately rendered whenever the effect setting changed, even the video layer is not changed (e.g. in the case the video playback is paused). The simplest solution is to revalidate all the layers, but @kalyankondapally you want to fine tune the composition logic, can you please review it and make change accordingly?

@xuguangxin FYI, Video effect setting under PAUSED state still doesn't work.

xuguangxin:
@kalyankondapally could you help check this. 
We counting on this to deliver runtime vpp adjustment. We plan to add this function this week and flow back to APL/KBL next week....
thanks

kalyankondapally:
@XiaosongWei Does this solve it: https://github.com/intel/IA-Hardware-Composer/commit/adc546cfc3f6b9a3ada13d52242249c6fddfe802 ? If not we need to investigate whats going on. Revalidating the whole layers just for this is not an option.

XiaosongWei:
@kalyankondapally it doesn't work, when the video layer is unchanged, the composition returned from GetCachedLayers doesn't contain the video plane.

kalyankondapally:
GetCachedLayers will have video layer only if u are playing video content, video effects being enabled or disabled doesn't make a difference.  This setting should only come into picture if u have a video layer right ?

XiaosongWei:
When the video playback is paused, GetCachedLayers can't get the video plane back.

kalyankondapally:
So, video is paused. Plane layer is not Video layer anymore ? In this case effects will be paused automatically since we dont anyway go through Media compositor if the layer is not type Video right ?

XiaosongWei:
updated to validate layers once the video effect setting is changed. currently in the case video playback is paused, set video effect on/off, the effect can be applied immediately sometimes but sometimes it shows black frame and then effect is on in the next ValidateLayer for idle_frame.

The composition logic now is too complicated, @kalyankondapally Can you please help to check it?

common/display/displayqueue.cpp
```diff
@@ -536,6 +536,8 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
     force_media_composition = true;
     applied_video_effect_ = requested_video_effect_;
     requested_video_effect = requested_video_effect_;
+    validate_layers = true;
```
kalyankondapally: This is really full validation. Hmm we should be doing some fundamental mistake if we don't reset media state immediately. Did you check what happens when video is paused and video effect is disabled, why didn't we call SetMediaEffectsState ?

XiaosongWei: If we dont do full validation, GetCachedLayers will get back only one single plane with all layers GPU composited when video is paused and therefore we don't get a chance to update video plane.

kalyankondapally: k in this case video effects anyway will not be present right ? If all layers are gpu composited to a single plane we would go via the 3D Backend. Also, if we had multiple planes being used and now we suddenly have all layers for one plane than we should have already gone through validation. GetCachedLayers doesn't reset layer + plane combinations in anyway.

kalyankondapally: I suspect if we actually ignore the update because layer damage is zero ? Can you make sure that the layer damage is set to displayframe in overlaylayer in case effects state has changed.

XiaosongWei: The use case is as following:
1) Pause video playback and wait for a moment
2) HWC run into idle state, a ValidateLayers is requested with force_gpu set to true because of idle_frame.
3) previous_plane_state_ now contains a single plane with all layers GPU composited
4) GetCachedLayers check previous_plane_state_ and get back the sole plane
5) Video effect setting changed, but GetCachedLayers doesn't see any video plane.

To ensure a video plane is present again, a full validation is needed.

kalyankondapally: Ya, in this case it's expected that Video effects don't have any impact. What we should do here is, DisplayQueue knows it's in idle mode. If it's in this mode and we get a request to change the video effect state we should put a request for self refresh to SurfaceFlinger. This should make sure the impact is impact and any frame update get combined. Also,  once we prepare idle_frame and the next update should actually call Validate layers to make sure we take optimal path for showing the layers https://github.com/intel/IA-Hardware-Composer/blob/master/common/display/displayqueue.cpp#L584. First check if in this case we have a bug here if not request a refresh to SurfaceFlinger once effect state changes.

xuguangxin:
@kalyankondapally , seems @XiaosongWei  take the vacation now, Who can pick up this? thanks

kalyankondapally:
@xuguangxin Is this something u can test on Android IA, would make life really easy

xuguangxin:
@kalyankondapally , I do not have AIA env, could someone from graphic team help us on this? else I saw big risk for delivery on WW05.

kalyankondapally:
Unfortunately, we haven't root caused the issue here. I will try to provide hypothetical fix, however is your team planning to move new media driver into Android IA and enabling all these features ?

xuguangxin:
No, we do not have resources to cover AIA. And the integration/backporting work will move to Steve after org change. We may need to ask resource from Steve's team for AIA.

kalyankondapally:
@xuguangxin  Can someone in your team try this branch https://github.com/intel/IA-Hardware-Composer/commits/effects and see if it works as expected in this case.

PR #204, opened 2018-01-08
Fix width/height setting in vaCreateSurfaces

js0701:
The origin implementation is using buffer width/height

Tests:video playback sucessful
Jira:None


kalyankondapally:
Thanks for tracking this down.

PR #206, opened 2018-01-11
Support scaling mode for VA render.

feiwan1:
Set scaling mode through environment variable as fast/high quality
and default.

Jira: None

### DIFF
--- a/common/compositor/va/varenderer.cpp
+++ b/common/compositor/va/varenderer.cpp
@@ -27,6 +27,7 @@
 
 #ifdef ANDROID
 #include <va/va_android.h>
+#include <cutils/properties.h>
 #endif
 
 #define ANDROID_DISPLAY_HANDLE 0x18C34078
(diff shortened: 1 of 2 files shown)


feiwan1:
@XiaosongWei @xuguangxin, Please have a review. Thanks.

common/compositor/va/varenderer.cpp
```diff
@@ -42,11 +43,27 @@ VARenderer::~VARenderer() {
...
+      return VA_FILTER_SCALING_DEFAULT;
+  }
+  return VA_FILTER_SCALING_DEFAULT;
+}
 bool VARenderer::Init(int gpu_fd) {
 #ifdef ANDROID
   unsigned int native_display = ANDROID_DISPLAY_HANDLE;
   va_display_ = vaGetDisplay(&native_display);
   UNUSED(gpu_fd);
+  filter_flags_ = GetScalingModeFromEnv();
```
xuguangxin: no need get every time, just get at startup

common/compositor/va/varenderer.cpp
```diff
@@ -42,11 +43,27 @@ VARenderer::~VARenderer() {
   }
 }
 
+static uint32_t GetScalingModeFromEnv() {
+  char value[PROPERTY_VALUE_MAX];
+  property_get("board.hwc.scaling.mode", value, "0");
```
kalyankondapally: This cannot go here but in iahwc side.

common/compositor/va/varenderer.cpp
feiwan1: @xuguangxin , Yes, thie Init just do when startup. But not every time.

common/compositor/va/varenderer.cpp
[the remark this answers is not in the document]
feiwan1: @xuguangxin ，yes, this should be. I will fix it.

feiwan1: @kalyankondapally , If we put this in iahwc class, it will need tons of api to pass only this one flag from iahwc to varender which need to much effort. So I think it's better to keep this here. How do you think of this?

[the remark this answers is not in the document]
kalyankondapally: We don't prefer OS specific flags in core code. Can you add API to NativeDisplay and plumb the setting through ? Check explciit sync support example in iahwc. Even better would be to define enum flags in hwcdefs (i.e. RuntimeSettings) and set them by reading adb properties in iahwc and plumb it through.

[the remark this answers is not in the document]
kalyankondapally: Also, Can you please document what these different scaling modes mean ? We plane to scale content in Display side when we need up scaling. If down scaling is needed than via this renderer.

xuguangxin:
@feiwan1 ,any update on this

xuguangxin:
@kalyankondapally , we do have release pressure on this feature. I discussed with @feiwan1 , Seems it need take some time to move them  iahwc,
Could we merge this first(with comment you suggested) and then @feiwan1 , provide some patches to move it to iahwc?

PR #208, opened 2018-01-18
Enabling Render Buffer Compression support in HWComposer.

YuanjunHuang:
Jira: https://jira01.devtools.intel.com/browse/AREQ-23273
Test: Build passes and validates with hwcomposer testlayer APP.

### DIFF
--- a/common/display/displayplanestate.cpp
+++ b/common/display/displayplanestate.cpp
@@ -487,7 +487,7 @@ bool DisplayPlaneState::Scanout() const {
   return private_data_->state_ == DisplayPlanePrivateState::State::kScanout;
 }
 
-bool DisplayPlaneState::NeedsOffScreenComposition() {
+bool DisplayPlaneState::NeedsOffScreenComposition() const {
   if (private_data_->state_ == DisplayPlanePrivateState::State::kRender)
     return true;
 
--- a/common/display/displayplanestate.h
+++ b/common/display/displayplanestate.h
@@ -166,7 +166,7 @@ class DisplayPlaneState {
 
   // Returns true if offscreen composition
   // is needed for this plane.
-  bool NeedsOffScreenComposition();
+  bool NeedsOffScreenComposition() const;
 
   // Returns type of validation needed by the plane
   // with current source layer. This will be the case


os/android/gralloc1bufferhandler.cpp
```diff
@@ -200,7 +214,15 @@ bool Gralloc1BufferHandler::ImportBuffer(HWCNativeHandle handle) const {
   gralloc1_device_t *gralloc1_dvc =
       reinterpret_cast<gralloc1_device_t *>(device_);
   register_(gralloc1_dvc, handle->imported_handle_);
-  return ImportGraphicsBuffer(handle, fd_);
+  if (ImportGraphicsBuffer(handle, fd_)) {
+    if (handle->rbc_enabled_) {
+      handle->meta_data_.numplanes = 2;
+      handle->meta_data_.modifier = handle->modifier_;
+    }
+  } else {
```
kalyankondapally: you can avoid else here, just return true here and line 225 can return false

wsi/drm/drmbuffer.cpp
```diff
@@ -282,8 +306,17 @@ bool DrmBuffer::CreateFrameBuffer(uint32_t gpu_fd) {
   image_.drm_fd_ = 0;
   media_image_.drm_fd_ = 0;
 
-  int ret = drmModeAddFB2(gpu_fd, width_, height_, frame_buffer_format_,
-                          gem_handles_, pitches_, offsets_, &image_.drm_fd_, 0);
+  int ret = 0;
+  if(modifier_) {
```
kalyankondapally: in drmplane, we would need to do something like this:
during initialization query what kind of modifier is supported for the format:

	for (int i = 0; i < (int)drm.plane->plane->count_formats; i++) {
		if (drm.plane->plane->formats[i] == DRM_FORMAT_XRGB8888)
			format_index = i;
	}

	for (int i = 0; i < (int)drm.plane->props->count_props; i++) {
		if (!strcmp(drm.plane->props_info[i]->name, "IN_FORMATS")) {
			drmModePropertyBlobPtr blob =
					drmModeGetPropertyBlob(drm.fd, drm.plane->props->prop_values[i]);

			struct drm_format_modifier_blob *m = blob->data;
			struct drm_format_modifier *mod = modifiers_ptr(m);

			for (int j = 0; j < (int)m->count_modifiers; j++, mod++) {
				if (mod->formats & (1ULL << format_index))
					drm.count_modifiers++;
			}
			drm.modifiers = calloc(drm.count_modifiers,
					       sizeof(uint64_t));
			mod = modifiers_ptr(m);
			for (int j = 0; j < (int)m->count_modifiers; j++, mod++) {
				if (mod->formats & (1ULL << format_index))
					drm.modifiers[j] = mod->modifier;
			}
		}
	}


You need to store that for each plane. You can check how we store it for rotation here:
https://github.com/intel/IA-Hardware-Composer/blob/master/wsi/drm/drmplane.cpp#L47

We can query from plane if modifier is supported before we try to create the offscreen buffer for that plane in displayplanemanager and also here.

YuanjunHuang: will do.

os/android/gralloc1bufferhandler.cpp
YuanjunHuang: will do.

wsi/drm/drmbuffer.cpp
kalyankondapally: You want to send this first as a separate PR.

YuanjunHuang: @kalyankondapally Do we really need to store the supported modifier for every supported format in this place?  drmModeAddFB2WithModifiers in the underlying driver will check if the modifiers are supported and return error if it is not. And if it is for RBC purpose, we already know DRM_FORMAT_XRGB8888/I915_FORMAT_MOD_Y_TILED_CCS is supported.

kalyankondapally: Let's make the solution generic, so it's easily extendable for all platforms. i.e. platforms which dont have this support. So, ya please store it and than add API to query supported modifier for a given format.

PR #209, opened 2018-01-18
debug down_scale

js0701:
downscale causing some artifacts
maybe it is not good idea for HWC

Change-Id: Ia21bf2e86c60973d5fa521175434603b6996296f

### DIFF
--- a/common/compositor/gl/glrenderer.cpp
+++ b/common/compositor/gl/glrenderer.cpp
@@ -97,24 +97,32 @@ bool GLRenderer::Draw(const std::vector<RenderState> &render_states,
 
   glViewport(left, top, frame_width, frame_height);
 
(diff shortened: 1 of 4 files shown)


js0701:
@kalyankondapally still something wrong. But for scaling, I see artifacts causing by scaling, so maybe we should avoid to do so

kalyankondapally:
pushed rebased patch here:
https://github.com/intel/IA-Hardware-Composer/commit/bad20b61fd61baf65e10c9240415c9853300b347

Height issues are fixed but I see bugs with inputs like if surface is translated either right or left sometimes. Feels like still we have tracking issues somewhere causing mismatch between viewport rect in glrenderer and source rect in displayplanestate.

PR #212, opened 2018-01-18
[AIA-O] Keep BOARD_USES_LIBVA flag in code

sgnanase:
Revert "Enable VA support by default."
This reverts commit 6abcdd05e7c8ba0c8edc1c15606b85ef89dd1d6c.

Jira: None
Test: Device boots to homescreen

### DIFF
--- a/Android.common.mk
+++ b/Android.common.mk
@@ -111,12 +111,14 @@ LOCAL_CPPFLAGS += \
 	-DUSE_GL
 endif
 
+ifneq ($(strip $(DISABLE_MEDIA_COMPOSITOR)), true)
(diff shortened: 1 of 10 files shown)


Android.common.mk
```diff
@@ -111,12 +111,14 @@ LOCAL_CPPFLAGS += \
 	-DUSE_GL
 endif
 
+ifeq ($(strip $(BOARD_USES_LIBVA)), true)
```
kalyankondapally: We want to change this and have something like DISABLE_MEDIA_COMPSITOR and only your target would enable this config.

PR #218, opened 2018-01-23
Add more test case in hwcservice_test.

feiwan1:
1. Add sharpness test case.
2. Add RestoreDefaultDeinterlace implementation

Jira: None.

### DIFF
--- a/common/compositor/va/varenderer.cpp
+++ b/common/compositor/va/varenderer.cpp
@@ -169,6 +169,7 @@ bool VARenderer::GetVAProcDeinterlaceFlagFromVideo(HWCDeinterlaceFlag flag) {
 }
 bool VARenderer::SetVAProcFilterDeinterlaceMode(
     const HWCDeinterlaceProp& prop) {
+  VAProcDeinterlacingType mode;
   if (prop.flag_ == HWCDeinterlaceFlag::kDeinterlaceFlagNone) {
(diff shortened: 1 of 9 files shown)


common/compositor/va/varenderer.cpp
```diff
@@ -169,39 +169,42 @@ bool VARenderer::GetVAProcDeinterlaceFlagFromVideo(HWCDeinterlaceFlag flag) {
...
+          deinterlace_caps_.mode_ != mode) {
+        deinterlace_caps_.mode_ = mode;
+        update_caps_ = true;
+        return true;
+      }
     }
-  }
-  ETRACE("VA Filter unsupport deinterlace mode\n");
-  return false;
+    ETRACE("VA Filter unsupport deinterlace mode\n");
```
kalyankondapally: hmm if  deinterlace caps hasn't changed, we will print this error ?

feiwan1: @kalyankondapally , deinterlace_caps_.caps_[i] saves the supported deinterlace mode queried from LIBVA driver. Before we set a deinterlace mode to driver, we need to know if the mode supported by LIBVA driver, and if not, here will give out a message that LIBVA driver does not support this deinterlace mode.

kalyankondapally:
merged

PR #224, opened 2018-01-31
Layer Hash Generator

harishkrupo:
The LayerHashGenerator are a set of utility methods in NativeDisplay
which can be used to get layer ids. It maintains with a pool of ids whose
initial size is determined by the size parameter passed to
InitializeLayerHashGenerator. Call AcquireId when a new id is required
and release that id with ReleaseId. The size of the pool is variable
and increases when required.

Jira: None
Test:

### DIFF
--- a/public/nativedisplay.h
+++ b/public/nativedisplay.h
@@ -304,6 +304,20 @@ class NativeDisplay {
   virtual void HotPlugUpdate(bool /*connected*/) {
   }
 
+  virtual int InitializeLayerHashGenerator(int) {
+    return 0;
+  }
+
+  virtual uint64_t AcquireId() {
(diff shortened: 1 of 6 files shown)


harishkrupo:
@kalyankondapally

public/nativedisplay.h
```diff
@@ -304,6 +304,20 @@ class NativeDisplay {
   virtual void HotPlugUpdate(bool /*connected*/) {
   }
 
+  virtual int InitializeLayerHashGenerator(int) {
```
kalyankondapally: Document all the APIS.

harishkrupo: sure

PR #225, opened 2018-01-31
NativeDisplay: Layer Hash Generator

harishkrupo:
The LayerHashGenerator are a set of utility methods in NativeDisplay
which can be used to get layer ids. It maintains with a pool of ids whose
initial size is determined by the size parameter passed to
InitializeLayerHashGenerator. Call AcquireId when a new id is required
and release that id with ReleaseId. The size of the pool is variable
and increases when required.

Jira: None
Test:


harishkrupo:
@kalyankondapally Please check if the documentation is adequate.

kalyankondapally:
thanks merged

PR #227, opened 2018-02-05
Correct video plane damage rect caculation

js0701:
Video plane does not need combation of display
rects from layers. Instead it should use the
display rect of its source layer

Change-Id: I6c6e46f96175946b3285cd035cc7d94f9971df71
Jira: None
Test: Youtube

### DIFF
--- a/common/display/displayplanestate.cpp
+++ b/common/display/displayplanestate.cpp
@@ -202,7 +202,15 @@ void DisplayPlaneState::ResetLayers(const std::vector<OverlayLayer> &layers,
 
 void DisplayPlaneState::UpdateDisplayFrame(const HwcRect<int> &display_frame) {
   HwcRect<int> &target_display_frame = private_data_->display_frame_;
-  CalculateRect(display_frame, target_display_frame);
+  if (!IsVideoPlane())
+    CalculateRect(display_frame, target_display_frame);
+  else {
+    target_display_frame.bounds[0] = display_frame.bounds[0];
+    target_display_frame.bounds[1] = display_frame.bounds[1];
+    target_display_frame.bounds[2] = display_frame.bounds[2];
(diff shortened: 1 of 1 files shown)


common/display/displayplanestate.cpp
```diff
@@ -202,7 +202,15 @@ void DisplayPlaneState::ResetLayers(const std::vector<OverlayLayer> &layers,
 
 void DisplayPlaneState::UpdateDisplayFrame(const HwcRect<int> &display_frame) {
   HwcRect<int> &target_display_frame = private_data_->display_frame_;
-  CalculateRect(display_frame, target_display_frame);
+  if (!IsVideoPlane())
+    CalculateRect(display_frame, target_display_frame);
+  else {
+    target_display_frame.bounds[0] = display_frame.bounds[0];
```
kalyankondapally: Is this solving any Bug ?

kalyankondapally: Hmm, instead of IsVideoPlane, the check should be private_data_->source_layers_.size() == 1, you can just do target_display_frame = display_frame;

common/display/displayplanestate.cpp
```diff
@@ -202,7 +202,15 @@ void DisplayPlaneState::ResetLayers(const std::vector<OverlayLayer> &layers,
...
+    target_display_frame.bounds[1] = display_frame.bounds[1];
```
kalyankondapally: Same with UpdateSourceCrop ?

common/display/displayplanestate.cpp
kalyankondapally: Hmm, instead of IsVideoPlane, the check should be private_data_->source_layers_.size() == 1, you can just do target_display_frame = display_frame;

js0701:
It will resolve youtube shrink bug.
Will rebase the code

PR #231, opened 2018-02-08
VARenderer: add surface sync

js0701:
We rely on fence for GPU/display sync
But for varenderer, currently there is not fence mechanism
And vaEndPicture does not sync surface.
So Need to add sync at the end.

Change-Id: Ied562741b8ff7d7a9170969cdd2038befd99be6b
Jira:None
Tests:None


js0701:
@kalyankondapally can you review and let Guanxin review

js0701:
Will close as it does not have a bug

PR #239, opened 2018-02-15
Check for setting offscreen target when there is a fall back

shwethabolumbu:
We need to check if the offscreen target is set properly for the plane, when we are handling more layers in it.

JIRA: OAM-54054

Test: No flicker when we touch/mouse click during video play back.

Signed-off-by: bshwetha <shwetha.b@intel.com>

### DIFF
--- a/common/display/displayplanemanager.cpp
+++ b/common/display/displayplanemanager.cpp
@@ -188,7 +188,10 @@ bool DisplayPlaneManager::ValidateLayers(
 
       last_plane.AddLayer(i->GetZorder(), i->GetDisplayFrame(), false);
     }
-
+    if (!last_plane.GetOffScreenTarget() || last_plane.GetSurfaces().size() < 3) {
+    SetOffScreenPlaneTarget(last_plane);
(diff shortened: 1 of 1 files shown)


common/display/displayplanemanager.cpp
```diff
@@ -188,7 +188,10 @@ bool DisplayPlaneManager::ValidateLayers(
 
       last_plane.AddLayer(i->GetZorder(), i->GetDisplayFrame(), false);
     }
-
+    if (!last_plane.GetOffScreenTarget() || last_plane.GetSurfaces().size() < 3) {
+    SetOffScreenPlaneTarget(last_plane);
```
kalyankondapally: Below we already do           if (fall_back) {
            ResetPlaneTarget(last_plane, commit_planes.back());
            validate_final_layers = true;
          }

perhaps last_plane.SwapSurfaceIfNeeded(); should be called after thatt ?

kalyankondapally:
closing this, please validate against master

PR #244, opened 2018-02-19
Fix travis builds

harishkrupo:
Install python3-pip as pip3 is required to install meson.


kalyankondapally:
merged

PR #253, opened 2018-03-06
Remove opengl, vulkan, va use for dummy compositor

csravelar:
Adjust makefiles in order to build without certain functionality that we won't need enabled in the splash screen with dummy compositor flag set. 

Jira: None.
Test: Build passes when using dummy compositor flag

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,6 +76,9 @@ script:
 - make clean
 - ./autogen.sh --prefix=$WLD --disable-hotplug-support
 - make -j5
+- make clean
+- ./autogen.sh --prefix=$WLD --enable-dummy-compositor
+- make -j5
 
 # Test Weston plugin
 - export WLD=/tmp/wl-install
@@ -149,14 +152,6 @@ script:
 - git clean -xfd
 - ./autogen.sh --prefix=$WLD
 - make -j5 && make install
-- popd
-
-#Build hwcomposer with dummy compositor enabled
-- pushd .
-- cd /tmp/common
-- git clean -xfd
-- ./autogen.sh --prefix=$WLD --enable-dummy-compositor
-- make && make install
 - popd
 
 #Build weston and weston plugin
--- a/.travis.yml
+++ b/.travis.yml
@@ -151,6 +151,14 @@ script:
 - make -j5 && make install
 - popd
 
+#Build hwcomposer with dummy compositor enabled
+- pushd .
+- cd /tmp/common
+- git clean -xfd
+- ./autogen.sh --prefix=$WLD --enable-dummy-compositor
+- make && make install
+- popd
+
 #Build weston and weston plugin
 - git clone https://anongit.freedesktop.org/git/wayland/weston.git /tmp/weston
 - os/linux/weston/build_script.sh --apply-patches --weston-dir=/tmp/weston --iahwc-dir=$(pwd) --build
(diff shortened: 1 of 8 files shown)


common/compositor/compositordefs.h
```diff
@@ -19,7 +19,10 @@
 
 #include <stdint.h>
 
-#ifdef USE_GL
+#ifdef USE_DC
+#include <EGL/egl.h>
```
kalyankondapally: Why do you need these ?

common/compositor/compositordefs.h
```diff
@@ -42,7 +45,15 @@ static float TransformMatrices[] = {
 };
 // clang-format on
 
-#ifdef USE_GL
+#ifdef USE_DC
+typedef unsigned GpuResourceHandle;
+typedef struct gl_import {
+  EGLImageKHR image_ = 0;
```
kalyankondapally: do u need image_ ?

common/compositor/compositordefs.h
```diff
@@ -42,7 +45,15 @@ static float TransformMatrices[] = {
...
+  HWCNativeHandle handle_ = 0;
+  uint32_t drm_fd_ = 0;
+} ResourceHandle;
+typedef EGLDisplay GpuDisplay;
```
kalyankondapally: This can be just void*

kalyankondapally:
Also seems like defautlt linux build is broke. And we want to add another build option in our travis config file to test it with --enable-dummy-compositor

kalyankondapally:
https://github.com/intel/IA-Hardware-Composer/blob/master/.travis.yml#L133 you can check how we have different builds for mesa, we should add one for hwcomposer with this option enabled

common/compositor/compositordefs.h
strassek: I believe this is referenced outside of the USE_GL block in GetGpuResource(), I've been seeing similar issues with the Vulkan build.

kalyankondapally: we should fix that

common/compositor/compositordefs.h
csravelar: Many errors occur with the EGLImage and Gpu Resource Handle if these aren't added for the struct EGLDisplay and GpuDisplay. Compositorthreads, compositor, and nativesurface try to access these struct values regardless of using opengl, vulkan, or not and don't both checking if the struct has these attributes available. They try to use them and this results in errors in multiple places. I added them initially to quite these errors since files require them regardless of what is being used.

.travis.yml
```diff
@@ -151,6 +151,14 @@ script:
 - make -j5 && make install
 - popd
 
+#Build hwcomposer with dummy compositor enabled
```
kalyankondapally: You can move this to # finally, build the hardware composer itself part

common/compositor/compositordefs.h
kalyankondapally: You should have here void* image_ = 0; instead of EGLImageKHR

common/compositor/compositordefs.h
kalyankondapally: If you address comments below and change things to void* you will not need this.

kalyankondapally:
Can u squash both patches to 1?

kalyankondapally:
Also please check the commit message format here: https://github.com/intel/IA-Hardware-Composer/wiki/Contributions

PR #257, opened 2018-03-07
Configure Custom Resolution

gowthamhosamane:
Setting the custom resolution as configured instead of display default


wsi/physicaldisplay.cpp
```diff
@@ -555,6 +557,43 @@ bool PhysicalDisplay::GetDisplayAttribute(uint32_t /*config*/,
   return true;
 }
 
+/* Setting custom resolution instead of preferred as fetched from display */
+bool PhysicalDisplay::SetResolution(const HwcRect<int32_t> &rect) {
+  if ((rect.right - rect.left) && (rect.bottom - rect.top)) {
```
kalyankondapally: u can just use rect.empty() here

common/core/gpudevice.cpp
```diff
@@ -518,6 +560,50 @@ void GpuDevice::HandleHWCSettings() {
     temp_displays.swap(total_displays_);
   }
 
+  // Now set floating display configuration
+  // Get the floating display index and the respective rectangle
+  // TODO Logical display on & mosaic display on scenario
```
kalyankondapally: I think we shouldn't support this feature for these display modes. Any reason we want to ?

common/core/gpudevice.cpp
```diff
@@ -324,6 +334,38 @@ void GpuDevice::HandleHWCSettings() {
...
+            size_t float_rect_val = atoi(float_rect_str.c_str());
+
+            // Save the rectangle - left, top, right & bottom
+            float_rect.emplace_back(float_rect_val);
+          }
+
+          // If entire rect is available, store the parameters
+	  // TODO remove hard code
+	  if (float_rect.size() == 4) {
+	    float_display_indices.emplace_back(index);
```
kalyankondapally: indentation issues git diff | clang-format-diff-3.5 -p 1 -style=file -i should fix these

wsi/physicaldisplay.cpp
gowthamhosamane: Default values are set all the time. Also even if left and right or top and bottom are same, the code should handle. So, empty may not work in that case.

common/core/gpudevice.cpp
```diff
@@ -518,6 +560,50 @@ void GpuDevice::HandleHWCSettings() {
...
+  // Get the floating display index and the respective rectangle
+  // TODO Logical display on & mosaic display on scenario
+  if (use_float && !use_logical && !use_mosaic) {
+    size_t size = float_display_indices.size();
+    bool ret = false;
+
+    // Set custom resolution to desired display instance
+    for (size_t i = 0; i < size; i++) {
+      std::vector<int32_t> &float_rect = float_displays.at(i);
+      HwcRect<int32_t> rect = HwcRect<int32_t>(float_rect.at(0),
```
kalyankondapally: Instead of here, you can do it line 367 above ?

common/core/gpudevice.cpp
```diff
@@ -518,6 +560,50 @@ void GpuDevice::HandleHWCSettings() {
...
+      }
+    }
+  } else {
+    // Set dummy resolution to other display instances
+    // so that flags are set consistently
+    // TODO remove dummy setting
+    HwcRect<int32_t> rect_dummy = HwcRect<int32_t>(0,0,0,0);
+    bool ret = false;
+    for (size_t i = 0; i < total_displays_.size(); i++) {
+      ret = total_displays_.at(i)->SetResolution(rect_dummy);
```
kalyankondapally: Do we really need this, if default flag in PhysicalDisplay is false ?

common/display/virtualdisplay.h
```diff
@@ -41,6 +41,10 @@ class VirtualDisplay : public NativeDisplay {
 
   bool SetActiveConfig(uint32_t config) override;
 
+  bool SetResolution(const HwcRect<int32_t>& ) override {
+    return false;
```
kalyankondapally: We could have a default implementation in NativeDisplay returning false and we can avoid overrides in multiple inherited classes here

common/core/gpudevice.cpp
gowthamhosamane: total displays won't be updated at 370. Both ordering and physical connectivity of displays happen after 400.

wsi/physicaldisplay.cpp
kalyankondapally: 1,1,1,1 in this case is the expectation to use full resolution than ?

common/core/gpudevice.cpp
kalyankondapally: Sorry, I mean instead of converting vector to rect, you could do that at line 370

common/core/gpudevice.cpp
gowthamhosamane: We should be able to remove from line 578 till 603. Will do that.

wsi/physicaldisplay.cpp
gowthamhosamane: Yes, as 1,1,1,1 is not a valid rectangle.

common/core/gpudevice.cpp
gowthamhosamane: Oh yes, that can be done. I will do that.

common/display/virtualdisplay.h
gowthamhosamane: Yes, will do that. Multiple inherited classes may be extended as needed in future.

common/core/gpudevice.cpp
gowthamhosamane: This feature is needed for logical display for some use-cases. Mosaic, may not be needed.

public/nativedisplay.h
```diff
@@ -87,6 +87,9 @@ class NativeDisplay {
   virtual int GetDisplayPipe() = 0;
   virtual bool SetActiveConfig(uint32_t config) = 0;
   virtual bool GetActiveConfig(uint32_t *config) = 0;
+  virtual bool SetResolution(const HwcRect<int32_t> &) {
```
kalyankondapally: Can u add documentation :)

wsi/physicaldisplay.cpp
```diff
@@ -37,8 +37,10 @@ PhysicalDisplay::PhysicalDisplay(uint32_t gpu_fd, uint32_t pipe_id)
     : pipe_(pipe_id),
       width_(0),
       height_(0),
+      custom_resolution_(false),
       gpu_fd_(gpu_fd),
       power_mode_(kOn) {
+  rect_.left = rect_.top = rect_.right = rect_.bottom = 0;
```
kalyankondapally: It's initialized to zero, no need to do again

wsi/physicaldisplay.h
```diff
@@ -63,6 +63,10 @@ class PhysicalDisplay : public NativeDisplay, public DisplayPlaneHandler {
   bool SetActiveConfig(uint32_t config) override;
   bool GetActiveConfig(uint32_t *config) override;
 
+  bool SetResolution(const HwcRect<int32_t>&) override;
+  bool IsCustomResolution();
+  const HwcRect<int32_t> &GetResolution();
```
kalyankondapally: GetCustomResolution or GetFloatingRect ? same for setter ?

wsi/physicaldisplay.h
```diff
@@ -63,6 +63,10 @@ class PhysicalDisplay : public NativeDisplay, public DisplayPlaneHandler {
   bool SetActiveConfig(uint32_t config) override;
   bool GetActiveConfig(uint32_t *config) override;
 
+  bool SetResolution(const HwcRect<int32_t>&) override;
+  bool IsCustomResolution();
```
kalyankondapally: Do we need this ?

hwc_display.ini
```diff
@@ -25,6 +26,11 @@ LOGICAL_DISPLAY="0:3"
 # sub-display-index: start from 0, should be the available displays number, follow the order of the connected logical displays
 MOSAIC_DISPLAY="0+1+2"
 
+# Float display, with format "sub-display-index:left+top+right+bottom"
+# sub-display-index: start from 0, should be the available displays number, follow the order of the connected logical displays
```
kalyankondapally: We should add that this becomes the resolution of the display irrespective of the modeset ?

wsi/physicaldisplay.h
gowthamhosamane: Are you commenting on 'override' keyword?

hwc_display.ini
gowthamhosamane: Didn't get what did you mean by irrespective of modeset. If FLOAT=true then only it becomes the resolution. Esample, if LOGICAL=true then it becomes on top of it. (future)

wsi/physicaldisplay.cpp
gowthamhosamane: Should I remove? I was not sure because even the width_ and height_ were getting initialized to 0.

wsi/physicaldisplay.h
gowthamhosamane: I can change to GetCustomResolution and same for set.

hwc_display.ini
kalyankondapally: I mean ir-respective of display resolution being 1920x1080, this becomes the resolution of display in this case

wsi/physicaldisplay.h
kalyankondapally: No this API bool IsCustomResolution();

wsi/physicaldisplay.cpp
kalyankondapally: ya HWCRect initializes everything to 0 in constructor

public/nativedisplay.h
gowthamhosamane: Added documentation

wsi/physicaldisplay.cpp
gowthamhosamane: Removed initialization to 0

wsi/physicaldisplay.h
gowthamhosamane: Yes, for cleaner implementation, this is required

hwc_display.ini
gowthamhosamane: Elaborated the description

common/core/gpudevice.cpp
```diff
@@ -518,6 +563,26 @@ void GpuDevice::HandleHWCSettings() {
...
+    size_t size = float_display_indices.size();
+    size_t num_displays = total_displays_.size();
+
+    // Set custom resolution to desired display instance
+    for (size_t i = 0; i < size; i++) {
+      int index = float_display_indices.at(i);
+
+      // Ignore float index if out of range of connected displays
+      if (index < num_displays) {
+        HwcRect<int32_t> &rect = float_displays.at(i);
```
kalyankondapally: this can be const

common/core/gpudevice.cpp
```diff
@@ -324,6 +334,41 @@ void GpuDevice::HandleHWCSettings() {
...
+            std::string::npos)
+            continue;
+
+          int32_t index = atoi(index_str.c_str());
+
+          // Got rectangle configuration
+          while (std::getline(i_value, float_rect_str, '+')) {
+            if (float_rect_str.empty() ||
+                float_rect_str.find_first_not_of("0123456789") !=
+                    std::string::npos)
```
kalyankondapally: { } just to separate the continue below from rest

common/core/gpudevice.cpp
gowthamhosamane: const added

common/core/gpudevice.cpp
gowthamhosamane: Flower braces added

wsi/drm/drmdisplay.cpp
```diff
@@ -403,8 +434,19 @@ void DrmDisplay::SetDrmModeInfo(const std::vector<drmModeModeInfo> &mode_info) {
 }
 
 void DrmDisplay::SetDisplayAttribute(const drmModeModeInfo &mode_info) {
-  width_ = mode_info.hdisplay;
-  height_ = mode_info.vdisplay;
+  // Default resolution of the display
+  if (!IsCustomResolution()) {
```
kalyankondapally: I am still debating the usage of this and GetCustomResolution() in drmdisplay. This is inherited from PhysicalDisplay and should have access to all the protected variables. I think we can just use the variables directly here

wsi/drm/drmdisplay.cpp
```diff
@@ -162,16 +162,36 @@ bool DrmDisplay::GetDisplayAttribute(uint32_t config /*config*/,
 
   float refresh;
   bool status = true;
+
+  // Check if different resolution is configured
+  bool custom_resolution = IsCustomResolution();
```
kalyankondapally: DrmDisplay is inherited from PhysicalDisplay and should have access to all the protected variables. I think we can just use the variables directly here ?

gowthamhosamane: Technically yes, but accessing base class variable directly instead of through methods is not something I would prefer. It helps in debugging as well.

wsi/drm/drmdisplay.cpp
gowthamhosamane: Accessing variables of super class in derived class directly instead of defining a method to refer that is something debatable. It makes the code difficult to understand and hard to maintain going forward. I would prefer methods in such scenarios.

gowthamhosamane: If you still think otherwise, then we can change to either private variable or access directly.

kalyankondapally: access directly is fine as we do the same for others.

gowthamhosamane: Accessed directly now!

kalyankondapally:
look good, can u update commit message though: https://github.com/intel/IA-Hardware-Composer/wiki/Contributions

kalyankondapally:
merged

PR #260, opened 2018-03-08
Handle layer transform in VARenderer

Kishore409:
VA only support 90/180/270 rotation so we only handle these transforms
in VARenderer.

Jira: None.
Test: Video rotation should work


kalyankondapally:
merged

PR #263, opened 2018-03-12
Add API to enable/disable HDCP support from HWCService API.

kalyankondapally:
Jira: None.
Test: We are able to toggle HDCP state via the API tested by
      hwcservice_test.

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -524,6 +524,44 @@ void GpuDevice::HandleHWCSettings() {
   initialization_state_lock_.unlock();
 }
 
+void GpuDevice::EnableHDCPSessionForDisplay(uint32_t display,
+                                            HWCContentType content_type) {
+  if (total_displays_.size() <= display) {
+    ETRACE("Tried to enable HDCP for invalid display %u \n", display);
+    return;
+  }
+
+  NativeDisplay *native_display = total_displays_.at(display);
+  native_display->SetHDCPState(HWCContentProtection::kDesired, content_type);
+}
+
+void GpuDevice::EnableHDCPSessionForAllDisplays(HWCContentType content_type) {
+  size_t size = total_displays_.size();
+  for (size_t i = 0; i < size; i++) {
+    total_displays_.at(i)
+        ->SetHDCPState(HWCContentProtection::kDesired, content_type);
+  }
+}
+
+void GpuDevice::DisableHDCPSessionForDisplay(uint32_t display) {
+  if (total_displays_.size() <= display) {
+    ETRACE("Tried to enable HDCP for invalid display %u \n", display);
+    return;
+  }
+
+  NativeDisplay *native_display = total_displays_.at(display);
+  native_display->SetHDCPState(HWCContentProtection::kUnDesired,
+                               HWCContentType::kInvalid);
+}
+
+void GpuDevice::DisableHDCPSessionForAllDisplays() {
+  size_t size = total_displays_.size();
+  for (size_t i = 0; i < size; i++) {
+    total_displays_.at(i)->SetHDCPState(HWCContentProtection::kUnDesired,
+                                        HWCContentType::kInvalid);
+  }
+}
+
 void GpuDevice::InitializeHotPlugEvents(bool take_lock) {
   initialization_state_lock_.lock();
   if ((initialization_state_ & kInitializeHotPlugMonitor) ||
--- a/os/android/hwcservice.cpp
+++ b/os/android/hwcservice.cpp
@@ -364,6 +364,28 @@ status_t HwcService::Controls::DisplayModeSetMode(uint32_t display,
   return OK;
 }
 
+status_t HwcService::Controls::EnableHDCPSessionForDisplay(
+    uint32_t display, EHwcsContentType content_type) {
+  mHwc.EnableHDCPSessionForDisplay(display, content_type);
+  return OK;
+}
+
+status_t HwcService::Controls::EnableHDCPSessionForAllDisplays(
+    EHwcsContentType content_type) {
+  mHwc.EnableHDCPSessionForAllDisplays(content_type);
+  return OK;
+}
+
+status_t HwcService::Controls::DisableHDCPSessionForDisplay(uint32_t display) {
+  mHwc.DisableHDCPSessionForDisplay(display);
+  return OK;
+}
+
+status_t HwcService::Controls::DisableHDCPSessionForAllDisplays() {
+  mHwc.DisableHDCPSessionForAllDisplays();
+  return OK;
+}
+
 status_t HwcService::Controls::VideoEnableEncryptedSession(
     uint32_t sessionID, uint32_t instanceID) {
   // TO DO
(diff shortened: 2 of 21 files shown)


os/android/libhwcservice/hwcserviceapi.h
```diff
@@ -169,6 +152,35 @@ status_t HwcService_DisplayMode_SetMode(HWCSHANDLE hwcs, uint32_t display,
 
 // VideoControl
 
+// The control enables the usage of HDCP for all planes supporting this feature
+// on display. Some displays can support latest HDCP specification and also
+// have ability to fallback to older specifications i.e. HDCP 2.2 and 1.4
+// in case latest speicification cannot be supported for some reason. Fallback
```
MicheleL-Intel: fix typo 'specification'

os/android/libhwcservice/hwcserviceapi.h
```diff
@@ -169,6 +152,35 @@ status_t HwcService_DisplayMode_SetMode(HWCSHANDLE hwcs, uint32_t display,
...
+status_t HwcService_Video_EnableHDCPSession_ForDisplay(HWCSHANDLE hwcs,
+                                                       uint32_t display,
+                                                       bool fallback);
+
+// The control enables the usage of HDCP for all planes supporting this feature
+// on all connected displays. Some displays can support latest HDCP
+// specification
+// and also have ability to fallback to older specifications i.e. HDCP 2.2 and
+// 1.4
+// in case latest speicification cannot be supported for some reason. Fallback
```
MicheleL-Intel: Typo 'specification' and remove extraneous carriage returns or line feeds.

common/core/gpudevice.cpp
```diff
@@ -524,6 +524,43 @@ void GpuDevice::HandleHWCSettings() {
...
 
+void GpuDevice::EnableHDCPSessionForDisplay(uint32_t display,
+                                            bool /*fallback*/) {
+  size_t size = total_displays_.size();
+  if (size <= display) {
+    ETRACE("Tried to enable HDCP for invalid display %d \n", display);
+    return;
+  }
+
+  NativeDisplay *native_display = total_displays_.at(display);
```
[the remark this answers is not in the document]
kalyankondapally: This should be handled by above checks where we ensure index is not greater than the size of vector

os/android/hwcservice.cpp
```diff
@@ -364,6 +364,27 @@ status_t HwcService::Controls::DisplayModeSetMode(uint32_t display,
   return OK;
 }
 
+status_t HwcService::Controls::EnableHDCPSessionForDisplay(uint32_t display,
+                                                           bool fallback) {
+  mHwc.EnableHDCPSessionForDisplay(display, fallback);
```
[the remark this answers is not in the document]
kalyankondapally: Handshake may or may not happen during the call. This is just acknowledgement that we have received the request and will act upon in the next frame

kalyankondapally:
@strassek @csravelar ping if you want also to give u feedback.

os/android/libhwcservice/test/hwcservice_test.cpp
```diff
@@ -28,6 +28,10 @@ using namespace hwcomposer;
 static void usage() {
   aout << "Usage: hwcservice_test \n"
           "\t-g: Get current display mode\n"
+          "\t-h: Enable HDCP support for a given Display. \n"
+          "\t-i: Disable HDCP support for a given Display. \n"
+          "\t-j: Enable HDCP support for all displays. \n"
+          "\t-k: Disable HDCP support for all displays. \n"
```
strassek: Drop trailing whitespace

os/android/libhwcservice/hwcserviceapi.h
```diff
@@ -127,23 +127,6 @@ status_t HwcService_Display_SetDeinterlaceParam(HWCSHANDLE hwcs,
...
-
-/// Enumerations for common aspect ratios
-/// Any ratio can be supported, with the upper 16 bits containing one dimension,
-/// the lower 16 bits contains the lower dimension
-typedef enum _EHwcsModeAspectRatio {
-  HWCS_MODE_ASPECT_RATIO_ANY = 0x00000000,
-  HWCS_MODE_ASPECT_RATIO_4_3 = 0x00040003,
-  HWCS_MODE_ASPECT_RATIO_16_9 = 0x00100009,
-} EHwcsModeAspectRatio;
-
```
strassek: Why is this block getting removed? Are you cleaning up stale code? If so, could we put this in a separate patch?

common/core/gpudevice.cpp
```diff
@@ -524,6 +524,43 @@ void GpuDevice::HandleHWCSettings() {
   initialization_state_lock_.unlock();
 }
 
+void GpuDevice::EnableHDCPSessionForDisplay(uint32_t display,
+                                            bool /*fallback*/) {
+  size_t size = total_displays_.size();
```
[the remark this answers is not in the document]
strassek: Also `size` is only used in 1 place. Could we rewrite the conditional as `if (display > total_displays_.size()) {`?

os/android/libhwcservice/hwcserviceapi.cpp
```diff
@@ -221,6 +221,46 @@ status_t HwcService_DisplayMode_SetMode(HWCSHANDLE hwcs, uint32_t display,
   return pContext->mControls->DisplayModeSetMode(display, config);
 }
 
+status_t HwcService_Video_EnableHDCPSession_ForDisplay(HWCSHANDLE hwcs,
+                                                       uint32_t display,
+                                                       bool fallback) {
+  HwcsContext* pContext = static_cast<HwcsContext*>(hwcs);
+  if (!pContext) {
+    return -1;
```
strassek: Given that we are using status_t return type, we ought to use a real return code. I'd suggest `BAD_VALUE`.

kalyankondapally:
@ramalingampc2008 k thanks for the info.

os/android/libhwcservice/hwcserviceapi.h
kalyankondapally: dropped from here

os/android/libhwcservice/hwcserviceapi.cpp
kalyankondapally: ackowledeged, however I would like to keep it consistent with other API and do all clean up once. Added task here, if anyone wants to try this out: https://github.com/intel/IA-Hardware-Composer/issues/268

kalyankondapally:
@ramalingampc2008 updated

MicheleL-Intel:
One issue is that, 'BAD_VALUE' is defined under android namespace, whereas 'status_t' is an 'int'.  If the return type is changed to 'BAD_VALUE' for consistency across the board, then we're confining the return values to strictly android type defines, which is contradictory to IA-HW-Composer being OS-agnostic.  

One possible solution will be to change 'status_t' into an 'enum' to include 'BAD_VALUE'.  This will also minimize the number of changes, including limiting changes to function prototype definitions.

MicheleL-Intel:
Just realized this particular change is strictly for android.

kalyankondapally:
@MicheleL-Intel Future this can be extended outside Android. If you plan to fix the errors, you can come up with a generic sol.

kalyankondapally:
@MicheleL-Intel Also using uint8_t appropriately could go into a separate task, so we unify all API at same time.

kalyankondapally:
merged

PR #264, opened 2018-03-13
Remove opengl, vulkan, va use for dummy compositor

csravelar:
Adjust makefiles in order to build without certain functionality that
we won't need enabled in the splash screen with dummy compositor flag set.

Jira: None.
Test: Build passes when using dummy compositor flag

### DIFF
--- a/common/compositor/factory.cpp
+++ b/common/compositor/factory.cpp
@@ -58,7 +58,11 @@ Renderer* Create3DRenderer() {
 }
 
 Renderer* CreateMediaRenderer() {
+#ifdef USE_DC
+  return NULL;
+#else
   return new VARenderer();
+#endif
 }
 
 NativeGpuResource* CreateNativeGpuResourceHandler() {


configure.ac
```diff
@@ -42,6 +42,17 @@ AC_PROG_LIBTOOL
 
 AC_SUBST(ASSEMBLER_WARN_CFLAGS)
 
+# For splash screen dummy compositor
+AC_ARG_ENABLE(dummy-compositor,
+  AS_HELP_STRING([--enable-dummy-compositor],
```
strassek: The issue I see here is that each backend is mutually exclusive. What would happen if I passed both "--enable-vulkan" and "--enable-dummy-compositor"? I think a better approach would be to have a '--with-compositor=' option that could take one of "gl,vulkan,dummy", at least until we get runtime selection working.

common/compositor/compositordefs.h
```diff
@@ -42,7 +42,15 @@ static float TransformMatrices[] = {
 };
 // clang-format on
 
-#ifdef USE_GL
+#ifdef USE_DC
+typedef unsigned GpuResourceHandle;
+typedef struct gl_import {
```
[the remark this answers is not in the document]
kalyankondapally: hmm ya makes sense.

configure.ac
kalyankondapally: hmm good point. @csravelar Can you do that change ?

kalyankondapally: Or other way around would be to have checks that only one backend is requested and throw out an error if thats not the case.

common/compositor/factory.cpp
```diff
@@ -17,7 +17,9 @@
 #include "factory.h"
 #include "platformdefines.h"
 
-#ifdef USE_GL
+#ifdef USE_DC
+#include "nativesurface.h"
```
kalyankondapally: CreateMediaRenderer still returns VARenderer. 3D Renderer returns NULL.Did we miss changes in factory class ?

PR #267, opened 2018-03-14
Disable buffer cache

js0701:
Resource manager use buffer age to manage
buffer lifecyle

Jira: None
Tests: UI no flicker

Change-Id: I3bb55c4d25ba7c031c2dc2a2dbfeaa71dcbae9a1

### DIFF
--- a/common/core/overlaylayer.cpp
+++ b/common/core/overlaylayer.cpp
@@ -24,6 +24,7 @@
 #include "hwcutils.h"
 
 #include "resourcemanager.h"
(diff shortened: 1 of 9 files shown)


kalyankondapally:
Do you still want to go ahead with this, seems complicated. The only thing we need to fix is really figure out why sometimes cache returns wrong buffers.

js0701:
@kalyankondapally  We can leave it here. Right

js0701:
Closed as obosoltete

PR #272, opened 2018-03-15
Fix warnings when building hwc

csravelar:
Adjust casting and include gaurds to silence redefined and narrow conversion warnings. 

Jira: None.
Test: Build passes without warnings

### DIFF
--- a/common/compositor/gl/glrenderer.cpp
+++ b/common/compositor/gl/glrenderer.cpp
@@ -84,9 +84,9 @@ bool GLRenderer::Init() {
 }
 
 bool GLRenderer::Draw(const std::vector<RenderState> &render_states,
-                      NativeSurface *surface) {
-  GLuint frame_width = surface->GetLayer()->GetDisplayFrameWidth();
-  GLuint frame_height = surface->GetLayer()->GetDisplayFrameHeight();
+		      NativeSurface *surface) {
+  GLuint frame_width = surface->GetWidth();
+  GLuint frame_height = surface->GetHeight();
   GLuint left = 0;
   GLuint top = 0;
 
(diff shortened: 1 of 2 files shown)


kalyankondapally:
@csravelar Can you apply your changes after the merge commit, so the diff is clear ?

tests/third_party/json-c/math_compat.h
```diff
@@ -19,12 +19,16 @@
 
 #ifndef HAVE_DECL_INFINITY
 #include <float.h>
+#ifndef INFINITY
```
strassek: We should try to avoid carrying local changes for json-c, if this is really needed then lets get it merged upstream first.

common/core/overlaylayer.cpp
```diff
@@ -61,18 +61,22 @@ int32_t OverlayLayer::ReleaseAcquireFence() const {
...
+}
+
 void OverlayLayer::SetBuffer(HWCNativeHandle handle, int32_t acquire_fence,
                              ResourceManager* resource_manager,
                              bool register_buffer, HwcLayer* layer) {
   std::shared_ptr<OverlayBuffer> buffer(NULL);
 
-  // FIXME: Enable cache again.
-  /*if (resource_manager && register_buffer) {
+  if (resource_manager && register_buffer) {
```
strassek: Wait, this looks like more than a warning fix. If you are enabling some feature then that should be made clear in the commit message.

csravelar: That's not part of the patch?
https://github.com/intel/IA-Hardware-Composer/pull/272/files
I think you're looking at a pre git pull rebase patch

kalyankondapally: hmm for some reson pre git pull rebase patch is here....

PR #273, opened 2018-03-17
Use buffer handle to identify buffer

js0701:
Instead of using fd since fds are dynamical
Reenable cache again

Change-Id: Ic3bcc53ef813ec9eb9056ce316b2e1a51e76b952
Jira:None
Tests: No UI flicker on OneA, and video is good

### DIFF
--- a/os/android/platformdefines.h
+++ b/os/android/platformdefines.h
@@ -33,6 +33,7 @@
 #include "platformcommondefines.h"
 #include <cros_gralloc_handle.h>
 
+
 #define DRV_I915 1
 #include <i915_private_android_types.h>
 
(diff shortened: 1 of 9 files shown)


wsi/drm/drmbuffer.cpp
```diff
@@ -38,8 +38,6 @@ DrmBuffer::~DrmBuffer() {
     resource_manager_->MarkResourceForDeletion(image_, image_.texture_ > 0);
   } else {
     if (image_.texture_ > 0) {
-      image_.handle_ = 0;
```
kalyankondapally: why is this removed ?

os/android/platformdefines.h
```diff
@@ -54,44 +55,24 @@ struct gralloc_handle {
...
 #define ITRACE(fmt, ...) ALOGI(fmt, ##__VA_ARGS__)
 #define WTRACE(fmt, ...) ALOGW("%s: " fmt, __func__, ##__VA_ARGS__)
 #define ETRACE(fmt, ...) ALOGE("%s: " fmt, __func__, ##__VA_ARGS__)
 #define STRACE() ATRACE_CALL()
+
+inline uint32_t GetNativeBuffer(uint32_t gpu_fd, HWCNativeHandle handle)
+{
+  uint32_t id = 0;
+  uint32_t prime_fd = handle->handle_->data[0];  
+  if (drmPrimeFDToHandle(gpu_fd, prime_fd, &id)) {
```
kalyankondapally: hmm will this not change for every call ? do we hit cache ever ? also, don't we need to close the fds.

kalyankondapally: hmm overall seems to work fine. This is my only concern, dont we need to close this id here.

wsi/drm/drmbuffer.cpp
kalyankondapally: hmm these will be deleted twice, here and  when u delete the media handle. Any reason why u removed these lines ?

os/android/platformdefines.h
js0701: id is the same with imported handle. id is the only reliable identification for buffer. When you release graphics buffer, already told kernel the id is not used anymore

js0701: @kalyankondapally  I removed the video handle=0 because it was a WA before. I removed those 2 lines and I don't see any warnings and errors. In theory what I am doing here is right

kalyankondapally:
fixed few issues and merged thanks

PR #276, opened 2018-03-19
videoeffect: Query vpp caps before use it

xuguangxin:
two things fixed here:
1. We query caps on UpdateCaps, but we use the caps in SetVAProcFilterDeinterlaceMode.
But SetVAProcFilterDeinterlaceMode call before UpdateCaps, this makes
SetVAProcFilterDeinterlaceMode use a invalid value.
2. when we changed the video effect, we mmeset the param_, it will make
the param_.surface_region and param_.output_region useless

Change-Id: I75a1f30b06fb76c7af37310c947dd9b99837bc90
Jira: None.
Test: 1. Video playback works fine.
2. adb shell hwcservice_test -b 100, you see the video turn bright on screen

### DIFF
--- a/common/compositor/va/varenderer.cpp
+++ b/common/compositor/va/varenderer.cpp
@@ -335,6 +335,36 @@ bool VARenderer::DestroyMediaResources(
   return true;
 }
 
+bool VARenderer::LoadCaps()
+{
+    VAProcFilterCapColorBalance colorbalancecaps[VAProcColorBalanceCount];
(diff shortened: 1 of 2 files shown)


kalyankondapally:
Thanks @xuguangxin  Had to fix idnentation, https://github.com/intel/IA-Hardware-Composer/wiki/Contributions#coding-style

xuguangxin:
thanks help to indent, I will do it next for next commit

Issue #277, opened 2018-03-19
coverity: Dead code in Render State.


kalyankondapally:
Reported Issue:

equality_cond: Jumping to case hwcomposer::kTransform90.
 69      case HWCTransform::kTransform90: {
 70        swap_xy = true;
 71        if (transform & HWCTransform::kReflectX) {
   CID 1465620 (#1 of 2): Logically dead code (DEADCODE) [select issue]
 72          flip_xy[0] = true;
 73          flip_xy[1] = true;
   const: At condition transform & hwcomposer::kReflectY, the value of transform must be equal to 4.
   dead_error_condition: The condition transform & hwcomposer::kReflectY cannot be true.
 74        } else if (transform & HWCTransform::kReflectY) {
   CID 1465620 (#2 of 2): Logically dead code (DEADCODE)dead_error_line: Execution cannot reach this statement: flip_xy[0] = false;.
 75          flip_xy[0] = false;
 76          flip_xy[1] = false;
 77        } else {
 78          flip_xy[0] = false;
 79          flip_xy[1] = true;

strassek:
I'll start on this now

strassek:
The fact that the code in question is dead shows us that there isn't sufficient test coverage. (kTransform90 | kReflectX) and (kTransform90 | kReflectY) ought to be valid transform modes. I will investigate that problem further.

PR #279, opened 2018-03-19
Remove dead code in RenderState

strassek:
This code has never been reachable since its introduction in the initial
code import. Simply drop it for now, making room for someone to correctly
add whatever transform case it was originally designed to cover.

Jira: None
Test: Build and run testlayers

Fixes-coverity-id: 1465620

### DIFF
--- a/common/compositor/renderstate.cpp
+++ b/common/compositor/renderstate.cpp
@@ -68,16 +68,8 @@ void RenderState::ConstructState(std::vector<OverlayLayer> &layers,
       }
       case HWCTransform::kTransform90: {
         swap_xy = true;
-        if (transform & HWCTransform::kReflectX) {
-          flip_xy[0] = true;
-          flip_xy[1] = true;
-        } else if (transform & HWCTransform::kReflectY) {
-          flip_xy[0] = false;
-          flip_xy[1] = false;
-        } else {
-          flip_xy[0] = false;
-          flip_xy[1] = true;
-        }
+        flip_xy[0] = false;
+        flip_xy[1] = true;
         break;
       }
       default: {


strassek:
This patch fixes issue #277

common/compositor/renderstate.cpp
```diff
@@ -68,16 +68,8 @@ void RenderState::ConstructState(std::vector<OverlayLayer> &layers,
       }
       case HWCTransform::kTransform90: {
```
kalyankondapally: I think right way would to move this kTransform90 case and handle it separately here. if (transform & kTransform90) {
/// blah blah
} else {
existing switch statement without kTransform90 case ?

strassek: Maybe, but that would go a step beyond dead code removal. We've never handled the case where both kTransform90 and kReflectX have both been set, and so I wouldn't expect there to be any meaningful test coverage. I would suggest we clean this up now, and then add a separate issue for supporting that functionality.

kalyankondapally: You can check here: https://github.com/intel/IA-Hardware-Composer/blob/master/os/android/iahwc2.cpp#L965 And now if I remember in some rotation modes, we had some layers rendered in-correctly which would be explained by this. My suggestion is fix already now as we have the plumbing in rest of the stack.

strassek: > we had some layers rendered in-correctly which would be explained by this

Was that buggy behavior ever reported?

> My suggestion is fix already now as we have the plumbing in rest of the stack.

I'm not saying we shouldn't try to fix the rotation issues now, only that the scope of this PR and the issue that it fixes is the removal of unreachable code. No more. No less.

I'm happy to investigate, identify any relevant test cases, and then open a separate issue for the rotation mode behavior before we close #277.

kalyankondapally: hmm No, removing the code is not really the scope of #277. If that was the fix than yes, but it's clear that it's not the fix(i.e. If we didnt take those into account anywhere else, ya what u did here made sense). Yes bugs where reported in our internal database.

kalyankondapally: Dont confuse with what coverty is reporting, scope of #277 was to find the right fix, not just follow what the tool recommended.

strassek: I don't think coverity recommended any particular course of action, it just states that this bit of code is dead, and that is all that was included in the text of #277. The right fix for dead code is to remove it, as it only serves to confuse people who are trying to read it. I would even go as far as saying that dead code is worse for readability than outdated comments.

The discovery (or rediscovery) of an underlying bug is a fortunate but secondary side effect of the process of going through a code cleanup. That is really the essence of why I'm pushing back on this, I want it to be acceptable for people to submit code cleanups.

I'm not trying to diminish our responsibility to report bugs and in this case due diligence requires that we open an issue, detailing the transform cases that obviously aren't being covered.

As I already stated, I'm fine with keeping #277 open until we fix all the issues that fall out of this exercise, but I really hope we don't set a precedent that code cleanups are unwelcome.

kalyankondapally: I dont think we are setting any precedence for code clean ups, Code clean ups doesn't mean removing code :)  This is a bug for me where we didn't notice that these are bit flags. (Copy paste typos) If someone is doing code clean up, I would expect him to fix this as part of code clean up. This is by no means dead code....  

I would expect to file a feature in case transforms are not supported, they are today.

kalyankondapally: So in other words what I am saying is, yes we did have buggy code and now let's not continue doing it but fix it with have proper logic to handle these cases. As far as your test cases, thats a genuine concern but we don't have tests for these kind of features today and that's something we should start building.

strassek: > If someone is doing code clean up, I would expect him to fix this as part of code clean up.

Then that is precisely what we disagree on, and I won't push the topic any further. Just a difference in expectations.

> This is by no means dead code....

All I mean by dead code is that it is unreachable, and can be safely removed without any affect to the final program. Whatever context contributes to it being there in the first place doesn't make it any less dead.

kalyankondapally: Fair enough :)

PR #282, opened 2018-03-20
kRrefreshRate should return refresh period

js0701:
in nanoseconds

Change-Id: I8a0d1ad641f921cc334716a8e0a8649c17bf5dd3
Jira:OAM-53716
Tests:None

### DIFF
--- a/common/core/nesteddisplay.cpp
+++ b/common/core/nesteddisplay.cpp
@@ -393,7 +393,7 @@ bool NestedDisplay::GetDisplayAttribute(uint32_t /*config*/,
       break;
     case HWCDisplayAttribute::kRefreshRate:
       // in nanoseconds
-      *value = 60;
+      *value = 16666666;
       break;
     case HWCDisplayAttribute::kDpiX:
       // Dots per 1000 inches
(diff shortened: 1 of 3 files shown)


common/core/resourcemanager.cpp
```diff
@@ -90,18 +90,22 @@ void ResourceManager::RegisterBuffer(const uint32_t& native_buffer,
 
 void ResourceManager::MarkResourceForDeletion(const ResourceHandle& handle,
                                               bool has_valid_gpu_resources) {
+  lock_.lock();
```
kalyankondapally: hmm why is this case ?

js0701: purged_resources_ has to be protected as it is also used in PreparePurgedResources. There could be situation where PreparePurgedResoruces are not in the same thead

js0701: @kalyankondapally  I did not meet any bug. But I think in theory it require lock here

kalyankondapally: ya to avoid these locks we actually separate destroy_media_resources_ and gpu resources which are accessed from multiple thread. I dont see the need for this.

common/core/nesteddisplay.cpp
```diff
@@ -393,7 +393,7 @@ bool NestedDisplay::GetDisplayAttribute(uint32_t /*config*/,
       break;
     case HWCDisplayAttribute::kRefreshRate:
       // in nanoseconds
-      *value = 60;
+      *value = 16666666;
       break;
```
kalyankondapally: @txu4  Any comments ?

txu4: This patch is right, AOSP framework will convert the nanosecond to real refresh rate.

PR #285, opened 2018-03-21
Refine VA hsbc module.

feiwan1:
1. Ensure HSBC filter can be created from 1st frame by using
default color value. Otherwise there will be green screen.
3. Do memset before set parameters into libva.

Jira: None.

### DIFF
--- a/common/compositor/va/varenderer.cpp
+++ b/common/compositor/va/varenderer.cpp
@@ -302,6 +302,7 @@ bool VARenderer::Draw(const MediaState& state, NativeSurface* surface) {
     ETRACE("Failed to update capabailities. \n");
     return false;
   }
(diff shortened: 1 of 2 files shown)


kalyankondapally:
For some reason full screen video gets hung with this change on Android IA. I am going to do bit more testing, if you want to check in mean time though.

kalyankondapally:
seems like a false alarm, pls ignore

feiwan1:
Add another path for CSC scaling mode setting in varender.

feiwan1:
Synced with Guanxin, if create HSBC with all 4 default value maybe will have an influence on performance. So I will repush a patch to use only 1 default to create HSBC filter if the filter.size is 0.

PR #288, opened 2018-03-21
Refine VA hsbc module.

feiwan1:
1. Ensure HSBC filter can be created from 1st frame. Otherwise there
will be green screen.
3. Do memset before set parameters into libva.
4. Ensure scaling mode can be set to libva correctly.

Jira: None.

### DIFF
--- a/common/compositor/va/varenderer.cpp
+++ b/common/compositor/va/varenderer.cpp
@@ -127,7 +127,9 @@ bool VARenderer::SetVAProcFilterColorValue(HWCColorControl mode,
     } else if (prop.value_ != colorbalance_caps_[mode].value_) {
       if (prop.value_ > colorbalance_caps_[mode].caps_.range.max_value ||
           prop.value_ < colorbalance_caps_[mode].caps_.range.min_value) {
-        ETRACE("VA Filter value out of range\n");
+        ETRACE("VA Filter value out of range. Mode %d range shoud be %f~%f\n",
+               mode, colorbalance_caps_[mode].caps_.range.min_value,
+               colorbalance_caps_[mode].caps_.range.max_value);
         return false;
       }
       colorbalance_caps_[mode].value_ = prop.value_;
@@ -144,7 +146,9 @@ bool VARenderer::SetVAProcFilterColorValue(HWCColorControl mode,
     } else if (prop.value_ != sharp_caps_.value_) {
       if (prop.value_ > sharp_caps_.caps_.range.max_value ||
           prop.value_ < sharp_caps_.caps_.range.min_value) {
-        ETRACE("VA Filter sharp value out of range\n");
+        ETRACE("VA Filter sharp value out of range. should be %f~%f\n",
+               sharp_caps_.caps_.range.min_value,
+               sharp_caps_.caps_.range.max_value);
         return false;
       }
       sharp_caps_.value_ = prop.value_;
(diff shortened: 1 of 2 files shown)


common/compositor/va/varenderer.cpp
```diff
@@ -440,18 +448,33 @@ bool VARenderer::UpdateCaps() {
...
+      cbparam[index].type = VAProcFilterColorBalance;
+      cbparam[index].value = itr->second.value_;
+      cbparam[index].attrib = itr->second.caps_.type;
+      index++;
     }
   }
 
+  // workarround if filter num is 0, then set HSBC/Deinterlace will meet green
+  // screen.
+  if (!index) {
```
xuguangxin: Just for reminder, this may have performance issue, since it always set the hue. @feiwan1 . please reproduce the issue using libyami, we report to driver team and fix the green screen issue in driver.

kalyankondapally: Can we just disable setting HSBC in this case and fix the driver issue. Dont want to take the W/A here

xuguangxin: Currently, we find if we do not set the filter, and then after some frame later, we set the filter, it will turn green. 
Yeah, I do not like this too, but it takes time to fix this in the driver. 
Is it possible to merge this first, to unblock this milestone.
then we will fix this in the driver ASAP

kalyankondapally: Let's fix this in driver @xuguangxin, otherwise these things get lost easily. Milestone pressure gets things fixed quickly :)

stripes416: @feiwan1 can you help me understand the work around needed here? I'm having a hard time understanding how this prevents the issue. Thanks!

common/compositor/va/varenderer.cpp
```diff
@@ -440,18 +448,33 @@ bool VARenderer::UpdateCaps() {
...
+    cbparam[index].type = VAProcFilterColorBalance;
+    cbparam[index].value = colorbalance_caps_[HWCColorControl::kColorHue]
+                               .caps_.range.default_value;
+    cbparam[index].attrib =
+        colorbalance_caps_[HWCColorControl::kColorHue].caps_.type;
+    index++;
+  }
+
+  if (index) {
```
csravelar: @feiwan1 since the **if** condition on line 460 will always cause index to be incremented on line 466 (if index is ever 0), would the other **if** condition on line 469 be necessary? Currently, it is possible for both **if** conditions to run `if index is 0`. Otherwise, if the intended effect is to have only one or the other we could change the **if** on line 469 to **else if** instead of removing it.

common/compositor/va/varenderer.cpp
```diff
@@ -440,18 +448,33 @@ bool VARenderer::UpdateCaps() {
...
-      if (!cb_elements[static_cast<int>(itr->first)].CreateBuffer(
-              va_context_, VAProcFilterParameterBufferType,
-              sizeof(VAProcFilterParameterBufferColorBalance), 1, &cbparam)) {
-        return false;
-      }
-      filters_.push_back(cb_elements[static_cast<int>(itr->first)].buffer());
+      cbparam[index].type = VAProcFilterColorBalance;
+      cbparam[index].value = itr->second.value_;
+      cbparam[index].attrib = itr->second.caps_.type;
+      index++;
```
csravelar: Do we ever expect the line 440 **for** loop to not run at least once? Otherwise, won't index always be at least 1 and the **if** on line 460 always be ignored?

common/compositor/va/varenderer.cpp
feiwan1: When set param_.filter_flags as none-zero, the issue will disappear. But not sure why this happens, no va error returns, no hwc error.

PR #290, opened 2018-03-21
Fix conversion warning when building in androidia

csravelar:
Adjust casting to remove warning that occurs when building make
androidia with hwc.

Jira: None.
Test: Build passes without the warning.


wsi/pixelbuffer.cpp
```diff
@@ -55,7 +55,7 @@ void PixelBuffer::Initialize(const NativeBufferHandler *buffer_handler,
     return;
   }
 
-  for (i = 0; i < height; i++)
```
strassek: Can we also remove the declaration of 'i'?

kalyankondapally:
merged

PR #292, opened 2018-03-22
Pre-built shader support

downor:
These two changes are for making compositor be able to use prebuilt-shader program to save time to build it from sources.

There are two steps to generate prebuilt shaders, First one is to create shader source file that can be parsed by shader-db offline compiler tool using common/compositor/gl/gl_shader_pre_built/generate_shader_test.sh. 

Second step is to use shader-db tool to compile "hwc_shader_prog_<n_layers>.shader_test" generated from first step.

The pre-built shader needs to be placed in the same folder where the application is running.

Jira: None.
Test: Verify testlayer works with and without pre-built shaders

Signed-off-by: Dongwon Kim <dongwon.kim@intel.com>

### DIFF
--- a/common/compositor/gl/glprogram.cpp
+++ b/common/compositor/gl/glprogram.cpp
@@ -152,7 +152,7 @@ static GLint GenerateProgram(unsigned num_textures,
   if (shader_prog_fp) {
     /* check the size of file */
     fseek(shader_prog_fp, 0, SEEK_END);
-    int binary_sz = ftell(shader_prog_fp);
+    long binary_sz = ftell(shader_prog_fp);
     rewind(shader_prog_fp);
 
     char *binary_prog = (char *)malloc(binary_sz);


common/compositor/gl/glprogram.cpp
```diff
@@ -128,6 +128,60 @@ static std::string GenerateFragmentShader(int layer_count) {
...
+
+  // try pre-built shader program first
+  std::ostringstream shader_program_fname;
+  shader_program_fname << "hwc_shader_prog_"
+                       << num_textures
+                       <<".shader_test.bin";
+
+  FILE *shader_prog_fp;
+  static PFNGLPROGRAMBINARYOESPROC program_binary_oes =
+     (PFNGLPROGRAMBINARYOESPROC)eglGetProcAddress("glProgramBinaryOES");
```
strassek: For consistency, can you place this with the other get_proc calls in common/compositor/gl/shim.cpp

common/compositor/gl/glprogram.cpp
```diff
@@ -128,6 +128,60 @@ static std::string GenerateFragmentShader(int layer_count) {
...
+      *shader_log << "Failed to create program."
+                  << "\n";
+    return 0;
+  }
+
+  // try pre-built shader program first
+  std::ostringstream shader_program_fname;
+  shader_program_fname << "hwc_shader_prog_"
+                       << num_textures
+                       <<".shader_test.bin";
```
strassek: Don't we need a full path here? Where would these files be stored on Android?

common/compositor/gl/glprogram.cpp
```diff
@@ -128,6 +128,60 @@ static std::string GenerateFragmentShader(int layer_count) {
...
+        if (shader_log)
+          *shader_log << "Pre-built shader program binary has been loaded "
+                      << "Successfully\n";
+
+        return program;
+      }
+    }
+
+    fclose(shader_prog_fp);
+    free(binary_prog);
```
strassek: This looks like a possible double free. Maybe you meant these two calls to be in an else statement?

csravelar:
We should probably follow:
https://github.com/android-ia/manifest/wiki/Contributions

common/compositor/gl/glprogram.cpp
```diff
@@ -128,6 +128,60 @@ static std::string GenerateFragmentShader(int layer_count) {
...
+
+  shader_prog_fp = fopen(shader_program_fname.str().c_str(), "rb");
+
+  if (shader_prog_fp) {
+    /* check the size of file */
+    fseek(shader_prog_fp, 0, SEEK_END);
+    int binary_sz = ftell(shader_prog_fp);
+    rewind(shader_prog_fp);
+
+    char *binary_prog = (char *)malloc(binary_sz);
```
strassek: This seems like a bad idea without doing some bounds checking first.

common/compositor/gl/glprogram.cpp
downor: yes, you are right. I will fix this.

common/compositor/gl/glprogram.cpp
downor: ok

common/compositor/gl/glprogram.cpp
downor: Assumption was this shader program is placed in the same location as executable. I am also wondering if there's any dedicated location of those binaries loaded by compositor. I will check this and update the code.

strassek: Do we *need* to read in the shader program from disk at runtime? It seems far safer to bake the binary data into the HWC binary.

common/compositor/gl/glprogram.cpp
downor: yeah, will do check on binary_prog before moving further

common/compositor/gl/glprogram.cpp
strassek: To make that last point more clear, you can take a look at how we are including SPIR-V (binary) shaders in the vulkan backend. https://github.com/intel/IA-Hardware-Composer/blob/4004258e14bae04ec0f6897be4f0889b4b52e0d7/common/compositor/vk/vkprogram.cpp#L22-L28

downor: yeah, I got your point.

downor: BTW, how is this binary stream generated? Also isn't this architecture independent? Does it generate  by make script?

strassek: `glslangValidator` is used to compile the glsl down to SPIR-V format. It isn't generated during the build, we have the full binary checked into git. Now that you mention it, we really ought to have some script to automate the process for when we want to modify the shaders.

You're right, SPIR-V is architecture independent. If we don't have that luxury with pre-built GL shaders, then we need to be careful about which binaries we are generating / installing / loading.

common/compositor/gl/glprogram.cpp
```diff
@@ -128,6 +128,60 @@ static std::string GenerateFragmentShader(int layer_count) {
...
+
+    if (fread(binary_prog, 1, binary_sz, shader_prog_fp) == binary_sz) {
+      fclose(shader_prog_fp);
+      program_binary_oes(program, GL_PROGRAM_BINARY_FORMAT_MESA, binary_prog,
```
kalyankondapally: can you please query if GL_MESA_program_binary_formats extension is supported first and also same with shader binary.

common/compositor/gl/glprogram.cpp
downor: GL program binary is a platform dependent so we just can't have one fixed binary data as done for SPIR-V. I think this is another reason why we need to automate the process.

downor: I checked the fragment shader under vk/. It seems that LayerCount is fixed to 1. Does this mean only the case with 1 layer will be supported when Vulkan is used?

strassek: SPIR-V gives you a chance to modify constants just before the program gets compiled down for the GPU. That mechanism is called specializations.
https://github.com/intel/IA-Hardware-Composer/blob/037f54610351266d0f59487821a52afc213785c8/common/compositor/vk/vkprogram.cpp#L167-L171

downor: I see... In case of GL (i965 DRI), the program binary is already in GPU code (with NIR attached to it in case NIR->Gen code conversion is required), so changing constants won't be allowed there. Because of this, I initially thought we should create multiple binary files each of which is from shaders with different # of layers then let HWC code to select the right shader program. We can of course get all of these binary data in the source code but the data structure might be huge if we have too many cases with different # of layers.

strassek: Understood. So you are concerned that the sheer number of shader binaries would cause libhwcomposer to grow too large? I think the max number of layers is currently limited by number of textures the hardware supports, but I don't think we would ever hit that number on Android. Would you mind taking a look and reporting back how much memory one of these shaders actually takes? Maybe 16 shaders or so wouldn't be such a huge impact.

downor: Total is 579841 ~= 580KB
Yes, it is not that huge. So I assume I can probably create data structure in the source code with these binary programs.

-rw-rw-r-- 1 dongwonk dongwonk 39061 Mar 23 17:44 hwc_shader_prog_10.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 42457 Mar 23 17:44 hwc_shader_prog_11.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 45853 Mar 23 17:44 hwc_shader_prog_12.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 49233 Mar 23 17:44 hwc_shader_prog_13.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 52565 Mar 23 17:44 hwc_shader_prog_14.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 55961 Mar 23 17:44 hwc_shader_prog_15.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 59357 Mar 23 17:44 hwc_shader_prog_16.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk  8409 Mar 23 17:44 hwc_shader_prog_1.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 12069 Mar 23 17:44 hwc_shader_prog_2.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 15465 Mar 23 17:44 hwc_shader_prog_3.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 18789 Mar 23 17:44 hwc_shader_prog_4.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 22249 Mar 23 17:44 hwc_shader_prog_5.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 25573 Mar 23 17:44 hwc_shader_prog_6.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 28969 Mar 23 17:44 hwc_shader_prog_7.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 32357 Mar 23 17:44 hwc_shader_prog_8.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 35737 Mar 23 17:44 hwc_shader_prog_9.shader_test.bin
-rw-rw-r-- 1 dongwonk dongwonk 35737 Mar 23 17:44 wc_shader_prog_12.shader_test.bin

kalyankondapally: We probably can max at 16 layers, if we do get layers beyond that fallback to compile the shader run time ?

kalyankondapally: Also we should have README doc, if someone wants to pre-compile shaders supporting more layers if needed in their own environment.

downor: yes, current code already does this. If it can't find the shader program file including targeted # of layers in its name, it falls into run-time build process. By the way, I thought about getting binary data carried as a C data structure in the code but it is kinda messy unless we automate this binary generation because it's platform and i965 dependent. I am wondering if it would be acceptable to get those loaded from files now and take care of code-embedded binary later after more discussions on build automation.

kalyankondapally: if the concern is this adds Mesa as a strong dependency here, we can take this path:
1)We can have x86 file, which has all these hard-coded c array.  Have it under build flags and enabled by default. Anyone can switch this off if not needed.
2)If implementation  is missing or 1 is disabled explicitly we can fallback loading from file system. On android the path should be /vendor/etc/ You can define it here:
https://github.com/intel/IA-Hardware-Composer/blob/master/os/android/platformdefines.h
need to come up something for Clear Linux/Yocto which can go in here: https://github.com/intel/IA-Hardware-Composer/blob/master/os/linux/platformdefines.h

Exporting the files to /vendor/etc during build time should be something we should be able to handle and @downor you can ignore it.
3)In case any of these binary shaders fail to compile or we cannot find it on system, program should fall back to run time compilation like we do now.

If you want to implement 2 first and than 1, I am k with that. But dont see much work for postponing 1 as I see you need to build all needed infrastructure for 2 anyway

downor: I can work on 1) however, do you want this arrays created automatically during make or you want to pre-generate these manually with fixed target and the driver version (i965_dri.so)? Having all possible arrays for different combinations is unrealistic since the shader program binary comes with sha-1 key in its header that contains i965's build id that is re-generated every time i965 is rebuilt (even if the source code is same)

kalyankondapally: hmm k Let's go with 2 for now than while we figure out how to handle 1 @strassek Any concerns ?

strassek: >  the shader program binary comes with sha-1 key in its header that contains i965's build id that is re-generated every time i965 is rebuilt (even if the source code is same)

So mesa rejects anything that doesn't match? This seems like a sensible restriction on Mesa's part, but I don't think we can lean on this behavior. At least not unless that restriction is coming from spec. And if it isn't, we are required to supply and load only the binaries that match the current hardware and driver build. In other words, we can't just keep pushing various binaries to Mesa until we get a match (not that anyone was suggesting otherwise).

On the HWC side how can we check if the binary we want to load is suitable for the underlying driver and hardware? Can we even ask the driver to give us its build id? @downor does shader-db give you the option of building the shader binary for anything other than the current environment?

We don't have a good reason to even attempt to shipping binaries for anything but the exact environment of the target machine, and that just isn't known until installation time.

> do you want this arrays created automatically during make or you want to pre-generate these manually with fixed target and the driver version (i965_dri.so)?

I think the reality that we are arriving at is that it just isn't tractable to pre-generate, and you can't expect products to do anything manually.

Can we instead explore the possibility of allowing the driver to generate the binaries on first run? I know that may be unacceptable for some products, but it seems like a much saner default case.

downor: So mesa rejects anything that doesn't match? This seems like a sensible restriction on Mesa's part, but I don't think we can lean on this behavior.

DW] This is because glProgramBinary we use to load the pre-built binary is paired with glGetProgramBinary that we also use to generate shader program binary. Both API should be from the same driver (i965) to guarantee the pre-built binary works. This is done by checking build-id embedded.

@downor does shader-db give you the option of building the shader binary for anything other than the current environment?

DW] Yes, shader-db does the cross compilation for any hardware and driver (as far as it's intel). We specify the target driver by setting "LIBGL_DRIVERS_PATH" and target architecture by passing in target GEN's PCI_ID.

Can we instead explore the possibility of allowing the driver to generate the binaries on first run? I know that may be unacceptable for some products, but it seems like a much saner default case.

DW] Unfortunately the whole point of using pre-built shaders is to make things boot up faster by letting the system bypass shader build process. BTW, "shader-cache" mechanism is already up and running in i965. This allows the system to cache shader program/object whenever it's built for the same purpose but this doesn't cover very first running.

DW] Anyway, I already started to work on a script and simple tool that generates .h with binary arrays that can be embedded in composer library. I hope we can use this for the automation. I will share these in couple of days.

kalyankondapally: Also, if program fails to link the pre-compiled shaders we should blacklist and compile it during run time. hmm Instead of PCI_ID, can we just pass in the Gen (i.e. 7, 7.5) ?

strassek: > We specify the target driver by setting "LIBGL_DRIVERS_PATH" and target architecture by passing in target GEN's PCI_ID.

Got it, so it sounds like we need to require that the target PCI_ID gets set at configure time, otherwise don't try enabling the feature.

tpalli:
We have now EGL_ANDROID_blob_cache support, is there reason why hwcomposer cannot use that? I've noticed Surfaceflinger does not use it, why is this? As example systemui utilizes it. If it would be used, the shaders would get generated on first boot and then sequential boots would utilize prebuilt shaders.

common/compositor/gl/glprogram.cpp
downor: @strassek : Yeah, it should be pre-determined and yes, if not, we need to disable it.
@kalyankondapally : Each gen architecture has several variants and each variant in a group has its own PCI ID. So only way to differentiate it completely is using PCI_ID for now.

common/compositor/gl/glprogram.cpp
downor: yeah, I will add code for query.

kalyankondapally:
@tpalli  I remember with SF, we need to enable build flag. I am waiting for all latest code to settle down a bit before turning it on.

tpalli:
The effort required here looks quite a bit just for first boot, I wast thinking maybe enabling blob cache would be more seamless approach. With Weston you will have the same functionality as what EGL_ANDROID_blob_cache provides via regular disk cache, it is enabled now (from Mesa 18.0) for Linux desktop builds.

downor:
@tpalli Would shader program stay in the database even after power-cycle if we use EGL_ANDROID_blob_cache? As far as I know, disk-shader-cache helps when the same application is re-executed over and over again without rebooting the system. Is EGL_ANDROID_blob_cache different?

tpalli:
@downor Yes, disk cache does not get destroyed during reboot, on regular Linux desktop you can find it from $HOME/.cache/mesa_shader_cache/ and on Android it is under /data/user_de/0, for example for Settings app it is /data/user_de/0/com.android.settings/code_cache/com.android.opengl.shaders_cache.

kalyankondapally:
@tpalli Missed u comments hmm

eero-t:
For things that are built as part of the OS image, shader cache content could be pre-generated at the build time.  That way it would match Mesa version on the OS image and cover things needed at first boot of the device.

eero-t:
Yes, PCI ID would obviously need to be overriden (like e.g. Mesa shader-db does).   Instead of the real target applications, which may be hard to run on build environment, there could be a separate shader compilation program that just compiles the relevant shaders with suitable GL state setup, to seed the shader cache.

Note that some GL state differences would result in different shader compilation results, that's why state needs to match fairly well, otherwise this program would do redundant compilations for things that wouldn't match the run-time shader identification.

Maybe one possibility for that kind of a program could be enhancing Mesa's shader-db shader source dumping facility:
https://www.mesa3d.org/shading.html#capture

to contain all necessary state information (the items that cause shader compilation differences), and then use shader-db program to do the cache preceeding during build?

As last step, one would then need to include the generated shader cache content to the final OS image.

downor:
@eero-t 
Have you checked a new pull request:
https://github.com/intel/IA-Hardware-Composer/pull/310

I added automation there in the new pull request. With PCI-ID fed from configure script, it runs shader-db to create shader programs for the target then integrate it to hw composer (after doing binary to arrays). Compositor code tries to load this pre-built shaders from arrays on run-time. If it fails, it checks if there are any shader binary files under dedicated location in the file system. If this fails as well, it goes through run-time compilation path. 

I think this is pretty much identical with what you described except that it use get_program_binary instead of directly using "shader-cache". 

To cover difference in GL state, output shader program binary contains not only GEN program byt also NIR form of shader program. So I hope most of GL state difference can be handled by regenerating GEN program from NIR if there are any.

eero-t:
Are there any additional security concerns that need to be taken into account?

Mesa's shader cache is shared by all processes running under same user, but I don't think all such processes  (e.g. browser) should be able to substitute other processes cache items -> either the cache storage needs to be read-only, or these locations should be isolated. If OS already doesn't do that...

tpalli:
If I've understood correctly this cache implemented here is only for the HWC so there should be no security concerns, nobody else uses this cache.

What comes to disk cache, on Android the disk cache is separated across processes, every process gets it's own directory of items to mess with.

kalyankondapally:
https://github.com/intel/IA-Hardware-Composer/pull/310 merged as part of this PR

PR #293, opened 2018-03-23
Fixed one display hang when plug 3 displays in clone mode.

shuangwan01:
JIRA: #1242
Test: Boot into Android home screen.

### DIFF
--- a/wsi/physicaldisplay.cpp
+++ b/wsi/physicaldisplay.cpp
@@ -128,6 +128,20 @@ void PhysicalDisplay::Connect() {
   SPIN_LOCK(modeset_lock_);
 
   connection_state_ &= ~kDisconnectionInProgress;
+
+  SPIN_UNLOCK(modeset_lock_);
+
+  if (source_display_) {
+    // Current display is a cloned display, set the source_display_'s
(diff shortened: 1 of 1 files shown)


kalyankondapally:
merged thanks

kalyankondapally:
causes regression with virtual environment. Can you work with yugang to check why.

yugang:
i tried again, manually add the patch to latest master, and can't reproduce the reg again, not sure why, anyway i will work with shuang to confirm this soon later

shuangwan01:
I will have a try to see if we could reproduce the issue on master in virtual environment and share the update latterly.

shuangwan01:
Just tried the patch on top of HWC master and using master branch of virtual environment. The hang issue is not reproduced. 
Here are some details of testing conducted:
1. HWC master + master branch of virtual environment: reproduced 3 times of third display hang over 15 times of boot up testing and no system boot up issue found;
2. HWC master **+ this patch** + master branch of virtual environment: didn't reproduce third display hang over 15 times of boot up testing and system boot up issue is not found as well.

Based on the test result, the risk of  bring regression is low per evaluation.

@kalyankondapally Could you help to merge the patch?

kalyankondapally:
merged

PR #298, opened 2018-03-26
Use global cache for KMS resources

js0701:
KMS Reousrces include framebuffers and gem handles
should be process widely shared
Do not close gem handle when other
drmbuffer are still refernce it

Change-Id: I880e88194333d36d3ae3df79c017b500de41449c
Jira: None
Tests: Extend mode, video playback

### DIFF
--- a/common/core/framebuffermanager.h
+++ b/common/core/framebuffermanager.h
@@ -1,0 +1,122 @@
+/*
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+*/
+
+/*
+Design of ResourceManager:
+The purpose is to add cache magagement to external buffer owned by hwcLayer
+to avoid import buffer and glimage/texture generation overhead
+
+1: the ResourceManager is owned per display, as each display has a
+separate
+GL context
+2: ResourceManager stores a refernce of external buffers in a vector
+   cached_buffers, each vector member is hash map.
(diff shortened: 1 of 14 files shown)


tpalli:
Something does not feel right about this. Cross-process sharing of buffers is done using dmabuf file descriptors, as long as these descriptors are open, the framebuffer behind should not get released.

tpalli:
Ah now I think I got it, this is adding reference counting for BO's created directly by hwcomposer itself. Those should be refcounted if create/destroy in man places but gem handles is then a different thing and those get refcounted via file descriptor mechanism.

common/core/framebuffermanager.h
```diff
@@ -0,0 +1,132 @@
...
+};
+
+struct FBEqual {
+  bool operator()(const FBKey &p1, const FBKey &p2) const {
+    bool equal = (p1.gpu_fd == p2.gpu_fd) && (p1.width == p2.width) &&
+                 (p1.height == p2.height) &&
+                 (p1.frame_buffer_format == p2.frame_buffer_format);
+    if (equal) {
+      for (int i = 0; i < 4; i++) {
+        equal = equal && (p1.gem_handles[i] == p2.gem_handles[i]);
```
tpalli: I think the handle check is enough here, handles are unique to the process.

js0701: Right. Put here for safety.... 
Thanks for this and other comments.
This patch is here for catching the beta release.
As you can see, We may need further rework. I will follow up

kalyankondapally: is this needed still ? I thought this didn't turn out to be issue right

kalyankondapally: Also, since this is such a big change, I would rather do it properly now. We can take an exception for the flicker which seem to happen randomly because of the resource deletion I guess.   I like to the idea as long as we can keep the buffer looks up as fast as possible and not incur a linear exponential cost.

js0701: @kalyankondapally  You mean we can just use handle check here right

js0701:
code updated to remove redudant equal check

kalyankondapally:
This is a bad W/A which is removed. Let's track why damage rect is not correctly updated instead of bringing this back

kalyankondapally:
Took the damage rect W/A for now while we figure out the final solution

kalyankondapally:
thx cleaned up and merged. I think we can improve like instead of having a static instance, gpudevice should allocate it  and pass it around to the core code.

PR #299, opened 2018-03-26
Fixes deadlock

munish-b:
Jira: None
Tests: hwc-val tests pass

With this display is seen. Frame rate is not correct after this.

### DIFF
--- a/tests/hwc-val/tests/hwc/framework/common/HwcvalDebug.cpp
+++ b/tests/hwc-val/tests/hwc/framework/common/HwcvalDebug.cpp
@@ -48,27 +48,8 @@ int Mutex::lock() {
   timespec timeStart;
   clock_gettime(CLOCK_REALTIME, &timeStart);
   timespec timeNow, timeEla;
-  int timecount;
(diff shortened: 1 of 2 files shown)


tests/hwc-val/tests/hwc/framework/common/HwcvalDebug.cpp
```diff
@@ -105,7 +86,7 @@ int Mutex::unlock() {
 
 bool Mutex::tryLock() {
   ALOGD_IF(MUTEX_CONDITION_DEBUG, "Testing mutex %p thread %u", this, gettid());
-  mspinlock.lock();
+  //mspinlock.lock();
```
kalyankondapally: we should lock and immediately unlock ? I dont see where we have the counterpart unlock here

munish-b: Adding lock is the correct.  it is unlocked by Mutex::unlock
Thanks

munish-b:
I will fix fps, squash and update this patch.Thanks

PR #300, opened 2018-03-26
Fix warnings generated in common core hwc

csravelar:
Adjust casting for coversion warnings and adjust ifdef gaurd which
causes a param unused warning for source_layers.

Jira: None.
Test: Build passes without warnings


common/core/nesteddisplay.cpp
```diff
@@ -298,7 +300,6 @@ bool NestedDisplay::Present(std::vector<HwcLayer *> &source_layers,
     } while (rc != msg_size && rc >= 0);
   }
   memset(buf, 0, METADATA_BUFFER_SIZE);
```
kalyankondapally: I dont think METADATA_BUFFER_SIZE is defined ?

kalyankondapally:
needs rebasing

PR #301, opened 2018-03-27
Refine VA module.

feiwan1:
1. Ensure HSBC/Sharp/Deinterlace/Scaling mode can be set to libva
correctly.
3. Do memset before quary from libva.
4. Set default csc scaling mode as High Quality mode.

Jira: None.

### DIFF
--- a/os/android/iahwc2.cpp
+++ b/os/android/iahwc2.cpp
@@ -140,7 +140,7 @@ HWC2::Error IAHWC2::Init() {
   else
     ALOGI("EXPLICIT SYNC support is enabled");
 
-  property_get("board.hwc.scaling.mode", value, "0");
+  property_get("board.hwc.scaling.mode", value, "2");
   scaling_mode_ = atoi(value);
   switch (scaling_mode_) {
     case 1:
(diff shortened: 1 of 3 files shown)


os/android/iahwc2.cpp
```diff
@@ -143,14 +143,14 @@ HWC2::Error IAHWC2::Init() {
...
   scaling_mode_ = atoi(value);
   switch (scaling_mode_) {
+    case 0:
+      ALOGI("HWC Scaling Mode High Quality");
+      break;
     case 1:
       ALOGI("HWC Scaling Mode Fast");
       break;
-    case 2:
-      ALOGI("HWC Scaling Mode High Quality");
```
xuguangxin: please keep original order, this will match the order with libva, else it may introduce confuse.

xuguangxin:
other things are ok to me. thanks @feiwan1

os/android/iahwc2.cpp
feiwan1: If this, it will become:
==============
case 0:
case 2:
    High quality mode
    break;
case 1:
    Fast mode
    break;
default:
    unsupport mode
    break; 
=============
That means 0&2 all will get into high quality mode. That's also a little confused:)
BTW, if don't set board.hwc.scaling.mode in android environment, property_get will return 0.

xuguangxin: case 2:
High quality mode
break;
case 1:
Fast mode
break;
default:
unsupport mode
break;

kalyankondapally:
merged

PR #308, opened 2018-03-29
Should also release handle if native buffer

js0701:
Otherwise it will cause memory leak

Tests: app switch in Android IVI no memory grow
Jira: None


os/android/gralloc1bufferhandler.cpp
```diff
@@ -172,7 +172,7 @@ bool Gralloc1BufferHandler::CreateBuffer(uint32_t w, uint32_t h, int format,
 
 bool Gralloc1BufferHandler::CanReleaseGemHandles(HWCNativeHandle handle) const {
   if (handle->hwc_buffer_) {
-    return false;
+    return true;
```
kalyankondapally: for hwc_buffer_ minigbm takes care of releasing the handle right.

Issue #302, opened 2018-03-27
Add support for custom position for float display


gowthamhosamane:
By default HWC sets the display hardware config mode as the resolution and position it to 0,0 for top left. The additional mechanism of setting the top left position to non-zero in case of float display enabled.

PR #309, opened 2018-03-29
Custom position for float display

gowthamhosamane:
By default HWC sets the display hardware config mode as the resolution
and position it to 0,0 for top left. The additional mechanism of
setting the top left position to non-zero in case of float display enabled.
The float display resolution could be smaller than display resolution.

Jira: #302

Test: When float mode is set, config changes to set custom resolution
at custom position for the required pipe to be working as desired.
If not enabled, it should continue to work as before.


gowthamhosamane:
Modified test app and linux frontend.

kalyankondapally:
lgtm thanks

PR #310, opened 2018-03-29
Pre-built shader support (NEW)

downor:
This pull request replaces the previous pull request (#292).

Now there are three changes under this pull request, which basically make build process now takes care of pre-building of shaders automatically if provided this new configuration option:

--prebuilt-shader-pci-id=<PCI ID in hex number form (e.g 0x1234) >

AND

shader-db library and executable (intel_stub.so, run) exist under common/compositor/gl/gl_shader_pre_built/shader-db/

Once binary data (shader program binary) is generated, it will be added to the HWC source as header files and compiled together with it.

glprogram.cpp tries to load this static data first. If it does not exist or fail to load it, it tries to get load this binary data from files at /vendor/etc if available. Then it goes to run-time compilation path if none of those prebuilt shaders are available or functional.

Jira: None.
Test: 
A. built in binary test

0. Place <shader-db> project under ./common/compositor/gl/gl_shader_pre_built/

1. Build HWC:
./autogen.sh --prefix=$WLD --prebuilt-shader-pci-id=<current GPU's PCI-ID>
make && make install

2. Verify testlayer works with built-in pre-built shaders

B. shader program files test

1. Build HWC:
  ./autogen.sh --prefix=$WLD
  make && make install

2. run ./common/compositor/gl/gl_shader_pre_built/generate_shader_test.sh
    use shader-db to build shader program binaries from generated shader-test files 
    move *.shader_test.bin over to under /vendor/etc
   
3. Verify testlayer works with built-in pre-built shaders

Signed-off-by: Dongwon Kim dongwon.kim@intel.com


common/compositor/gl/glprogram.h
```diff
@@ -21,6 +21,8 @@
 
 #include "shim.h"
 
+#define PREBUILT_SHADER_FILE_PATH "/vendor/etc/"
```
harishkrupo: Instead of hardcoding this here, can we pass this from the build system?
For example in Android.mk -DPREBUILT_SHADER_FILE_PATH="/vendor/etc/"
On linux -DPREBUILT_SHADER_FILE_PATH="${prefix}/etc"

common/compositor/gl/gl_shader_pre_built/generate_shader_test.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
+
+if [ $# -eq 0 ]
+  then
+    echo "No arguments supplied"
+    echo "Using layer_cnt = 1"
+    layer_cnt=1
+else
+    layer_cnt=$1
```
harishkrupo: should be $i?

common/compositor/gl/glprogram.h
downor: ok I will move it to Makefile.

common/compositor/gl/glprogram.cpp
```diff
@@ -130,24 +130,23 @@ static GLint LoadPreBuiltBinary(GLint gl_program,
                                 void *binary, long size) {
   GLint status;
 
-  std::string extensions((const char *)glGetString(GL_EXTENSIONS));
-  std::string mesa_prog_bin_fmts("GL_MESA_program_binary_formats");
-  std::string::size_type pos;
-
```
strassek: Squash this hunk into the first commit, so you aren't adding code and then immediately removing it.

common/compositor/gl/glprogram.cpp
```diff
@@ -130,24 +130,23 @@ static GLint LoadPreBuiltBinary(GLint gl_program,
...
+   * TODO: will enable this once this is fixed in MESA
+   *
+   * std::string extensions((const char *)glGetString(GL_EXTENSIONS));
+   * std::string mesa_prog_bin_fmts("GL_MESA_program_binary_formats");
+   * std::string::size_type pos;
+   *
+   * pos = extensions.find(mesa_prog_bin_fmts);
+   *
+   * if (pos == std::string::npos)
+   *   return 0;
```
strassek: Please don't commit commented out code, a short description will suffice.

common/compositor/gl/glprogram.cpp
```diff
@@ -126,8 +126,93 @@ static std::string GenerateFragmentShader(int layer_count) {
...
+
+  if (shader_prog_fp) {
+    /* check the size of file */
+    fseek(shader_prog_fp, 0, SEEK_END);
+    binary_sz = ftell(shader_prog_fp);
+    rewind(shader_prog_fp);
+
+    char *binary_prog = (char *)malloc(binary_sz);
+    if (binary_prog) {
+      if (fread(binary_prog, 1, binary_sz, shader_prog_fp) == binary_sz) {
```
strassek: Given that we generate shaders on-demand, I don't think it is such a great idea to block on disk access. @kalyankondapally have you considered generating all shaders in bulk during initialization?

common/compositor/gl/glprogram.cpp
```diff
@@ -126,8 +126,93 @@ static std::string GenerateFragmentShader(int layer_count) {
...
+
+  shader_prog_fp = fopen(shader_program_fname.str().c_str(), "rb");
+
+  if (shader_prog_fp) {
+    /* check the size of file */
+    fseek(shader_prog_fp, 0, SEEK_END);
+    binary_sz = ftell(shader_prog_fp);
+    rewind(shader_prog_fp);
+
+    char *binary_prog = (char *)malloc(binary_sz);
```
strassek: I'd still like to see some bounds checking first. Come up with some maximum acceptable binary size and don't accept files that are bigger.

common/compositor/gl/gl_shader_pre_built/bin_to_c_array.c
```diff
@@ -0,0 +1,85 @@
...
+  long binary_sz = ftell(binary_fp);
+  rewind(binary_fp);
+
+  fprintf(header_fp, LICENSE_TERM"\n");
+
+  fprintf(stdout, "file size: %ld\n", binary_sz);
+
+  /* write binary size to the beginning of the file */
+  for (i = 0; i < sizeof(binary_sz); i++) {
+    fprintf(header_fp, "0x%02x, ", (unsigned char)(binary_sz & 0xFF));
```
MicheleL-Intel: Consider using fwrite instead because it's a bit faster than fprintf.  You can optimize further by hardcoding certain parameters such as 'size' to 16, and adjust your shifts accordingly.

common/compositor/gl/gl_shader_pre_built/bin_to_c_array.c
```diff
@@ -0,0 +1,85 @@
...
+  if (!binary_fp) {
+    fprintf(stderr, "input file does not exist\n");
+    print_usage();
+    exit(-1);
+  }
+
+  /* getting the size */
+  fseek(binary_fp, 0, SEEK_END);
+  long binary_sz = ftell(binary_fp);
+  rewind(binary_fp);
```
MicheleL-Intel: Check binary_sz is non-zero here.

common/compositor/gl/gl_shader_pre_built/bin_to_c_array.c
```diff
@@ -0,0 +1,85 @@
...
+  fprintf(stdout, "input file:%s\n", argv[1]);
+  fprintf(stdout, "output file:%s\n", argv[2]);
+
+  binary_fp = fopen(argv[1], "rb");
+  header_fp = fopen(argv[2], "wb");
+
+  if (!binary_fp) {
+    fprintf(stderr, "input file does not exist\n");
+    print_usage();
+    exit(-1);
```
MicheleL-Intel: fclose any opened handles before exiting.

common/compositor/gl/gl_shader_pre_built/bin_to_c_array.c
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /* print out information */
+  fprintf(stdout, "input file:%s\n", argv[1]);
+  fprintf(stdout, "output file:%s\n", argv[2]);
+
+  binary_fp = fopen(argv[1], "rb");
+  header_fp = fopen(argv[2], "wb");
+
+  if (!binary_fp) {
```
MicheleL-Intel: Also check 'header_fp' for invalid file descriptor.

common/compositor/gl/glprogram.cpp
```diff
@@ -126,8 +126,116 @@ static std::string GenerateFragmentShader(int layer_count) {
   return fragment_shader_stream.str();
 }
```
MicheleL-Intel: Does entire LoadPrebuiltBinary need to be encapsulated with "ENABLE_PREBUILT_SHADER_BIN_ARRAY "?

common/compositor/gl/glprogram.cpp
strassek: Kalyan and I just talked about this off-line, in the long run we are going to want to do bulk shader generation as close to program init as possible. For now you can keep this here and leave the feature disabled by default.

kalyankondapally:
@strassek  @tpalli Anymore concerns before starting to merge this ?

common/compositor/gl/glprogram.cpp
```diff
@@ -126,8 +126,83 @@ static std::string GenerateFragmentShader(int layer_count) {
...
+            *shader_log << "Pre-built shader program binary has been loaded "
+                        << "Successfully (from files)\n";
+          return program;
+        }
+      } else {
+          free(binary_prog);
+          fclose(shader_prog_fp);
+      }
+    } else {
+        fclose(shader_prog_fp);
```
strassek: formatting here looks off. When in doubt, use clang-format.

common/compositor/gl/glprogram.cpp
downor: @strassek : I see your request is still outstanding. Can you approve this if you are ok with this part now?

common/compositor/gl/glprogram.cpp
downor: hmm I thought I replied back to this comment but it seems it was lost..
so we can't do it because LoadPrebuiltBinary is needed for second option as well where it loads shader programs from files.

common/compositor/gl/gl_shader_pre_built/bin_to_c_array.c
downor: Please review this new file. It is now closing all file ptrs.

common/compositor/gl/gl_shader_pre_built/bin_to_c_array.c
downor: (I think I lost my comment because of my mistake..)
I don't know what you meant by using fwrite instead. What it tries to do is to write binaries in ASCII format. I think fprintf is most general and standard way to do it. If you think I am missing something here, please kindly educate me. Thanks

common/compositor/gl/glprogram.cpp
downor: We could do that. Is it ok to set it to 10MB for now?

common/compositor/gl/glprogram.cpp
downor: @strassek : yeah thanks for the tip. I see the format is corrected via clang-format. Would this be enough or should I run clang-format with a specific option?

strassek: The .clang-format file we have in the tree *should* take care of all the options. I'd always recommend checking the diff after running the script, it can make mistakes.

common/compositor/gl/glprogram.cpp
strassek: That sounds reasonable to me

common/compositor/gl/glprogram.cpp
strassek: I don't think it is a good idea to try to do this read by default, so I'd like to put this behind a build flag and have it disabled by default. Once we start doing bulk shader generation at init, then we can get rid of the build flag.

downor: I see. I will define a temporary build flag.

common/compositor/gl/gl_shader_pre_built/bin_to_c_array.c
MicheleL-Intel: Reviewed and file handles are properly managed now.

common/compositor/gl/gl_shader_pre_built/bin_to_c_array.c
```diff
@@ -0,0 +1,99 @@
...
+
+  /* print out information */
+  fprintf(stdout, "input file:%s\n", argv[1]);
+  fprintf(stdout, "output file:%s\n", argv[2]);
+
+  binary_fp = fopen(argv[1], "rb");
+
+  if (!binary_fp) {
+    fprintf(stderr, "input file does not exist\n");
+    print_usage();
```
MicheleL-Intel: Is print_usage() necessary here, since the error has nothing to do with command format?

common/compositor/gl/gl_shader_pre_built/bin_to_c_array.c
MicheleL-Intel: I see what you're doing here.  If you're copying file size as ASCII then fprintf is appropriate.  Have you confirmed binary_sz byte order is correct when read?

downor: yes, I verified it (it was actually the first and most important thing I tested after building this ;-))

common/compositor/gl/gl_shader_pre_built/bin_to_c_array.c
downor: Doesn't seem to be needed. I will remove it.

downor:
@kalyankondapally @strassek @MicheleL-Intel 
I think I took care of all recommended code changes. Please review the latest one and let me know if you have any comments.

MicheleL-Intel:
LGTM.

MicheleL-Intel:
LGTM.

kalyankondapally:
@strassek Any feedback or good to get merged ?

kalyankondapally:
merged, after fixing commit message (Added test, jira) and build issue on Android.

PR #314, opened 2018-03-30
hwc-val: code cleanup

strassek:
This series doesn't make any changes that will alter anyone's build, just cleaning out some things that obviously aren't needed.


strassek:
FYI @ashaarunkumar @munish-b

PR #316, opened 2018-03-31
Weston

kalyankondapally:


os/linux/weston/plugin/compositor-iahwc.c
```diff
@@ -1495,6 +1495,7 @@ static int vsync_callback(iahwc_callback_data_t data, iahwc_display_t display,
 
   if (output->pageflip_timer && output->frame_commited) {
     wl_event_source_timer_update(output->pageflip_timer, 1);
+    // FIXME: This fails to launch weston.
```
harishkrupo: I forgot to remove this code. This should never be there in the first place. Placing it here would race with the main thread and will set a wrong state in the state machine.

.travis.yml
```diff
@@ -154,18 +154,12 @@ script:
...
 #Build weston and weston plugin
 - git clone https://github.com/intel/external-weston.git /tmp/weston
-- pushd .
 - cd /tmp/weston
-- ./autogen.sh --prefix=$WLD --disable-setuid-install
-- make -j9 install
-- popd
+- git checkout origin/iahwc-plugin
+- cd -
+- os/linux/weston/build_script.sh --weston-dir=/tmp/weston --iahwc-dir=$(pwd) --build
```
harishkrupo: hmm... I missed the checkout step.
Also --weston-dir must be ../weston because weston assumes that the iahwc-plugin is placed in ../iahwc/.../compositor-iahwc.c
Will make that change and test

PR #317, opened 2018-04-02
Should always reset plane targes when new DiplayPlaneState

js0701:
Otherwise it will flicker due to not enough target surfaces

Jira: https://jira01.devtools.intel.com/browse/OAM-55851
Tests: Playback video 10 times thers is no flicker at beginnng

### DIFF
--- a/common/display/displayplanemanager.cpp
+++ b/common/display/displayplanemanager.cpp
@@ -215,6 +215,9 @@ bool DisplayPlaneManager::ValidateLayers(
 
             ResetPlaneTarget(last_plane, commit_planes.back());
           }
+          else if (last_plane.NeedsSurfaceAllocation()) {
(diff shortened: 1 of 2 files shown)


common/display/displayplanemanager.cpp
```diff
@@ -212,10 +212,10 @@ bool DisplayPlaneManager::ValidateLayers(
           if (fall_back) {
             if (!validate_final_layers)
               validate_final_layers = !(last_plane.GetOffScreenTarget());
-
-            ResetPlaneTarget(last_plane, commit_planes.back());
           }
 
+          ResetPlaneTarget(last_plane, commit_planes.back());
```
kalyankondapally: hmm this forces offscreen composition and we should do it only when layer cannot be scanned out (i.e. if fallback is true)

PR #319, opened 2018-04-03
Keep the code of layers force GPU to fix black screen when do UI switch.

yugang:
The 3rd display may show black screen when remove the layers force GPU
for removed/added layers, so keep those code to fix the black screen
issue.

Jira: None.
Test: Switch UI to remove/add layers to check if there are black screen.

### DIFF
--- a/common/core/overlaylayer.cpp
+++ b/common/core/overlaylayer.cpp
@@ -89,6 +89,8 @@ void OverlayLayer::SetBuffer(HWCNativeHandle handle, int32_t acquire_fence,
     if (resource_manager && register_buffer) {
       resource_manager->RegisterBuffer(id, buffer);
     }
+  } else {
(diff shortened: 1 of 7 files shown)


common/display/displayplanemanager.cpp
```diff
@@ -235,12 +236,22 @@ bool DisplayPlaneManager::ValidateLayers(
...
                                    last_plane.GetOffScreenTarget()->GetLayer(),
                                    commit_planes)) {
+                  force_all_layers = false;
                   validate_final_layers = true;
                 }
               }
             }
-
-            break;
+            if (force_all_layers) {
```
kalyankondapally: Probably you can revert all changes related to displayplanemanager from the bad commit ?

kalyankondapally:
can u test with latest master and see.

PR #320, opened 2018-04-03
Fix incomplete Display after swipe up the screen to main

RakeshReddyGunda:
menu in portrait Mode.

Jira: 59439.
Test: Build passes on Android.

Signed-off-by: reddyr3x <rakeshx.reddy@intel.com>

### DIFF
--- a/common/core/overlaylayer.cpp
+++ b/common/core/overlaylayer.cpp
@@ -232,14 +232,6 @@ void OverlayLayer::InitializeState(HwcLayer* layer,
   }
 
   surface_damage_ = layer->GetLayerDamage();
-  // In case of layer using blening we need to force partial clear. Otherwise
-  // we see content not getting updated correctly. For example:
-  // on Android enable, settings put system user_rotation 1 and
-  // navigate to settings on Android.
-  if (((blending_ != HWCBlending::kBlendingNone) && !surface_damage_.empty())) {
-    state_ |= kForcePartialClear;
-    surface_damage_ = layer->GetDisplayFrame();
-  }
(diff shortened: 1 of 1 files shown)


common/core/overlaylayer.cpp
```diff
@@ -263,6 +255,15 @@ void OverlayLayer::InitializeState(HwcLayer* layer,
     }
   }
 
+ // In case of layer using blening we need to force partial clear. Otherwise
```
kalyankondapally: Why does moving the code chunk here fix the issues ?

stripes416: Also, while you didn't write the comment, just moved it, the misspelling should be fixed.
/blening/blending/

RakeshReddyGunda: swap should be  done on layer-> GetLayerDamage as we are rotating display to 90.
But in this case the swap was done on layer->GetDisplayFrame  which caused incomplete screen.

kalyankondapally: hmm but u still use DisplayFrame right ?

RakeshReddyGunda: Yes we use DisplayFrame after swapping the surface damage co-ordinates.

md-sami: After the rotation is done, the source layer dimensions are already updated, so why do we need to do swap again?

md-sami: After the rotation is done, the source layer dimensions are already updated, so why do we need to do swap again?

PR #321, opened 2018-04-04
Add support for the canvas color feature

vivekkreddy:


wsi/drm/drmdisplay.cpp
```diff
@@ -571,6 +573,48 @@ void DrmDisplay::ApplyPendingLUT(struct drm_color_lut *lut) const {
   drmModeDestroyPropertyBlob(gpu_fd_, lut_blob_id);
 }
 
+uint64_t DrmDisplay::DrmRGBA(uint32_t bpc, uint16_t red, uint16_t green,
```
stripes416: What does the value bpc represent here?

public/nativedisplay.h
```diff
@@ -244,6 +244,12 @@ class NativeDisplay {
     return false;
   }
 
+  /**
+  * API for setting canvas color for display.
+  */
```
kalyankondapally: we usually add the name of the variable and comment it out.

Can you explain the expectations of the inputs to this function

wsi/drm/drmdisplay.cpp
vivekkreddy: bpc = bits per color.

wsi/drm/drmdisplay.cpp
```diff
@@ -571,6 +573,48 @@ void DrmDisplay::ApplyPendingLUT(struct drm_color_lut *lut) const {
...
+			      uint16_t blue, uint16_t alpha) const {
+  int shift;
+  uint64_t result;
+
+  if (bpc > 16)
```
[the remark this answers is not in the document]
vivekkreddy: bpc cannot just be a multiple of 8; for instance, it can be 10. In other words, there are color formats that expect to have 10 bits per color component.

public/nativedisplay.h
vivekkreddy: Ok, will include this feedback in v2.

kalyankondapally: hmm we should just take in r,g,b,a values here. Converting that to something kernel driver understands should be internal logic in hwc side. btwn We could move the drmhelper functions to within hwc.

vivekkreddy: Yeah, now that I think about it, your suggestion seems like a better approach. Will send out a new version...

tests/apps/jsonlayerstest.cpp
```diff
@@ -337,6 +337,26 @@ class HotPlugEventCallback : public hwcomposer::DisplayHotPlugEventCallback {
...
+      return;
+
+    uint64_t uint_max = std::numeric_limits<unsigned int>::max();
+    if (color < 0)
+      color = 0;
+    if (color > uint_max)
+      color = uint_max;
+
+    for (auto &display : connected_displays_)
+      display->SetCanvasColor(bpc, color & 0x000000ff,
```
MicheleL-Intel: Lines 355 - 357 bit manipulation assume an 8-bit bpc and 32-bit color.  So if bpc happens to be 10 bits, this would be an issue because two bits are getting truncated.  I think you have to handle this as a 64-bit depth instead.

wsi/drm/drmdisplay.cpp
```diff
@@ -65,6 +66,7 @@ bool DrmDisplay::InitializeDisplay() {
   GetDrmObjectProperty("GAMMA_LUT", crtc_props, &lut_id_prop_);
   GetDrmObjectPropertyValue("GAMMA_LUT_SIZE", crtc_props, &lut_size_);
   GetDrmObjectProperty("OUT_FENCE_PTR", crtc_props, &out_fence_ptr_prop_);
+  GetDrmObjectProperty("background_color", crtc_props, &canvas_color_prop_);
```
kalyankondapally: to align with rest of properties should this be BACKGROUND_COLOR ?

csravelar: ++

tests/apps/jsonlayerstest.cpp
vivekkreddy: Right, the above logic indeed does not handle the 10 bpc case. However, my goal here with this patch is just to demonstrate how the hardware feature works and not to comprehensively cover all the formats. Including cases like 10 bpc, 12 bpc, 16 bpc and testing them would certainly increase the scope of what I planned on doing and I am not sure if I'd be able to do it any soon.

wsi/drm/drmdisplay.cpp
```diff
@@ -571,6 +573,44 @@ void DrmDisplay::ApplyPendingLUT(struct drm_color_lut *lut) const {
...
+  val |= blue << shift;
+  val <<= 16;
+  val |= alpha << shift;
+
+  return val;
+}
+
+void DrmDisplay::SetPipeCanvasColor(uint16_t bpc, uint16_t red, uint16_t green,
+                                    uint16_t blue, uint16_t alpha) const {
+  if (canvas_color_prop_ == 0)
```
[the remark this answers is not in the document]
vivekkreddy: Yes, InitializeDisplay should assign the property ID to canvas_color_prop_.

wsi/drm/drmdisplay.cpp
vivekkreddy: I did think about the lower case name but I am not sure if we can count on the fact the kernel driver always exposes property names that are upper case. For example, there are properties like audio, type whose names are exposed in lower case. Alternatively, do you think we should make GetDrmObjectProperty ignore the case of the property name to accommodate lower case property names?

tests/apps/jsonlayerstest.cpp
kalyankondapally: In this case can you document the restriction as part of the API in NativeDisplay, so everyone can easily figure this out. If this is something we need to improve in future, you can add a issue in Github issue tracker

wsi/drm/drmdisplay.cpp
kalyankondapally: this was just a quesiton, I dont see any restriction on having lower/higher case https://01.org/linuxgraphics/gfx-docs/drm/drm-kms-properties.html seems mixed here. Not sure how it's decided

kalyankondapally:
merged

PR #326, opened 2018-04-10
Keep the code of layers force GPU to fix black screen when do UI switch.

yugang:
The 3rd display may show black screen when remove the layers force GPU
for removed/added layers, so keep those code to fic the black screen
issue.

Jira: None.
Test: Switch UI with remove/add layers to check if there are black screen.

@kalyankondapally i tried the original PR in latest master and met crash, so i just did the rebase and submit new PR here. pls noticed that we can't revert this file to original commit as the revert around "if (composition.empty()) " will cause the crash of hwc.

### DIFF
--- a/common/display/displayplanemanager.cpp
+++ b/common/display/displayplanemanager.cpp
@@ -221,6 +221,7 @@ bool DisplayPlaneManager::ValidateLayers(
 
           break;
         } else {
+          bool force_all_layers = true;
           if (composition.empty()) {
             composition.emplace_back(plane, layer, this, layer->GetZorder(),
                                      display_transform_);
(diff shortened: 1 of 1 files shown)


common/display/displayplanemanager.cpp
```diff
@@ -247,12 +248,18 @@ bool DisplayPlaneManager::ValidateLayers(
...
             }
-
-            break;
+            if (force_all_layers) {
+              ForceGpuForAllLayers(commit_planes, composition, layers,
+                                   mark_later, false);
+              *re_validation_needed = false;
+              *commit_checked = true;
+              return true;
+            }
```
kalyankondapally: you will need a break; here. If we dont return we should make sure we move on to next plane + layer mapping.

kalyankondapally:
merged

yugang:
done, keep the break

PR #327, opened 2018-04-11
Fix build error for missing header file

mwang30:
Tracked-On:
Signed-off-by: Randy Xu <randy.xu@intel.com>

### DIFF
--- a/Android.common.mk
+++ b/Android.common.mk
@@ -40,6 +40,8 @@ LOCAL_SHARED_LIBRARIES := \
 
 LOCAL_C_INCLUDES := \
 	system/core/include/utils \
+        frameworks/native/libs/nativewindow/include \
+        frameworks/native/libs/nativewindow \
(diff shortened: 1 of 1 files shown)


Android.common.mk
```diff
@@ -40,6 +40,8 @@ LOCAL_SHARED_LIBRARIES := \
 
 LOCAL_C_INCLUDES := \
 	system/core/include/utils \
+        frameworks/native/libs/nativewindow/include \
```
kalyankondapally: @stripes416  pointed out, can u remove "Change-Id: I4d74489991861b8f1c6559c26a66d87548a572a5
Tracked-On:"

kalyankondapally: can u fix u commit message too ?

kalyankondapally: @mwang30  ping

kalyankondapally:
fixed commit message and merged

PR #328, opened 2018-04-11
hwc-val: Enable build for AIA and get tests running

strassek:
This series includes everything I needed to get a hwc-val test running on Android-IA.


strassek:
FYI @munish-b @ashaarunkumar

PR #329, opened 2018-04-12
Alpha layer

harishkrupo:


os/linux/weston/plugin/compositor-iahwc.c
```diff
@@ -445,9 +445,11 @@ static int iahwc_output_repaint(struct weston_output *output_base,
     output->cursor_plane.y = INT32_MIN;
   }
 
+#if 0
   iahwc_output_render(output, damage);
```
kalyankondapally: why do we need this ?

kalyankondapally: A better descriptive name than ? i.e. DISABLE_OVERLAY ?

harishkrupo: Okay. And set that variable in the makefile?

harishkrupo: okay. I got the makefile part also ready. WIll give a PR in weston if okay.

os/linux/weston/plugin/compositor-iahwc.c
```diff
@@ -1217,6 +1216,7 @@ static void iahwc_assign_planes(struct weston_output *output_base,
       struct weston_surface *es = ev->surface;
       es->keep_buffer = false;
 
+#ifdef IAHWC_DISABLE_OVERLAYS
```
kalyankondapally: hmm sorry this should be #ifndef ?

harishkrupo: that code will be compiled only when IAHWC_DISABLE_OVERLAYS is set.

kalyankondapally: but this will disable overlays usage by default. The code needs to be executed for using overlays right

harishkrupo: No, this code is required only when we want to go for single plane rendering from weston.

kalyankondapally:
check launching weston flower or weston egl demo, everything gets screwed up

kalyankondapally:
Actually just with this change, I see everything is rendered normally but gets screwed up when u launch a demo:

kalyan@kalyan-To-be-filled-by-O-E-M:~/external_drive1/weston/iahwc$ git log -p
commit 9c49fae34b911f047aa3d0a1fc94ee3b57f5d0be
Author: Harish Krupo <harish.krupo.kps@intel.com>
Date:   Thu Apr 12 19:56:46 2018 +0530

    Weston: Don't set alpha value for layers
    
    The raw pixel data that we get is has alpha value premultipled
    so set blending premultiplied instead.
    
diff --git a/os/linux/linux_frontend.cpp b/os/linux/linux_frontend.cpp
index 8c26bba..af00293 100644
--- a/os/linux/linux_frontend.cpp
+++ b/os/linux/linux_frontend.cpp
@@ -307,6 +307,7 @@ int IAHWC::IAHWCDisplay::PresentDisplay(int32_t* release_fd) {
   std::vector<hwcomposer::HwcLayer*> layers;
   hwcomposer::HwcLayer* cursor_layer = NULL;
 
+  raw_data_uploader_->Synchronize();
   for (std::pair<const iahwc_layer_t, IAHWCLayer>& l : layers_) {
     IAHWCLayer& temp = l.second;
     if (temp.GetLayer()->IsCursorLayer())
diff --git a/os/linux/pixeluploader.cpp b/os/linux/pixeluploader.cpp
index 4801719..0db624f 100644
--- a/os/linux/pixeluploader.cpp
+++ b/os/linux/pixeluploader.cpp
@@ -119,7 +119,6 @@ void PixelUploader::HandleRawPixelUpdate() {
   }
 
   std::vector<PixelData>().swap(pixel_data_);
-  pixel_data_lock_.unlock();
 
   for (auto& buffer : texture_uploads) {
     if (callback_) {
@@ -156,6 +155,7 @@ void PixelUploader::HandleRawPixelUpdate() {
     if (buffer.layer_callback_) {
       buffer.layer_callback_->UploadDone();
     }
+    pixel_data_lock_.unlock();
   }
 }

wondering if its now related to z order

kalyankondapally:
also having iahwc_layer_.SetBlending(HWCBlending::kBlendingPremult); in the constructor seems to help for sure

harishkrupo:
By screwed up are you referring to the layers not rendering properly? If so, then yes it is the problem with the z-order. I am already working on a patch to fix will. Will post the same here.
the egl app for some reason is unable to import fd. Will look into it after this patch.

kalyankondapally:
merged

PR #330, opened 2018-04-13
Fix incomplete display  after 90/270 degree rotation

md-sami:
After 90/270 degree rotation is done,the source layer dimensions are already updated, so swap is not required
here.

Jira: 59439
Test: Rendering works fine when displays are rotated 90/270

Signed-off-by: samiuddi sami.uddin.mohammad@intel.com

### DIFF
--- a/common/core/overlaylayer.cpp
+++ b/common/core/overlaylayer.cpp
@@ -250,9 +250,14 @@ void OverlayLayer::InitializeState(HwcLayer* layer,
       switch (plane_transform_) {
         case HWCTransform::kTransform270:
         case HWCTransform::kTransform90: {
-          std::swap(surface_damage_.left, surface_damage_.top);
-          std::swap(surface_damage_.right, surface_damage_.bottom);
-          break;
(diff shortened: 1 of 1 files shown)


kalyankondapally:
thanks merged

kalyankondapally:
hmm this broke scrolling when in layer has transform 90 or 270.

      switch (plane_transform_) {
        case HWCTransform::kTransform270:
        case HWCTransform::kTransform90: {
	  surface_damage_ = layer->GetDisplayFrame();
          break;
        }
        default:
          break;
      }

For now wondering we should damage the whole layer in this case or swap rects in case damage  is not equal to display frame ?

RakeshReddyGunda:
Issue is not observed with the above fix .

kalyankondapally:
@md-sami  @RakeshReddyGunda can you update the PR ?

md-sami:
@kalyankondapally ok, updated the PR

kalyankondapally:
thx merged

PR #335, opened 2018-04-16
linux_frontend: Preliminary hotplug support

harishkrupo:
Jira: None
Test: Weston launches properly


strassek:
The patch needs a description. See https://github.com/intel/IA-Hardware-Composer/wiki/Contributions
Also, isn't "Test: Weston launches properly" too limited for this change? Does it not launch properly without this patch?

harishkrupo:
@strassek Is this better?

strassek:
Commit message looks great, thanks!

I'd suggest we wait to land this until there is at least one user, such that we can test that it is correct.

kalyankondapally:
Their is actually a use now @strassek, when u launch weston it should automatially plumb to activate the thread and kill it when disconnected. You shouldn't see more than 1 pixel uploader thread active unlike now where you have three threads with two just sleeping.

PR #336, opened 2018-04-17
Adjust formatting in core files

csravelar:
In order to fix some of the formatting conventions in headers and indention that caused code to appear to be in another scope, I adjusted some of the lines in the core folder.

Jira: None
Test: Builds without error


kalyankondapally:
hmm thanks

kalyankondapally:
u missed Jira and test in actual commit message, added that and merged

PR #337, opened 2018-04-17
Fix bug where assignment has no effect

csravelar:
Currently we pass in addr and use it to check the condition. Once inside we assign a nullptr to addr. However, the assignment of function parameter has no effect outside the function. 

Jira: None
Test: Build passes without error


kalyankondapally:
u missed Jira and test in actual commit message, added that and merged

PR #338, opened 2018-04-18
varenderer: layer transform should also consider flip

js0701:
new camera tests add h-flip

Jira:https://jira01.devtools.intel.com/browse/OAM-61084
Tests: Camera Orientation of CTS verifier

### DIFF
--- a/common/compositor/va/varenderer.cpp
+++ b/common/compositor/va/varenderer.cpp
@@ -292,7 +292,10 @@ bool VARenderer::Draw(const MediaState& state, NativeSurface* surface) {
   param_.filter_flags = filter_flags_;
 // currently rotation is only supported by VA on Android.
 #ifdef DISABLE_CURSOR_PLANE
-  param_.rotation_state = HWCTransformToVA(state.layer_->GetTransform());
+  param_.rotation_state = HWCRotationToVA(state.layer_->GetTransform());
+#if VA_MAJOR_VERSION > 1
+  param_.mirror_state = HWCReflectToVA(state.layer_->GetTransform());
+#endif
 #endif
 
   ScopedVABufferID pipeline_buffer(va_display_);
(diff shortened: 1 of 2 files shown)


common/compositor/va/varenderer.cpp
```diff
@@ -292,7 +292,8 @@ bool VARenderer::Draw(const MediaState& state, NativeSurface* surface) {
   param_.filter_flags = filter_flags_;
 // currently rotation is only supported by VA on Android.
 #ifdef DISABLE_CURSOR_PLANE
```
kalyankondapally: we should probably change this check to be #ifdef VA_MAJOR_VERSION > 1

common/compositor/va/varenderer.cpp
```diff
@@ -487,18 +488,22 @@ bool VARenderer::UpdateCaps() {
...
+  if(transform & kTransform270)
+    return VA_ROTATION_90;
+  if(transform & kTransform180)
+    return VA_ROTATION_180;
+  if(transform & kTransform90)
+    return VA_ROTATION_270;
+  return VA_ROTATION_NONE;
+}
+
+uint32_t VARenderer::HWCReflectToVA(uint32_t transform) {
```
kalyankondapally: This function should be added if #ifdef VA_MAJOR_VERSION > 1 is true, can u build test it in u env and see if this gets included

common/compositor/va/varenderer.cpp
kalyankondapally: @xuguangxin Can u help here, how do we ensure we get the right libva versions which includes this params

common/compositor/va/varenderer.cpp
js0701: OK

common/compositor/va/varenderer.cpp
xuguangxin: @kalyankondapally , why we need to keep libva 1.0 code here.  We will remove libva from AIA and AIA will get libva 2.0 based solution in following weeks.

kalyankondapally: But till than we need to have the code compiling right ?

common/compositor/va/varenderer.cpp
```diff
@@ -487,18 +490,25 @@ bool VARenderer::UpdateCaps() {
...
-            break;
-    }
-    return VA_ROTATION_NONE;
+uint32_t VARenderer::HWCRotationToVA(uint32_t transform) {
+  if(transform & kTransform270)
+    return VA_ROTATION_90;
+  if(transform & kTransform180)
+    return VA_ROTATION_180;
+  if(transform & kTransform90)
+    return VA_ROTATION_270;
```
kalyankondapally: hmm why do  u change this to 270 ? Is this something u forgot ? why change original code ?

kalyankondapally:
merged

PR #340, opened 2018-04-18
Remove various dead code

csravelar:
utils_alios.h had an unnecessary assignment

yallocbufferhandler.h had an unused member

gralloc1bufferhandler.h had an unused member

iahwc2.cpp and .h had a color_mode_ that was not used anywhere

gbmbufferhandler.cpp needs a fixme note since I am not sure why
the mod variable is being used outside of the if statement when it
isn't initialized to begin with.

Jira: None
Test: HWC build passes without errors

### DIFF
--- a/os/android/iahwc2.cpp
+++ b/os/android/iahwc2.cpp
@@ -725,7 +725,9 @@ HWC2::Error IAHWC2::HwcDisplay::SetClientTarget(buffer_handle_t target,
 
 HWC2::Error IAHWC2::HwcDisplay::SetColorMode(int32_t mode) {
   supported(__func__);
-  color_mode_ = mode;
+  // TODO: Use the parameter mode to set the color mode for the display to be
+  // used.
+
   return HWC2::Error::None;
 }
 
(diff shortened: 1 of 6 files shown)


os/android/iahwc2.cpp
```diff
@@ -725,7 +725,6 @@ HWC2::Error IAHWC2::HwcDisplay::SetClientTarget(buffer_handle_t target,
 
 HWC2::Error IAHWC2::HwcDisplay::SetColorMode(int32_t mode) {
   supported(__func__);
-  color_mode_ = mode;
```
kalyankondapally: we should add proper implementation for this function, perhaps add some log here to indicate that ?

stripes416: @kalyankondapally Do you mean just adding a fixme / todo note? The fix should probably go in a different pr.

stripes416:
LGTM.

os/android/iahwc2.cpp
kalyankondapally: @stripes416 Ya, just a Fixme/TODO note.

kalyankondapally:
sorry, @csravelar  Can u rebase this patch.

PR #342, opened 2018-04-19
Weston: Use system clock for timestamps

harishkrupo:
Few applications depend on the timestamp sent in the
frame callback for deciding their animation rate.
This timestamp is in turn calculated from the timestamp sent
to the compositor in the weston_output_finish_frame call. Big
differences between two vblank timestamps causes sudden jerks
in the application's animation. So, instead of the vblank
timestamps use the system clock for the timestamp.

Jira: None
Test: Sudden jumps in the animation of weston-simple-damage
      should not be seen.


MicheleL-Intel:
Sorry didn't mean to close this.  It was the touchpad's fault :)

kalyankondapally:
merged

PR #343, opened 2018-04-23
Weston: Fix window animation

harishkrupo:
Weston animates the opening of new windows in a spring like motion.
This is achieved by setting src_{w,h} of the buffer smaller than
the actual buffer width/height and varying the dest_{w,h}.
Previously we were fixing the source buffer dimentions to the acutal
buffer dimentions. This patch uses the dimentions provided by weston
instead of the actual buffer dimentions.

Jira: None
Test: Windows should open with a spring like animation.


kalyankondapally:
merged

PR #344, opened 2018-04-24
VA: rotation should be reversed when mirror is used

js0701:
It is because VA is using rotation + mirror, while
camera expecting mirror + rotation. Reverse the rotation
will make sure identical result instead of diffrent
processing order.

Jira:  https://jira01.devtools.intel.com/browse/OAM-61084
Tests: CTS Camera orientation verifier

### DIFF
--- a/common/compositor/va/varenderer.cpp
+++ b/common/compositor/va/varenderer.cpp
@@ -290,10 +290,13 @@ bool VARenderer::Draw(const MediaState& state, NativeSurface* surface) {
   }
 
   param_.filter_flags = filter_flags_;
+
+#if VA_MAJOR_VERSION >= 1
 // currently rotation is only supported by VA on Android.
-#if VA_MAJOR_VERSION > 1
-  param_.rotation_state = HWCRotationToVA(state.layer_->GetTransform());
-  param_.mirror_state = HWCReflectToVA(state.layer_->GetTransform());
+  uint32_t rotation = 0, mirror = 0;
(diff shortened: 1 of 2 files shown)


common/compositor/va/varenderer.cpp
```diff
@@ -488,29 +491,38 @@ bool VARenderer::UpdateCaps() {
...
+      case kTransform270:
+        rotation = VA_ROTATION_270;
+      case kTransform180:
+        rotation = VA_ROTATION_180;
+      case kTransform90:
+        rotation = VA_ROTATION_90;
+      default:
+        break;
+    }
+  }
```
kalyankondapally: @xuguangxin  Can you review this please.

common/compositor/va/varenderer.cpp
```diff
@@ -488,29 +491,38 @@ bool VARenderer::UpdateCaps() {
...
+  else {
+    if (transform & kReflectX)
```
xuguangxin: what happend if ((transform & kReflexX) && (transform & kReflexY)) is true?

thanks

js0701: They can both be true, and both mirror can be possilbe, right, good findings

common/compositor/va/varenderer.cpp
js0701: @xuguangxin  code updated to consider both reflect and the same time

xuguangxin: thanks Johnson

kalyankondapally:
merged thanks

PR #345, opened 2018-04-24
Weston: Enable DPMS

harishkrupo:
This patch adds code to enable dpms in the linux frontend and implements
the set_dpms hook in the iahwc backend.

Jira: None
Test: Display turns off after an idle period and is powered
      on when input is received.


os/linux/iahwc.h
```diff
@@ -126,6 +127,13 @@ enum iahwc_layer_transform {
   IAHWC_TRANSFORM_FLIP_V_ROT_90
 };
 
+enum iahwc_power_modes {
+  IAHWC_POWER_MODE_ON,
```
kalyankondapally: Instead of adding new ones, can we re-use these https://github.com/intel/IA-Hardware-Composer/blob/master/public/hwcdefs.h#L98 directly

harishkrupo: okay. Including that file directly into iahwc.h would require __cplusplus ifdef blocks. Should I add them there or any other better option?

harishkrupo: Or shall I use DRM_MODE_DPMS_* as the intermediary and then convert them to k{On,Off,...}?

kalyankondapally: @harishkrupo Adding _cplusplus guards is fine

harishkrupo:
@kalyankondapally is this okay?

kalyankondapally:
@harishkrupo Can u rebase please.

harishkrupo:
@kalyankondapally rebased

PR #348, opened 2018-04-26
travis: Run clang-format

harishkrupo:
Make travis run clang-format on the pull request to check
if the patch needs formatting.

Jira: None
Test: travis build stops if patch requires formatting.

### DIFF
--- a/travisci/check_patch.sh
+++ b/travisci/check_patch.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+difference=$(git show origin/master..@ | clang-format-diff-3.9 -p 1 -style=file);
+[ ! -z "$difference" ] && echo "Code formatting is not according to style guidelines. Read https://github.com/intel/IA-Hardware-Composer/wiki/Contributions#coding_style" && exit 1;
+exit 0;
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,8 @@ install:
 - sudo apt-get update
 - sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
 - sudo apt-get update
-- sudo apt-get install automake autoconf libtool xutils-dev libpciaccess-dev python-mako bison flex libomxil-bellagio-dev libexpat1-dev clang-format-3.5 llvm-dev gcc-4.9 g++-4.9 python3-pip libudev-dev libmtdev-dev mtdev-tools libevdev-dev libx11-xcb-dev libxkbcommon-dev libxrandr-dev x11proto-*-dev libxcb* libxdamage* libx11-xcb-dev libxext-dev libxshmfence-dev libwacom-dev libgtk-3-dev check libpam0g-dev
+- sudo apt-get install automake autoconf libtool xutils-dev libpciaccess-dev python-mako bison flex libomxil-bellagio-dev libexpat1-dev clang-format-3.5 llvm-dev gcc-4.9 g++-4.9 python3-pip libudev-dev libmtdev-dev mtdev-tools libevdev-dev libx11-xcb-dev libxkbcommon-dev libxrandr-dev x11proto-*-dev libxcb* libxdamage* libx11-xcb-dev libxext-dev libxshmfence-dev libwacom-dev libgtk-3-dev check libpam0g-dev clang-format-3.9
+- travisci/check_patch.sh
 - sudo pip3 install pathlib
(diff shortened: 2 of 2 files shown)


kalyankondapally:
hmm wondering how to test this @csravelar FYI

harishkrupo:
I have tested the script manually. Maybe open a PR without formatting? :P

kalyankondapally:
@harishkrupo Merge this in your fork, and try with a PR in u fork and test it out ?

harishkrupo:
Ok, sure.

harishkrupo:
Okay, tested. It works. 
[PR](https://github.com/harishkrupo/IA-Hardware-Composer/pull/2)
[Travis log](https://travis-ci.org/harishkrupo/IA-Hardware-Composer/builds/371400634?utm_source=github_status&utm_medium=notification)

travisci/check_patch.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
```
csravelar: Sweet, this is great!
If we ever decide to move to clang-format 5 we could also add something like:
wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
sudo apt-add-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-5.0 main"
sudo apt-get update
sudo apt-get install -y clang-5.0

update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-3.8 100
update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-5.0 1000
update-alternatives --install /usr/bin/clang++ clang /usr/bin/clang-3.8 100
update-alternatives --install /usr/bin/clang clang /usr/bin/clang-3.8 100
update-alternatives --install /usr/bin/clang clang /usr/bin/clang-5.0 1000
update-alternatives --config clang
update-alternatives --config clang++

so you could run:
difference=$(git show origin/master..@ | clang -p 1 -style=file);

travisci/check_patch.sh
```diff
@@ -0,0 +1,5 @@
...
+difference=$(git show origin/master..@ | clang-format-diff-3.9 -p 1 -style=file);
```
csravelar: Are we enforcing clang format on all files? If not we can do something for example to prune for ones we care about:
$CLANG_FORMAT --version
files_to_lint="$(git diff --name-only --diff-filter=ACMRTUXB $TRAVIS_COMMIT_RANGE | grep '^src/[^.]*[.]\(cpp\|h\)$' || true)"
for f in $files_to_lint; do
    d=$(diff -u "$f" <($CLANG_FORMAT "$f") || true)
    if ! [ -z "$d" ]; then
        echo "!!! $f not compliant to coding style, here is the fix:"
        echo "$d"
        fail=1
    fi
done

harishkrupo: hmm, that is what clang-format-diff does :)

csravelar: Ahh, it looks like its cut off there ha, but I was referring to the right side where it prunes cpp and h files " grep '^src/[^.]*[.]\(cpp\|h\)$'"

harishkrupo: We have mostly cpp and h files in the repo with a few sh/readme files. IIRC clang only formats cpp and h files?

csravelar: Yeah, we don't need it. It's just to avoid it checking Android.mk and Makefiles since it'll try to format those too. But unless someone changes a file other than the cpp, h files I don't see too much of a big issue.

travisci/check_patch.sh
```diff
@@ -0,0 +1,5 @@
...
+[ ! -z "$difference" ] && echo "Read the coding style guidelines https://github.com/intel/IA-Hardware-Composer/wiki/Contributions#coding_style" && exit 1;
```
kalyankondapally: You want to add message like "Indentation is wrong",  so they know thats the issue.

travisci/check_patch.sh
harishkrupo: Looks like clang format diff already ignores makefiles. It uses this regex:
[Clang-format-diff regex](https://github.com/llvm-mirror/clang/blob/master/tools/clang-format/clang-format-diff.py#L47)

harishkrupo:
@kalyankondapally updated with the comment change

kalyankondapally:
merged

PR #350, opened 2018-04-26
Fix Vulkan build and add to CI

strassek:


kalyankondapally:
merged

strassek:
Second patch wasn't merged

strassek:
Ah, so you squashed the 2 commits into 1, fair enough.

PR #349, opened 2018-04-26
travis-ci update: Add cppcheck run on codebase

csravelar:
This patch adds a cppcheck install and execution command to
run cppcheck on each pull request to make sure that there is
no resulting errors. Addresses issue #357 

Jira: None
Test: Travis fails if cppcheck finds errors and passes otherwise.
Signed-off-by: Kelly Ledford kelly.ledford@intel.com
Signed-off-by: Richard Avelar richard.avelar@intel.com

PR #352, opened 2018-04-30
Discuss/Fix issues currently in weston plugin

csravelar:
There are currently two interesting errors in compositor-iahwc.c
I wanted to gain feedback from anyone who might have a better
understanding of this particular area of the code base, why it was
done this way.

Jira: None
Test: None

### DIFF
--- a/os/linux/weston/plugin/compositor-iahwc.c
+++ b/os/linux/weston/plugin/compositor-iahwc.c
@@ -1347,6 +1347,7 @@ static int create_output_for_connector(struct iahwc_backend *b) {
   b->iahwc_get_display_info(b->iahwc_device, 0, configs[0], IAHWC_CONFIG_DPIY,
                             &output->base.mm_height);
 
+  //FIXME Memory leak for iahwc_mode
   for (i = 0; i < num_configs; i++) {
     iahwc_mode = iahwc_output_add_mode(b, output, configs[i]);
     if (!iahwc_mode) {
@@ -1420,6 +1421,8 @@ static void session_notify(struct wl_listener *listener, void *data) {
     weston_log("activating session\n");
     weston_compositor_wake(compositor);
     weston_compositor_damage_all(compositor);
+    //FIXME: Output isn't initialized to anything yet we are trying to use its
+    //spin_lock member
     lock(&output->spin_lock);
 
     wl_list_for_each(output, &compositor->output_list, base.link) {


strassek:
Ping @harishkrupo @kalyankondapally

os/linux/weston/plugin/compositor-iahwc.c
```diff
@@ -1420,6 +1421,8 @@ static void session_notify(struct wl_listener *listener, void *data) {
     weston_log("activating session\n");
     weston_compositor_wake(compositor);
     weston_compositor_damage_all(compositor);
+    //FIXME: Output isn't initialized to anything yet we are trying to use its
+    //spin_lock member
```
harishkrupo: Good catch! the lock and unlock should be inside the for_each loop.
@kalyankondapally Is this correct?

os/linux/weston/plugin/compositor-iahwc.c
```diff
@@ -1347,6 +1347,7 @@ static int create_output_for_connector(struct iahwc_backend *b) {
   b->iahwc_get_display_info(b->iahwc_device, 0, configs[0], IAHWC_CONFIG_DPIY,
                             &output->base.mm_height);
 
+  //FIXME Memory leak for iahwc_mode
```
harishkrupo: @csravelar Could you please explain why this is a mem leak :)? I am unable to find the mem leak.

os/linux/weston/plugin/compositor-iahwc.c
kalyankondapally: hmm I dont understand the issue here,

harishkrupo: We are doing a 
`lock(&output->spin_lock);`
but output is defined only inside 
`wl_list_for_each(output, &compositor->output_list, base.link) {`

kalyankondapally: hmm I see, makes sense to move into the loop ya

os/linux/weston/plugin/compositor-iahwc.c
csravelar: Hey @harishkrupo. so I wanted a second opinion on this because I personally cannot see what the issue is either. However, cppcheck gives an error about the following line claiming there is a mem leak.

os/linux/weston/plugin/compositor-iahwc.c
csravelar: @kalyankondapally An error is raised by cppcheck about this

csravelar:
The reason I wanted to bring these two lines to attention was because cppcheck is raising errors for both these for travis. However, since travis ci only supports up to Ubuntu trusty, its using an older version of cppcheck. When I run my newer version of cppcheck on ubuntu 16.04, I will not receive any errors. So I am not sure if it is just the way that they these lines were written that appears like errors to an older version of cppcheck

harishkrupo:
Yes, that seems to be the reason. The mem leak error is not seen when using cppcheck version 1.66. This version is present in trusty-backports repo.
In Travis, you could add this to apt sources before running apt update :
` echo "deb http://archive.ubuntu.com/ubuntu trusty-backports main restricted universe multiverse" >> /etc/apt/sources.list`
and install using 
`apt-get install cppcheck/trusty-backports`
With this, I see only the output uninitialized issue.

harishkrupo:
The interesting thing is, with the latest cppcheck version 1.83, it doesn't even report the output uninitialized error (which is an issue).

csravelar:
Yeah I thought that was strange. Thanks for the info on the trusty-backport repos!

csravelar:
Closing this issue as it pertained to pr #349 . The changes are made there, thanks everyone.

PR #355, opened 2018-05-04
compositor thread: adding corner case check before wait in Draw.

YuanjunHuang:
Adding the check to avoid compositor thread not responsing in some
corner case and leads to UIWDT that found in android stability test.

Jira: https://jira01.devtools.intel.com/browse/OAM-61985
Test: stability test

### DIFF
--- a/common/compositor/compositorthread.cpp
+++ b/common/compositor/compositorthread.cpp
@@ -103,6 +103,13 @@ bool CompositorThread::Draw(std::vector<DrawState> &states,
   // succeed.
   draw_succeeded_ = true;
   tasks_lock_.unlock();
+
+  // Adding check to end up waiting in this
+  // thread in certain corner case.
+  if (states_.empty() && media_states_.empty()) {
+    return draw_succeeded_;
+  }
+
   Resume();
   Wait();
   return draw_succeeded_;


YuanjunHuang:
@kalyankondapally

common/compositor/compositorthread.cpp
```diff
@@ -104,6 +104,11 @@ bool CompositorThread::Draw(std::vector<DrawState> &states,
   draw_succeeded_ = true;
   tasks_lock_.unlock();
   Resume();
+
+  if (states_.empty() && media_states_.empty()) {
```
harishkrupo: if both the states are empty, will we ever do any drawing?
If not we could use this as a fast path and return true?

kalyankondapally: in this case we shouldn't be doing any drawing

harishkrupo: So we can check this case as soon as we enter the method and return true?

kalyankondapally: so far, their is no data to show this is needed. Holing on to this PR.

kalyankondapally:
seems to cause some rendering issues, need more testing.

PR #356, opened 2018-05-04
Transform surface damage to global co-ordinates

harishkrupo:
Surface damage for a layer is set in the layer local
co-ordintates. The surface damage has to be converted to the
display global co-ordinates before assigning it to the
current_rendering_damage variable.

Jira: None
Test: With the surface damage work around removed the contents of
      weston terminal should be updated properly for each frame.


harishkrupo:
@kalyankondapally

kalyankondapally:
can u test Android also, i.e. launch GFXBench and check that the initial status bar shows the progress correctly.

public/hwcutils.h
```diff
@@ -87,6 +88,34 @@ inline OverlapType AnalyseOverlap(const hwcomposer::HwcRect<int>& rect,
...
+
+  ss << "[";
+  for (auto& rect : region) {
+    ss << separator << StringifyRect(rect);
+    separator = ", ";
+  }
+  ss << "]";
+
+  return ss.str();
+}
```
strassek: Where do these get called from?

harishkrupo: These two are useful only while debugging, for dumping rects like display frame/surface damage... I was using them for convenience while adding logs.
Currently, it is not used anywhere, maybe we could use it when ENABLE_DISPLAY_DUMP is enabled, to print the source crop, surface damage and display frame?

harishkrupo:
@strassek Used the Pretty-print functions to dump source crop, display frame and surface damage rects when ENABLE_DISPLAY_DUMP is defined.

public/hwcutils.h
strassek: Seem reasonable. Just 1 additional nit, if these are just for debug printing I'd prefer to drop the 'inline' and let the compiler handle that for us.

harishkrupo:
@strassek updated.

common/core/hwclayer.cpp
```diff
@@ -220,11 +220,24 @@ void HwcLayer::Validate() {
...
+    // From observation: In Android, when the source crop doesn't
+    // begin from (0, 0) the surface damage is already translated
+    // to global display co-ordinates
     if (!surface_damage_.empty() &&
-        (display_frame_width_ < DAMAGE_THRESHOLD &&
-         display_frame_height_ < DAMAGE_THRESHOLD)) {
-      layer_cache_ |= kForceClear;
-      current_rendering_damage_ = display_frame_;
+        ((source_crop_.left == 0) && (source_crop_.top == 0))) {
+      if (display_frame_width_ < DAMAGE_THRESHOLD &&
```
kalyankondapally: You should be able to remove DAMAGE_THRESHOLD W/A also here.

harishkrupo: Ok, will do it.

PR #358, opened 2018-05-14
Skip HWC rendering if /etc/vendor/hwc.lock is locked by another process exclusively.

shuangwan01:
Skip rendering Android system UI if /vendor/etc/hwc.lock presents
  and exclusively locked by another process. This is to avoid
  screen flash if earlyEVS and HWC both rendering content into DRM.


shuangwan01:
included inrelevant commit wrongly.

PR #359, opened 2018-05-14
Skip HWC rendering if /etc/vendor/hwc.lock is locked by another process exclusively.

shuangwan01:
Skip rendering Android system UI if /vendor/etc/hwc.lock presents
and exclusively locked by another process. This is to avoid
screen flash if earlyEVS and HWC both rendering content into DRM.

JIRA: OAM-60710
Test: Boot into home screen in normal boot flow;
      Show earlyEVS camera content if boot with showing rear view
        camera mode;
      Show Android system UI if exit showing rear view camera mode.

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -61,11 +61,6 @@ bool GpuDevice::Initialize() {
   initialization_state_ &= ~kHWCSettingsDone;
   thread_sync_lock_.unlock();
 
-  if (use_thread && lock_fd_ == -1) {
-    // Exit thread as we don't need worker thread after
-    // Initialization.
-    HWCThread::Exit();
-  }
 
   return true;
 }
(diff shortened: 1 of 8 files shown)


common/core/gpudevice.cpp
```diff
@@ -652,39 +647,44 @@ void GpuDevice::InitializeHotPlugEvents(bool take_lock) {
...
 
-void GpuDevice::HandleWait() {
-  if (lock_fd_ == -1) {
-    HWCThread::HandleWait();
+    if (-1 == lock_fd_) {
+      ETRACE("/vendor/hwc.lock doesn't exist, skip the DRM locking");
+      use_lock_file = false;
+    }
+    else {
+      if (flock(lock_fd_, LOCK_EX|LOCK_NB) != 0) {
```
kalyankondapally: Hmm not sure I understand the difference with this new logic and previous one. Why do we see rendering before but not now ?

shuangwan01: The thoughts of changes are:
1) Utilize the asynchronize HWC thread for hwc.lock grabbing and maintaining HWC rendering disable -> enabled status transition in that thread. The relevant code pieces have been removed from GpuDevice::Initialize method as implementing hwc.lock exclusively lock waiting there will block whole HWC component initialization; 
2) Once HWC thread enabled (didn't call HWCThread::Exit method in GpuDevice::Initialize method), HandleWait will be the first method to be called and then HandleRoutine. It's one time logic on rendering disable -> rendering enable, so the implementation moved to HandleWait method;
3) The first time flock call with nonblocking flag is to make the HWC have a chance to know /vendor/hwc.lock locking status. If it's hold by another process, then HWC know it has to skip the rendering content into DRM. And then issue one additional flock with blocking wait flag until another process release the exclusive lock on hwc.lock.
4) Once HWC got the exclusive lock on hwc.lock, it set HWC status to enable content rendering into DRM.

wsi/physicaldisplay.cpp
```diff
@@ -586,6 +586,18 @@ bool PhysicalDisplay::GetDisplayConfigs(uint32_t *num_configs,
   return true;
 }
 
+void PhysicalDisplay::SetDisplayRenderingState(bool rendering)
```
shuangwan01: Though DisplayManager has an interface IgnoreUpdates to ignore Rendering content into DRM but status set by this interface will be affected by display hotplug events and behavior of this existing interface is not reliable. So we added a new interface to ensure the HWC content rendering skipping works as expected.

kalyankondapally: hmm that interface was added only for this purpose, if hot plug events are impacting that we should fix that rather than adding a new interface.

common/core/gpudevice.cpp
kalyankondapally: hmm k thanks

public/gpudevice.h
```diff
@@ -106,6 +106,8 @@ class GpuDevice : public HWCThread {
   SpinLock thread_stage_lock_;
   SpinLock thread_sync_lock_;
   int lock_fd_ = -1;
+  bool hwc_locked = false;
```
kalyankondapally: this should be hwc_locked_

public/gpudevice.h
```diff
@@ -106,6 +106,8 @@ class GpuDevice : public HWCThread {
...
+  bool use_lock_file = true;
```
kalyankondapally: why do we need this ?

common/core/gpudevice.cpp
```diff
@@ -652,39 +647,44 @@ void GpuDevice::InitializeHotPlugEvents(bool take_lock) {
...
-}
+  if (use_lock_file && -1 == lock_fd_) {
+    lock_fd_ = open("/vendor/hwc.lock", O_RDONLY);
 
-void GpuDevice::HandleWait() {
-  if (lock_fd_ == -1) {
-    HWCThread::HandleWait();
+    if (-1 == lock_fd_) {
+      ETRACE("/vendor/hwc.lock doesn't exist, skip the DRM locking");
+      use_lock_file = false;
```
kalyankondapally: hmm in this case where we fail to lock the file, the thread will always exist even if sleeping. That's why we should still try it in the main thread like in the original code and abort the thread in case we are not able to open this hwc.lock file.

common/core/gpudevice.cpp
kalyankondapally: The only difference I  see mainly is that we first try LOCK_EX|LOCK_NB, if that fails just try LOCK_EX. Probably thats the only thing needed for existing logic and not sure we need the rest here.

shuangwan01:
Based on review result, submitted new patch: https://github.com/intel/IA-Hardware-Composer/pull/362
Close this one.

PR #360, opened 2018-05-15
Fix some Klocwork issues

strassek:


common/compositor/va/varenderer.cpp
```diff
@@ -202,19 +202,15 @@ bool VARenderer::SetVAProcFilterDeinterlaceMode(
   return false;
 }
```
kalyankondapally: @xuguangxin  Can u review the patch please.

kalyankondapally: also @feiwan1 ping for review

kalyankondapally:
merged

PR #361, opened 2018-05-15
Review

kalyankondapally:


kalyankondapally:
merged without Revert "Optimize Cursor rendering using GPU" as this caused regression when moving mouse in mosaic mode.

PR #362, opened 2018-05-16
Skip rendering Android system UI if /vendor/hwc.lock locked by another process exclusively.

shuangwan01:
Skip rendering Android system UI if /vendor/hwc.lock presents and exclusively locked by another process. This is to avoid screen flash if earlyEVS and HWC both rendering content into DRM.

JIRA: OAM-60710
Test: Boot into home screen in normal boot flow;
      Show earlyEVS camera content if boot with showing rear view
        camera mode;
      Show Android system UI if exit showing rear view camera mode.

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -35,6 +35,8 @@ bool GpuDevice::Initialize() {
   if (initialization_state_ & kInitialized)
     return true;
 
+  lock_fd_ = open("/vendor/hwc.lock", O_RDONLY);
+
   bool use_thread = true;
   if (!InitWorker()) {
     ETRACE("Failed to initalize thread for GpuDevice. %s", PRINTERROR());
@@ -61,9 +63,10 @@ bool GpuDevice::Initialize() {
   initialization_state_ &= ~kHWCSettingsDone;
   thread_sync_lock_.unlock();
 
-  if (use_thread && lock_fd_ == -1) {
+  if (-1 == lock_fd_) {
     // Exit thread as we don't need worker thread after
     // Initialization.
+    ETRACE("Failed to open /vendor/hwc.lock file.");
     HWCThread::Exit();
   }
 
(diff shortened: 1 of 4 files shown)


wsi/physicaldisplay.cpp
```diff
@@ -80,7 +80,9 @@ void PhysicalDisplay::NotifyClientOfConnectedState() {
   SPIN_UNLOCK(modeset_lock_);
 
   if (refresh_needed) {
-    display_queue_->ForceRefresh();
+    if (!display_queue_->IsIgnoreUpdates()) {
```
kalyankondapally: hmm makes sense thanks.

common/core/gpudevice.cpp
```diff
@@ -35,6 +35,8 @@ bool GpuDevice::Initialize() {
   if (initialization_state_ & kInitialized)
     return true;
 
+  lock_fd_ = open("/vendor/hwc.lock", O_RDONLY);
```
kalyankondapally: Why not just do it in HandleRoutine as before ?

kalyankondapally: You have locks here which should take care of synchronization

common/core/gpudevice.cpp
```diff
@@ -61,9 +63,10 @@ bool GpuDevice::Initialize() {
   initialization_state_ &= ~kHWCSettingsDone;
   thread_sync_lock_.unlock();
 
-  if (use_thread && lock_fd_ == -1) {
+  if (-1 == lock_fd_) {
```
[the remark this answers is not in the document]
shuangwan01: I see use_thread variable has chance to be set to false from code logic. Remove this is to ensure HWCThread::Exit is called if the /vendor/hwc.lock open returns -1. After double reviewed code logic, I believe there is no risk if revert this little change since the thread will be fall into waiting status in HandleWait if lock_fd_ equals -1. I will revert this piece of change and submit a new patch.

common/core/gpudevice.cpp
shuangwan01: Will add a lock for this. Existing code logic will make the thread exit at the end of Initialization method if lock_fd_ remains as default value -1. And HandleWait method relies on value of lock_fd_ to justify if it needs to fall into sleep. So initialize lock_fd_ early to ensure the rest of code logic works as expected.

shuangwan01:
Close ticket and submitted a new pull request based on review result.
https://github.com/intel/IA-Hardware-Composer/pull/366

Issue #274, opened 2018-03-18
Avoid copying buffer details in DrmBuffer Initialize.


kalyankondapally:
https://github.com/intel/IA-Hardware-Composer/blob/master/wsi/drm/drmbuffer.cpp#L51  We copy gemhandles, offset, pitches here. We can avoid this as they are part of meta data already. We should clean this up at some point.

harishkrupo:
So, we just store a reference to the passed HwcBuffer and re-use in other places?

kalyankondapally:
ya

PR #364, opened 2018-05-16
Omit extraneous metadata copy in DrmBuffer

MicheleL-Intel:
DRM buffer details such as gem_handles, pitches, offsets etc are
already present in the image metadata.  Store a reference of the
passed-in HwBuffer object instead of copying.

Jira: https://github.com/intel/IA-Hardware-Composer/issues/274
Tests: Builds on Android

### DIFF
--- a/wsi/drm/drmbuffer.h
+++ b/wsi/drm/drmbuffer.h
@@ -37,15 +37,15 @@ class DrmBuffer : public OverlayBuffer {
                                   ResourceManager* buffer_manager) override;
 
   uint32_t GetWidth() const override {
-    return width_;
+    return mtdt.width_;
   }
 
   uint32_t GetHeight() const override {
-    return height_;
+    return mtdt.height_;
   }
 
   uint32_t GetFormat() const override {
-    return format_;
+    return mtdt.format_;
   }
 
   HWCLayerType GetUsage() const override {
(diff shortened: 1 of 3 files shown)


MicheleL-Intel:
@ashaarunkumar, this PR requires basic tests before I submit for review.  Thanks!

wsi/drm/drmbuffer.h
```diff
@@ -93,16 +93,16 @@ class DrmBuffer : public OverlayBuffer {
...
   void Initialize(const HwcBuffer& bo);
   uint32_t width_ = 0;
   uint32_t height_ = 0;
   uint32_t format_ = 0;
   uint32_t tiling_mode_ = 0;
   uint32_t frame_buffer_format_ = 0;
-  uint32_t pitches_[4];
-  uint32_t offsets_[4];
-  uint32_t gem_handles_[4];
+  HwcBuffer mtdt; /* metadata buffer object */
```
kalyankondapally: hmm no, why cannot we just retrieve the infor from HWcbuffer which is part of the handle

kalyankondapally: ya

MicheleL-Intel: We still need to save the reference to 'HWCBuffer bo' because it is passed in only thru DrmBuffer::Initialize() and being referenced elsewhere.

kalyankondapally: hmm, hwchandle has instance for its bo, which we should use everywhere. in initialize we should make sure that gets populated correctly and use that everywhere as needed.

MicheleL-Intel:
Closing this and creating a new PR.

PR #369, opened 2018-05-17
Fix more Klockwork issues

strassek:
2 more possible null pointer dereferences


strassek:
It looks like the NULL case should already be accounted for by this check:
https://github.com/intel/IA-Hardware-Composer/blob/d5d80636b03b0ca21a704510dd3bc21d01360813/wsi/drm/drmdisplay.cpp#L138

closing for now.

PR #370, opened 2018-05-17
More Klocwork fixes

strassek:


wsi/drm/drmdisplay.cpp
```diff
@@ -135,11 +135,13 @@ bool DrmDisplay::ConnectDisplay(const drmModeModeInfo &mode_info,
   SetPowerMode(power_mode_);
 
   // This is a valid case on DSI panels.
-  if (!broadcastrgb_props)
+  if (broadcastrgb_props == NULL)
```
MicheleL-Intel: Print error message here so that we're aware of problems during runtime.

wsi/drm/drmdisplay.cpp
```diff
@@ -135,11 +135,13 @@ bool DrmDisplay::ConnectDisplay(const drmModeModeInfo &mode_info,
...
     return true;
 
-  if (!(broadcastrgb_props->flags & DRM_MODE_PROP_ENUM))
+  if (!(broadcastrgb_props->flags & DRM_MODE_PROP_ENUM)) {
+    drmModeFreeProperty(broadcastrgb_props);
```
MicheleL-Intel: Good catch but just a minor nit - this is not a NULL pointer dereference issue.  Was this reported by Klockwork?

wsi/drm/drmplane.cpp
```diff
@@ -252,6 +252,8 @@ bool DrmPlane::Initialize(uint32_t gpu_fd,
 
       formats_modifiers_.emplace_back(modifiers_obj);
     }
+
+    drmModeFreePropertyBlob(blob);
```
MicheleL-Intel: Same comment as 'drmModeFreeProperty(broadcastrgb_props)' - this is not a NULL pointer dereference issue.  Was this reported by Klockwork?

wsi/drm/drmdisplay.cpp
strassek: Based on the comment above, I have to assume that this isn't an error path. Maybe an info print would be suitable.

wsi/drm/drmdisplay.cpp
strassek: Klocwork caught this as a possible memory leak

wsi/drm/drmplane.cpp
strassek: Klocwork caught this as a possible memory leak

strassek:
@MicheleL-Intel Added a warning to the case where we return early from ConnectDisplay()

MicheleL-Intel:
LGTM.

PR #371, opened 2018-05-18
Fix dead code and unused variable

strassek:


common/display/displayqueue.cpp
```diff
@@ -820,23 +820,18 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
...
 
-      // Prepare for final composition.
-      if (!compositor_.Draw(current_composition_planes, layers, layers_rects)) {
-        ETRACE("Failed to prepare for the frame composition. ");
-        composition_passed = false;
-      }
+    // Prepare for final composition.
+    if (!compositor_.Draw(current_composition_planes, layers, layers_rects)) {
+      ETRACE("Failed to prepare for the frame composition. ");
+      composition_passed = false;
```
MicheleL-Intel: If 'composition_passed' is no longer utilized in all 'if' conditional checks, remove the declaration, all references and assignments.

os/android/hwcservice.cpp
```diff
@@ -298,7 +298,7 @@ HwcService::Controls::DisplayModeGetAvailableModes(uint32_t display) {
   } else {
     phyDisplay = mHwc.GetExtendedDisplay(display - 1);
   }
-  uint32_t numConfigs, configIndex;
+  uint32_t numConfigs;
```
MicheleL-Intel: Suggest declaring 'numConfigs' as an 'uint8_t'  since the data range will be more than sufficient.

common/compositor/nativesurface.cpp
```diff
@@ -79,7 +79,7 @@ bool NativeSurface::Init(ResourceManager *resource_manager, uint32_t format,
       HWCNativeHandle native_handle;
       resource_manager->GetNativeBufferHandler()->CreateBuffer(
           width_, height_, format, &native_handle, usage, 0);
-      if (!native_handle_) {
+      if (!native_handle) {
```
MicheleL-Intel: This change is correct.  However, as a side comment, I'm wondering if the same call to CreateBuffer on line 54 needs to update a local 'HWCNativeHandle native_handle' similar to this snippet.  That way the 'native_handle_' class member could still maintain a valid handle throughout.

common/display/displayqueue.cpp
strassek: Unfortunately it still is used for checking if these Draw calls work, so I can only get rid of checks prior to the first site that it can possibly get set to false.

common/compositor/nativesurface.cpp
strassek: I like that idea. I'll see if I can make it work.

strassek:
Closing while I figure out errors

PR #373, opened 2018-05-18
Force to ABGR for video

js0701:
Using YUYV somehow can not reach 60FPS for
h264 1080p. This is a temp WA for it on 8Gb board.
Issue not resolved on 2Gb yet.

Jira: https://jira01.devtools.intel.com/browse/OAM-62484

### DIFF
--- a/wsi/Android.mk
+++ b/wsi/Android.mk
@@ -109,6 +109,10 @@ endif
 
 LOCAL_CPPFLAGS += -DENABLE_ANDROID_WA
 
+ifeq ($(strip $(HWC_VIDEO_PERF_WA)), true)
+LOCAL_CPPFLAGS += -DPREFER_ABGR
+endif
+
 LOCAL_MODULE := libhwcomposer_wsi
 LOCAL_CFLAGS += -fvisibility=default
 LOCAL_LDFLAGS += -no-undefined
--- a/wsi/drm/drmplane.cpp
+++ b/wsi/drm/drmplane.cpp
@@ -91,8 +91,15 @@ bool DrmPlane::Initialize(uint32_t gpu_fd,
   uint32_t total_size = supported_formats_.size();
   for (uint32_t j = 0; j < total_size; j++) {
     uint32_t format = supported_formats_.at(j);
+#ifdef PREFER_ABGR
+    // fixme? force to use ABGR8888 to WA h264 1080p 60fps issue
+    // http://jira01.devtools.intel.com/browse/OAM-62484
+    if(format == DRM_FORMAT_ABGR8888) {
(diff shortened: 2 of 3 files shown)


Android.common.mk
```diff
@@ -115,6 +115,9 @@ LOCAL_CPPFLAGS += \
 	-DENABLE_RBC
 endif
 
+LOCAL_CPPFLAGS += \
```
kalyankondapally: This should have a board config, perhaps you can use ifeq ($(strip $(BOARD_CURSOR_WA)), true) for now.

wsi/drm/drmplane.cpp
```diff
@@ -92,7 +92,13 @@ bool DrmPlane::Initialize(uint32_t gpu_fd,
   for (uint32_t j = 0; j < total_size; j++) {
     uint32_t format = supported_formats_.at(j);
     if (IsSupportedMediaFormat(format)) {
-      prefered_video_format_ = format;
+#ifdef PREFER_ABGR
+      // fixme? force to use ABGR8888 to WA h264 1080p 60fps issue
+      // http://jira01.devtools.intel.com/browse/OAM-62484
+      prefered_video_format_ = DRM_FORMAT_ABGR8888;
```
kalyankondapally: You should check if this format is actually supported by this plane, else we cannot do anything use format.

wsi/Android.mk
```diff
@@ -108,6 +108,7 @@ LOCAL_CPPFLAGS += \
 endif
 
 LOCAL_CPPFLAGS += -DENABLE_ANDROID_WA
+LOCAL_CPPFLAGS += -DPREFER_ABGR
```
kalyankondapally: Add Board config check here like i mentioned, you can use ifeq ($(strip $(BOARD_CURSOR_WA)), true)

js0701: If this this require anther PR in Android stack

kalyankondapally: It doesn't, BOARD_CURSOR_WA is already set their. That's why using this instead of adding a new one.

wsi/drm/drmplane.cpp
kalyankondapally: forgot to add these checks ?

js0701: @kalyankondapally  this is WA. If we do anything too further, we may need to do PNP test again

js0701: This WA has pass regression test

kalyankondapally: We might end of up with wierd commit failures, resulting in video composition with rest of content in this case. This could go for verse.

however in this case, Video is always in full screen is it ?

kalyankondapally:
merged

PR #376, opened 2018-05-22
More Klocwork fixes

strassek:
Cleaning up constructor use to protect from double free and defining a type for one of our enums


kalyankondapally:
please merge to beta branch also

PR #377, opened 2018-05-23
Varenderer: should call vaSyncSurfcae before commit

js0701:
now vaSyncSuface return an error 1. need to ignore it

Change-Id: I6aed3e50ef62721ffd76734aaba593df7c2974a9
Tracked-On: https://jira01.devtools.intel.com/browse/OAM-63127


kalyankondapally:
merged

PR #380, opened 2018-05-24
Don't compose Nesteddisplay surface while no SOS side connection

yugang:
submit the this patch for yifei from Randy's team

Nesteddisplay leverage SurfaceFlinger do composition, skip it while no
SOS side buffer sharing request to optimize GPU consumption.

Change-Id: I40d5ba07c45af9c4e2a0b58090cbb36641b311b0
Jira: None
Test: Tested on android
Signed-off-by: xue.yifei <yifei.xue@intel.com>


common/core/nesteddisplay.cpp
```diff
@@ -195,97 +195,97 @@ bool NestedDisplay::Present(std::vector<HwcLayer *> &source_layers,
...
-          mHyperDmaExportedBuffers[sf_handle].counter = 0;
-          mHyperDmaExportedBuffers[sf_handle].hyper_dmabuf_id = msg.hid;
-          mHyperDmaExportedBuffers[sf_handle].surface_id = (uint64_t)sf_handle;
-          strncpy(mHyperDmaExportedBuffers[sf_handle].surface_name, "Cluster",
-                  SURFACE_NAME_LENGTH);
-          mHyperDmaExportedBuffers[sf_handle].bbox[0] = display_frame.left;
-          mHyperDmaExportedBuffers[sf_handle].bbox[1] = display_frame.top;
-          mHyperDmaExportedBuffers[sf_handle].bbox[2] = buffer->GetWidth();
-          mHyperDmaExportedBuffers[sf_handle].bbox[3] = buffer->GetHeight();
+  if (mconnected) {
```
kalyankondapally: just do an early return here ? indentation needs to be fixed.

feifanx: yes, if no sharing request from SOS side, hwc will mark the nesteddisplay's layers to device type, then do nothing . While got sharing request from SOS side, hwc will leverage surfaceflinger do composition, then share the FB to SOS.

kalyankondapally: I mean u could just have if (!mconnected)
                                          return false;

PR #382, opened 2018-05-24
Remove unused files.

kalyankondapally:
Jira: None.
Test: Build passes on Android with and without hwc-val test suite.


kalyankondapally:
@strassek  Can u help test this with hwc-val included

strassek:
It looks like hwc-val includes both of these files.

Can we instead completely revert 98ba8f616c03a4cdb242cf65e252cc96a54f6087?
None of that code should live under common/ anyway... I'd rather land it properly when the reset of hwc-val becomes ready.

PR #383, opened 2018-05-24
Add google test structure for unittests

csravelar:
This patch kicks off what we need to start adding files for unittesting
our cpp files with google tests. Source files are from googletest
release-1.8.0 and can be found at https://github.com/google/googletest

Jira: GSE-1588
Test: Ran sample google tests to check compiled and ran units tests properly


strassek:
Thanks! In order to land this I think we would want at least 1 real test case and we need to figure out how to offer this as a build target. Also, in the patch description can you document the location/version of where you pulled this code from?

PR #384, opened 2018-05-25
Ensure DisplayManager is initialized before accessing it.

kalyankondapally:
Jira: None.
Test: No crash observed when IgnoreUpdates is called from
      Worker Thread.


kalyankondapally:
merged

PR #385, opened 2018-05-26
Add option to disable Plane Scaling.

kalyankondapally:
When running in VM/Hypervisour, gvtg might not have needed
support for plane scaling. Add option to disable plane
scaling for video in this case.

Jira: None.
Test: Video shows correctly when played in full screen.


strassek:
I don't see this in master. Can we try landing it there first?

kalyankondapally:
No, this is not meant for master (temporary solution only). Just for beta branch and restricted to hypervisor environment on 4.9 LTS only.

kalyankondapally:
not needed as scaling was disabled in kernel when using gvtg.

PR #386, opened 2018-05-28
Fixed the HWC crash due to improper thread initialization sequence.

shuangwan01:
The thread of GpuDevice is started before DRM Display Manager's Initialization method, there is a chance that GpuDevice thread call method of display manager but underlying DRM display manager implementation still not fully initialized yet. This will lead HWC crash.

JIRA: OAM-63288
Test: Switch clone & mosaic mode and no crash during boot.

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -51,11 +51,6 @@ bool GpuDevice::Initialize() {
 
   display_manager_.reset(DisplayManager::CreateDisplayManager(this));
 
-  if (!InitWorker()) {
-    ETRACE("Failed to initalize thread for GpuDevice. %s", PRINTERROR());
-    use_thread = false;
-  }
-
   bool success = display_manager_->Initialize();
   if (!success) {
     return false;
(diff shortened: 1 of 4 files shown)


common/core/gpudevice.cpp
```diff
@@ -64,6 +59,10 @@ bool GpuDevice::Initialize() {
   HandleHWCSettings();
   InitializeHotPlugEvents();
 
+  if (!InitWorker()) {
```
kalyankondapally: shouldn't be needed. DisplayManager is now created before DisplayManager is created.

kalyankondapally: check with latest code.

shuangwan01: Yes, DisplayManager is created but not fully initialized. If InitWorker called just next to DisplayManager creation, thread GpuDevice will be put into running status. It has a chance to call methods of display manager but the later one might be not fully initialized yet. The consequence of it is HWC might be crash unexpectedly.

kalyankondapally: in this case initializing thread doesn't make sense if lock_fd_ == -1 and also in HandleRoutine u can remove call to HandleHWCSettings

kalyankondapally: Also u can remove initialization_state_lock_ in  InitializeHotPlugEvents and HandleHWCSettings

shuangwan01: For commit afa1e13, the GpuDevice thread initialization sequence works for me, thanks

PR #387, opened 2018-05-29
Stop assuming that enums is a valid pointer when count_enums > 0

strassek:
If libdrm fails to allocate enums for some reason it will still give us a
value in count_enums that it meant to allocate. This would be a rare
condition but we ought to handle it.

Test: Build and boot on Android
Bug: GSE-1586


csravelar:
lgtm

Issue #303, opened 2018-03-27
FrameBufferManager Instance should be owned by GPUDevice.


kalyankondapally:
This instance manages all fb's life time and is unique per gpu device. Currently, it's allocated on demand and sticks around. Instead, GpuDevice should own it and pass it to all displays during initialization. Displays can pass it to compositor thread and drmbuffer for fb creation and deletion.

csravelar:
Hmm, I was able to rebuild this for android ai without errors, but I have some doubts about this patch. I need more context as to why we are trying this refactor. Is there some gain we get from making this change? It doesn't seem like anything fundamentally changes about how we are using FrameBufferManager but how compositorthread and drmbuffer come about getting the instance. 

Currently: Only drmbuffer and compositorthread are the only ones to need the FrameBuffer and only call GetInstance 2-3 times each. Getting a direct singleton instance of FrameBufferManager instance.

After patch: FrameBufferManager instance would be passed through gpudevice > drmdisplaymanager > displays > displayqueue > compositor > compositorthread and for drmbuffer it would need to go through a group of other places. This seems like we are propagating the framebuffer instance to a bunch of places that never use it or need it in addition to making the methods seem like they need the instance for more than just to pass it off somewhere else. Could we discuss this issue more?

PR #388, opened 2018-05-30
FrameBufferManager Instance owned by GPUDevice

csravelar:
This patch addresses issue #303 and has the GpuDevice take ownership
of the FrameBufferManager Instance which then gets passed to all
displays on initialization to pass to the drmbuffer and the
compositorthread as needed in place of calling the instances from
FrameBufferManager directly.

Jira: GSE-1502
Test: android_ia and HWC build passes without errors


kalyankondapally:
merged

PR #390, opened 2018-06-01
Fix Mesa GBM and drop minigbm for Linux build

strassek:
Supporting both minigbm and Mesa GBM seems to be intractable in their current state. One issue is that minigbm advertises itself as a standard gbm implementation, despite not having a compatible API. Let's drop bindings for minigbm and start using Mesa GBM full-time in Linux builds. If we get a user that has some hard requirement on minigbm, we can add it back as an experimental configuration, provided that it doesn't collide with the Mesa GBM installation.


kalyankondapally:
merged

PR #391, opened 2018-06-04
Fix compilation warnings.

harishkrupo:
As a consequence of this change, linux_test app doesn't
crash when compiled using gcc 8.1.0

Jira: None
Test: Compile the test app with gcc 8.1.0 and run. It shouldn't
      crash.


kalyankondapally:
merged removing GetAllDisplay related changes, let's do that as separate PR.

PR #398, opened 2018-06-12
ignore interlace modes

zhiquan-liu:
In AOSP PPR1.180528.001 version code, connector mode config list
reported to SurfaceFlinger are changed to unsorted.
Interlace mode config will be chosen and set in display. This will cause
display shake issue. Interlace mode configs are not requirement on
android, so we ignore them(the flags with DRM_MODE_FLAG_INTERLACE).

Change-Id: I2397045ce127046b942b974f30f1d4d8618a3f80
Tracked-On: https://jira01.devtools.intel.com/browse/OAM-64528
Signed-off-by: Liu Zhiquan <zhiquan.liu@intel.com>

### DIFF
--- a/wsi/drm/drmdisplay.cpp
+++ b/wsi/drm/drmdisplay.cpp
@@ -464,7 +464,12 @@ void DrmDisplay::SetDrmModeInfo(const std::vector<drmModeModeInfo> &mode_info) {
   uint32_t size = mode_info.size();
   std::vector<drmModeModeInfo>().swap(modes_);
   for (uint32_t i = 0; i < size; ++i) {
-    modes_.emplace_back(mode_info[i]);
+#ifdef ENABLE_ANDROID_WA
+    //FIXME: SurfaceFlinger can't distinguish interlace mode config.
+    //interlace mode is not requirement for android, ignore them.
+    if (!(mode_info[i].flags & DRM_MODE_FLAG_INTERLACE))
+#endif
+      modes_.emplace_back(mode_info[i]);
   }
 
   SPIN_UNLOCK(display_lock_);


js0701:
Hi,
HWC also used in Linux and AndroidOS
Can you explain why this WA in HWC instead in framework?

zhiquan-liu:
android framework could not get the attribute of interlace.  
In DrmDisplay::GetDisplayAttribute, we can get width, height, refresh rate, dpiX and dpiY attribute.

kalyankondapally:
we probably make an option to add configs in reverse order or unsorted in our side ?

zhiquan-liu:
this issue root cause is that surfaceflinger will find two mode configs and can't distinguish them.
For example:
id 80:"1920x1080" freq 60 clock 148500 hdisp 1920 hss 2008 hse 2052 htot 2200 vdisp 1080 vss 1084 vse 1089 vtot 1125 type 0x48 flags 0x5
id 82:"1920x1080i" freq 60 clock 74250 hdisp 1920 hss 2008 hse 2052 htot 2200 vdisp 1080 vss 1084 vse 1094 vtot 1125 type 0x40 flags 0x15
these two modes cat from i915_display_info, they have same attributes.(after refresh *2 for interlace mode).

the sort of mode configs can’t guarantee which one will be choose. Interlace mode configs are not requirement for android, so I prefer to ignore it. 
If I wrap the modification with “ENABLE_ANDROID_WA”, is it OK?

By the way, on Android if interlace mode config is chose, there have display issue. Is HWC support interlace mode config?

zhiquan-liu:
@js0701 surfaceflinger can't get all atrrubite.  
the modification is wrap with ENABLE_ANDROID_WA flag in the update.

PR #399, opened 2018-06-18
HwcLayer: Apply rotation transformation to damage.

harishkrupo:
This enables us to remove the workaround in Overlaylayer.

Jira: None
Test: No regressions seens on Android and Linux.

### DIFF
--- a/common/core/hwclayer.cpp
+++ b/common/core/hwclayer.cpp
@@ -332,17 +332,54 @@ const HwcRect<int>& HwcLayer::GetLayerDamage() {
 
   current_rendering_damage_.reset();
 
-  // From observation: In Android, when the source crop doesn't
-  // begin from (0, 0) the surface damage is already translated
-  // to global display co-ordinates
-  if (!surface_damage_.empty() &&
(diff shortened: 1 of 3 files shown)


common/core/hwclayer.cpp
```diff
@@ -325,25 +325,69 @@ bool HwcLayer::IsCursorLayer() const {
   return is_cursor_layer_;
 }
 
+HwcRect<int> HwcLayer::TransformDamage() {
+  HwcRect<int> ret;
+  int ox = 0, oy = 0;
+  HwcRect<int> tr_surface_damage =
```
kalyankondapally: instead of allocating temp rects, u can use current_Rendering_Damage_ ?

harishkrupo: I think that would impact the readability of the code? 
I was thinking of removing return value and just assign the ret rect to current_rendering_damage.

kalyankondapally: hmm no. You can move this chunk of code to the calling function itself I guess to keep it all in one place.

harishkrupo: okay.

kalyankondapally:
merged

PR #403, opened 2018-06-29
Fix Klockwork indirect path traversal issue.

harishkrupo:
Jira: None
Test: File path on
      Android: /vendor/etc/hwc_display.ini
      Linux: {prefix}/etc/hwc_display.ini


js0701:
LGTM if std::getenv is the klockwork issue

PR #404, opened 2018-07-06
MosaicMode: return more reliable fences

harishkrupo:
Currently we pick a display with the maximum refresh rate
as the preferred display and return its fence.
Instead, merge the fences from all the displays and return it.
Such merged fences are signaled when all the individual fences
are signaled.

Jira: None
Test: No regressions seen on Linux and Android

### DIFF
--- a/common/core/mosaicdisplay.cpp
+++ b/common/core/mosaicdisplay.cpp
@@ -16,6 +16,7 @@
 
 #include "mosaicdisplay.h"
 
+#include <libsync.h>
 #include <sstream>
 #include <string>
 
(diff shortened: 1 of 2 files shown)


harishkrupo:
@kalyankondapally travis failed because wayland-egl isn't available in external-mesa after the latest rebase.
We will need to update external-wayland as wayland-egl is now moved there.

kalyankondapally:
@harishkrupo updated wayland and weston repo.

harishkrupo:
@kalyankondapally weston has changed the output handling code which would need some changes in the iahwc plugin too. Could we handle that later and just update only wayland for now?

PR #406, opened 2018-07-09
 Omit extraneous metadata copy in DrmBuffer

MicheleL-Intel:
DRM buffer details such as width, height etc are already present
  in the image metadata.  Store a reference of the passed-in HwcBuffer
  object with the exception of the shorthand buffer aliases (such as
  pitches_[] and offsets_[] which are useful for readibility.

  Jira: GSE-1509
  Test: Tested on KabyLake NUC

### DIFF
--- a/wsi/drm/drmbuffer.cpp
+++ b/wsi/drm/drmbuffer.cpp
@@ -78,13 +78,6 @@ DrmBuffer::~DrmBuffer() {
 }
 
 void DrmBuffer::Initialize(const HwcBuffer& bo) {
-  width_ = bo.width_;
-  height_ = bo.height_;
-  for (uint32_t i = 0; i < 4; i++) {
-    pitches_[i] = bo.pitches_[i];
-    offsets_[i] = bo.offsets_[i];
-    gem_handles_[i] = bo.gem_handles_[i];
-  }
 
   format_ = bo.format_;
   if (format_ == DRM_FORMAT_NV12_Y_TILED_INTEL || format_ == DRM_FORMAT_NV21)
@@ -92,10 +85,7 @@ void DrmBuffer::Initialize(const HwcBuffer& bo) {
   else if (format_ == DRM_FORMAT_YVU420_ANDROID)
     format_ = DRM_FORMAT_YUV420;
 
-  tiling_mode_ = bo.tiling_mode_;
-  usage_ = bo.usage_;
-
-  if (usage_ == hwcomposer::kLayerCursor) {
+  if (image_.handle_->meta_data_.usage_ == hwcomposer::kLayerCursor) {
     // We support DRM_FORMAT_ARGB8888 for cursor.
     frame_buffer_format_ = DRM_FORMAT_ARGB8888;
   } else {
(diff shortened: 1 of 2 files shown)


wsi/drm/drmbuffer.cpp
```diff
@@ -77,23 +77,43 @@ DrmBuffer::~DrmBuffer() {
...
-    offsets_[i] = bo.offsets_[i];
-    gem_handles_[i] = bo.gem_handles_[i];
-  }
 
-  format_ = bo.format_;
+  SaveMetaData(bo);
+  width_ = GetWidth();
+  height_ = GetHeight();
+
+  format_ = GetFormat();
```
strassek: I think we talked about this off-line, but I can't remember the reasoning behind keeping some of these copies around. Would it be possible to drop these and use only the image_.handle_->meta_data_ copies?

wsi/drm/drmbuffer.cpp
```diff
@@ -77,23 +77,43 @@ DrmBuffer::~DrmBuffer() {
...
+  }
+
+  for (uint32_t i = 0; i < 8; i++) {
+    image_.handle_->meta_data_.fb_modifiers_[i] = bo.fb_modifiers_[i];
+  }
+}
+
 void DrmBuffer::Initialize(const HwcBuffer& bo) {
   width_ = bo.width_;
   height_ = bo.height_;
```
strassek: Looks like these 2 assignments can be dropped.

MicheleL-Intel:
Thanks Kevin for your comments.  

Based on my recollection, we discussed this fix as being dual-phased.  First phase is to get rid of the pitches_[], offsets_[], and gem_handles_[] buffer writes which this PR resolves.  

Other than the fact that we want to eliminate unnecessary copies, every time we instantiate DrmBuffer, it creates the class member buffer objects which can be a synchronization issue when or if the data does change.  As far as I can tell, the buffers are set only once during  DrmBuffer::Initialize()

Not sure what others call these but I refer to them as "shorthand aliases".  I believe they were created for readability.

The second phase is to introduce macros with token concatenation, which will eliminate these shorthand aliases altogether.

Example define and usage (which will be incorporated in Phase 2)
# define METADATA (PARAM)     {image_.handle_->meta_data_.##PARAM}
METADATA (width_)
METADATA (height_)
METADATA (pitches_)
METADATA (offsets_)

strassek:
Understood, for consistency I'd like to see the remaining copies (width_, height_, etc..) removed, and use the long-form image_.handle... in their place.

MicheleL-Intel:
I will revise as suggested,  Thanks again.

MicheleL-Intel:
Removed the shorthand aliases, however, I'm having build issues with latest and errors appear to be unrelated to the changes.  Cleaned and currently rebuilding.

wsi/drm/drmbuffer.cpp
```diff
@@ -77,29 +77,36 @@ DrmBuffer::~DrmBuffer() {
 #endif
 }
 
-void DrmBuffer::Initialize(const HwcBuffer& bo) {
-  width_ = bo.width_;
-  height_ = bo.height_;
+void DrmBuffer::SaveMetaData(const HwcBuffer& bo) {
+  image_.handle_->meta_data_.width_ = bo.width_;
```
kalyankondapally: hmm do we really need to do this ?

in Initialize(image_.handle_->meta_data_);  it's the same metadata struct right ? we can avoid this whole copy. ?

MicheleL-Intel: We discussed this briefly and I thought you said the image_.handle_.metadata doesn't get populated during init but until it's imported.  I will remove SaveMetaData.

harishkrupo:
@MicheleL-Intel Please rebase on top of master, travis issues are now fixed.

MicheleL-Intel:
Rebased, thanks.

wsi/drm/drmbuffer.cpp
```diff
@@ -78,28 +78,12 @@ DrmBuffer::~DrmBuffer() {
...
-  width_ = bo.width_;
-  height_ = bo.height_;
-  for (uint32_t i = 0; i < 4; i++) {
-    pitches_[i] = bo.pitches_[i];
-    offsets_[i] = bo.offsets_[i];
-    gem_handles_[i] = bo.gem_handles_[i];
-  }
-
-  format_ = bo.format_;
-  if (format_ == DRM_FORMAT_NV12_Y_TILED_INTEL || format_ == DRM_FORMAT_NV21)
```
kalyankondapally: You removed these checks ?

MicheleL-Intel: These checks are for setting format_ which is no longer needed since the shorthand aliases are being removed.

kalyankondapally: We still need this as these are custom formats on Android which we translate to standard DRM enums.

MicheleL-Intel: Is it just format_ that needs to be preserved?

kalyankondapally: I think u forgot to update the PR ?

MicheleL-Intel: Hmmm, I thought I did so.  Let me check tomorrow.

wsi/drm/drmbuffer.cpp
```diff
@@ -337,20 +327,20 @@ const MediaResourceHandle& DrmBuffer::GetMediaResource(MediaDisplay display,
 
   VASurfaceAttribExternalBuffers external;
   memset(&external, 0, sizeof(external));
-  uint32_t rt_format = DrmFormatToRTFormat(format_);
+  uint32_t rt_format = DrmFormatToRTFormat(image_.handle_->meta_data_.format_);
```
kalyankondapally: should be just format_ here and all other places which used format_ ? or wondering if we can store format value directly in meta_data_. did u check that and saw any failures ?

MicheleL-Intel: You're right, this should be format_, I didn't see any failures when I tested this.  I'll check to see if format_ is being referenced elsewhere just in case.

js0701:
@MicheleL-Intel  I like the idea of Phase2. How about do it and merge to this PR? Currently this PR make the code a little verbose

MicheleL-Intel:
Alright I'll submit a new patch with the macros.

MicheleL-Intel:
Replaced by #417

PR #408, opened 2018-07-18
  Fix hwcomposer ANDROID_VERSION parser

MicheleL-Intel:
Current parser yields the major version only and not the
  ANDROID_VERSION as intended. This fix first evaluates the
  major version, which is then used to map to the corresponding
  ANDROID_VERSION correctly.

  Jira: GSE-1615
  Test: Compiled successfully on Celadon

### DIFF
--- a/Android.common.mk
+++ b/Android.common.mk
@@ -20,8 +20,28 @@ HWC_PATH := $(call my-dir)
 HWC_VERSION_GIT_BRANCH := $(shell pushd $(HWC_PATH) > /dev/null; git rev-parse --abbrev-ref HEAD; popd > /dev/null)
 HWC_VERSION_GIT_SHA := $(shell pushd $(HWC_PATH) > /dev/null; git rev-parse HEAD; popd > /dev/null)
 
-# Obtain Android Version
-ANDROID_VERSION := $(word 1, $(subst ., , $(PLATFORM_VERSION)))
+
+#------------------------------------------------------------------------
+#  1. Evaluate ANDROID_VERSION from PLATFORM_VERSION
+#  2. First letter may be the major version or the dessert letter
+#------------------------------------------------------------------------
+ANDROID_M := 6
+ANDROID_N := 7
+ANDROID_O := 8
+ANDROID_VERSION_6 := $(ANDROID_M)
+ANDROID_VERSION_7 := $(ANDROID_N)
+ANDROID_VERSION_8 := $(ANDROID_O)
+ANDROID_VERSION_O := $(ANDROID_O)
+ANDROID_VERSION_OMR1 := $(ANDROID_O)
+ANDROID_SUPPORTED_VERSIONS := 6 7 O 8 OMR1
+
+ANDROID_MAJOR_VERSION := $(word 1, $(subst ., , $(PLATFORM_VERSION)))
+$(info "PLATFORM_VERSION $(PLATFORM_VERSION)")
+$(info "ANDROID_MAJOR_VERSION $(ANDROID_MAJOR_VERSION)")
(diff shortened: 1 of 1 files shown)


kalyankondapally:
LGTM

harishkrupo:
Suggestion: We could also use PLATFORM_SDK_VERSION (see [here](https://gitlab.freedesktop.org/mesa/mesa/blob/master/Android.common.mk#L48))
Then, instead of comparing for 7/8 we could compare for 26/27. 
I believe this would be a bit cleaner :)

MicheleL-Intel:
Thanks for the suggestion Harish.  I will take a look.  Do you happen to have a sample content for PLATFORM_SDK_VERSION?

harishkrupo:
Yes. Here: https://gitlab.freedesktop.org/mesa/mesa/blob/master/Android.common.mk#L48 
and its usage: https://gitlab.freedesktop.org/mesa/mesa/blob/master/src/egl/drivers/dri2/platform_android.c#L669

harishkrupo:
@MicheleL-Intel I have no objections with this patch. :)
I was only suggesting that using the PLATFORM_SDK_VERSION would have reduced all those calculations to a single line.

this way:
$(eval ANDROID_VERSION := $(PLATFORM_SDK_VERSION))

Then replace [This](https://github.com/intel/IA-Hardware-Composer/blob/f81a004cf8403b648aeac4c1cb4181aba171e7c4/os/android/libhwcservice/test/Android.mk#L25) and [This](https://github.com/intel/IA-Hardware-Composer/blob/e5de429aa90517b06bd127f55ce2dd1e9a3dbfd6/Android.common.mk#L134) to 
ifeq ($(shell test $(ANDROID_VERSION) -ge 27; echo $$?), 0)
You can refer to this [link](https://en.wikipedia.org/wiki/Android_version_history) to convert between android version numbers to api levels

js0701:
@MicheleL-Intel @harishkrupo  This PR is ready to merge? Will merge if no more issues

js0701:
@MicheleL-Intel 
Do we define Android_P? or need to define? GP2.0 is targeting Android_P

MicheleL-Intel:
I think it's best to merge now since this is a customer request.  I can add Android_P later after I test the change proposed by Harish.

js0701:
merged

MicheleL-Intel:
@harishkrupo , using $PLATFORM_SDK_VERSION as the seed as you've suggested yields API_LEVEL=27.  Then we have to convert from API_LEVEL to version, in addition to changing the "if" conditional statements in Android.mk and Android.common.mk.  

It is about the same amount of code.  Plus I think the current version is more intuitive.  Please let me know if I'm missing something.

harishkrupo:
@MicheleL-Intel Ok sure. Please pick your prefered implementation :)

MicheleL-Intel:
Thanks Harish.  @js0701, I will be adding ANDROID_P to the search list since it's required for GP2.0.

tpalli:
Note, this patch is causing a change error when building:

"/bin/bash: line 0: test: -ge: unary operator expected"

I'm not sure if it causes any problems but it looks a bit nasty.

harishkrupo:
Maybe ANDROID_VERSION is not defined?
@tpalli are you using Android P?

js0701:
@kalyankondapally  I think we can added P definition here since it is just a name and it does not expose any source code of P
@MicheleL-Intel

harishkrupo:
@js0701 FYI, P is now open sourced and Celadon has moved to P.

tpalli:
@MicheleL-Intel OK, yeah this issue is in P

MicheleL-Intel:
@tpalli  I'll check against Android P.

PR #410, opened 2018-07-19
Recalculate display frame when rotation is set in *.ini

harishkrupo:


kalyankondapally:
@harishkrupo can u check why travis is failing

harishkrupo:
Checked. Something to do with libva's wayland backend/plugin.

harishkrupo:
Sorry, not libva but vaapi
i965_decoder_utils.c: In function â€˜intel_decoder_check_avc_parameterâ€™:
  CC       libi965_drv_video_la-i965_encoder.lo
i965_decoder_utils.c:966:9: warning: â€˜num_slice_groups_minus1â€™ is deprecated (declared at /tmp/hwc-install/include/va/va.h:2930) [-Wdeprecated-declarations]
         if (pic_param->num_slice_groups_minus1 ||
         ^
  CC       libi965_drv_video_la-i965_encoder_utils.lo
i965_drv_video.c: In function â€˜i965_GetConfigAttributesâ€™:
i965_drv_video.c:1092:45: error: â€˜struct <anonymous>â€™ has no member named â€˜roi_rc_qp_delat_supportâ€™
                             roi_config->bits.roi_rc_qp_delat_support = 0;
                                             ^
i965_drv_video.c:1097:45: error: â€˜struct <anonymous>â€™ has no member named â€˜roi_rc_qp_delat_supportâ€™
                             roi_config->bits.roi_rc_qp_delat_support = 1;
                                             ^
make[3]: *** [libi965_drv_video_la-i965_drv_video.lo] Error 1
make[3]: *** Waiting for unfinished jobs....
make[3]: Leaving directory `/tmp/vaapi/src'
make[2]: *** [all-recursive] Error 1
make[2]: Leaving directory `/tmp/vaapi/src'
make[1]: *** [all] Error 2
make[1]: Leaving directory `/tmp/vaapi/src'
make: *** [all-recursive] Error 1

We are syncing the vaapi source from https://github.com/projectceladon/vaapi 
Is this right?

kalyankondapally:
@harishkrupo We should just use vaapi from upstream directly.

js0701:
@harishkrupo  Do you think still any issue with this PR?
Will merge if no

harishkrupo:
@js0701 it is still in progress. Will let you know once it is ready

js0701:
@harishkrupo 
I don't quite undestand the logic here.
If display rotation transform is applied at the input overlay layers, there are few problems
1) The logic is complicate, as we need also consider window level rotation
2) Does this mean we don't leverage display hardware rotation feature anymore, instead we use GPU rotation? If we can levergage display hardware rotation, the layer (display frame) has to be in the original not rotated rectangle.

harishkrupo:
Apologies, I should have commented the code. 
> I don't quite undestand the logic here.
> If display rotation transform is applied at the input overlay layers, there are few problems

> The logic is complicate, as we need also consider window level rotation

Sorry, I didn't understand window level rotation. Could you please explain?

> Does this mean we don't leverage display hardware rotation feature anymore, instead we use GPU rotation? If we can levergage display hardware rotation, the layer (display frame) has to be in the original not rotated rectangle.

No. When we have to leverage the display rotation and when the layers need to go for a composition, We need to get back the source layers' rotation to the original rotation. This is because once the composition is done, the composed layer as a whole would be rotated. For example, consider that there are 3 layers for gpu composition and the rotation is 180 degrees. Display size is (1920, 1080)
Display frames: 
Layer 1: (0,24) (1920,1080)
Layer 2: (32, 990) (88, 1080)
Layer 3: (1832, 990) (1888, 1080)

When rotation is applied to the input layers: the display frames are
Layer 1: (0,0) (1920, 1056)
Layer 2: (1832, 0) (1888, 90)
Layer 3: (32, 0) (88, 90)

Now suppose these layers were going for gpu composition and display rotation is in use, the composition would go wrong as the display frames are now wrong. Thus, there is a rotation to get the layer back to normal rotation just before the render pass.

I have tried my best to explain here. Please let me know if you need more clarification. :)

js0701:
@harishkrupo 
Thanks the explain.
Window level means input layer rotation instead of display rotation.

As your explaination, Can we avoid tranform twice?
Try to aoiding complicated logic here for complexity usually bring bugs

harishkrupo:
@js0701 The reason for rotating the layer in the very beginning is to check if the current source crop - display frame combination could be directly scanned out. Unfortunately, once the display plane state list is created, we will need to apply the transformation again but this time only for those cases which satisfy the properties: 
1. We need GPU composition
2. The frame is rotated by using plane rotation
3. There is some plane transformation set in the hwc_display.ini

However, once the composition is over, we need to get those layers back to the previous rotation so that they are set correctly for the plane properties. 
Also, the reason all the layers are rotated, not just the plane state's source layers' is because we need the dedicated layer's display frame too while calculating composition regions.

js0701:
@harishkrupo 
For complexity here, we are going to lunch test those those patchs before it merged to master. Here are tests I think need
1) Window rotation + display rotaion (90 180 270) combination
2) Mosaic Mode and Clone Mode display roation
Anyway please ping me when the patches ready for test. I think test is important here

harishkrupo:
@js0701 Ok sure

harishkrupo:
@js0701 What is the expectation of rotation in mosaic mode? I believe it should rotate the whole  (conjoint) display, correct? 
Currently, the rotation that we set is for a per-display basis. We will need to modify that before working on mosaic mode rotation. 
I will post the updated patches shortly. With them, I have tested single display rotation (window rotation + hwc rotation). I believe we should merge these for now and then start work on mosaic display as it would require a bit more work.

js0701:
I think we can define the behavior. There is no standard definition of "mosaic mode"

harishkrupo:
ok sure. Then I think that can be done after this. This PR targets single display ini rotation

stripes416:
@harishkrupo -- can you go ahead and comment your patches here? The explanations in the conversation between you and @js0701 were really helpful when trying to understand what was going on -- we should probably just add them into the code to aid in understanding in the future.

PR #414, opened 2018-07-30
Android: workaound for 3 display underrun issue.

js0701:
Without modificator no serious unerrun when
3 display in use
Need to revert this when root fix in kernel

Tests: 3 display in Android clone mode or mosaic mode
Jira: None

### DIFF
--- a/Android.common.mk
+++ b/Android.common.mk
@@ -160,4 +160,9 @@ LOCAL_CPPFLAGS += \
 	-DDISABLE_CURSOR_PLANE
 endif
 
+ifeq ($(strip $(BOARD_THREEDIS_UNDERRUN_WA)), true)
+LOCAL_CPPFLAGS += \
+        -DTHREEDIS_UNDERRUN_WA
+endif
+
 endif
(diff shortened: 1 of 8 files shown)


Android.common.mk
```diff
@@ -160,4 +160,9 @@ LOCAL_CPPFLAGS += \
 	-DDISABLE_CURSOR_PLANE
 endif
 
+ifeq ($(strip $(BOARD_THREEDIS_UNDERRUN_WA)), true)
+LOCAL_CPPFLAGS += \
+        -DTHREEDIS_UNDERRUN_WA
```
kalyankondapally: hmm We should enable this dynamically during run time. Probably, DrmDisplayManager, should check connected displays and set a flag in DisplayQueue, DisplayPlaneManager ? This way atleast we can use compression in supported combinations.

kalyankondapally: Even with kernel investigation temporary should be the dynamic path like we do with cursor now. Where we try to disable it's usage when all three pipes are enabled. You can check that and do the same here.

js0701: @kalyankondapally thanks for the comments.
Done the changes accrodingly

js0701:
closed as merged

PR #415, opened 2018-07-30
Add note to doxy installation instructions readme.

csravelar:
Since we are using the HAVE_DOT option in the doxygen config file,
we should make a note to the user that they may need to install
graphviz if they do not already have it installed on their machine.

Tests: Run Doxygen
Jira: None

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,12 @@ To install doxygen on ubuntu simply run:
 sudo apt install doxygen
 ```
 
+Note: You may also need to install graphviz if you do not have it already.
+
+```
+sudo apt install graphviz
+```
+
 Once installed, you should provide documentation to any new features you add to
 the project or modify/create documentation for any big changes in methods,
 classes, etc that you change.


Doxyfile
```diff
@@ -811,7 +811,7 @@ RECURSIVE              = YES
 # Note that relative paths are relative to the directory from which doxygen is
 # run.
 
-EXCLUDE                =
+EXCLUDE                = travisci/ tools/ tests/
```
strassek: As long as we are explicitly including files in INPUT does EXCLUDE really have any effect?

Doxyfile
```diff
@@ -811,7 +811,7 @@ RECURSIVE              = YES
 # Note that relative paths are relative to the directory from which doxygen is
 # run.
 
-EXCLUDE                = travisci/ tools/ tests/
+EXCLUDE                =
```
strassek: So now you are dropping these?

common/core/logicaldisplaymanager.h
```diff
@@ -29,21 +29,55 @@ class LogicalDisplayManager;
...
+  * @param physical_display a NativeDisplay object representing a physical
+  *                         display
+  */
   LogicalDisplayManager(NativeDisplay* physical_display);
   ~LogicalDisplayManager();
 
+  /**
+  * Initialize new logical displays up to the passed in total count.
+  *
+  * @param total an unsigned 32 bit integer to dictate the amount of logical
```
strassek: In general I don't think we *need* to document standard data types, unless the acceptable input is some subset of that data type.

common/core/logicaldisplaymanager.h
```diff
@@ -29,21 +29,55 @@ class LogicalDisplayManager;
 
 class LogicalDisplayManager {
  public:
+  /**
+  * Constructor that sets the physical display for the class.
+  *
+  * @param physical_display a NativeDisplay object representing a physical
+  *                         display
```
strassek: Formatting here looks off

csravelar:
Whoops, got a wip branch entangled with this one. Only the readme is supposed to be part of this patch.
@strassek, yes, I plan to remove test and tools from the documentation generation. There are no .h files and I don't see a good reason why we need to include them in the recursion list. Also don't see why we shouldn't document datatypes for granularity unless we are going to make note of it somewhere to inform new contributors of that decision, then that's fine.

stripes416:
@strassek It looks like the changes you requested are on commits that no longer exist. I'm going to go ahead and dismiss your request for changes since this commit is just additions to the readme.

PR #420, opened 2018-08-02
Add support for future doxygen comments in both *.h and *.cpp files

stripes416:
With these changes, every file will be listed, but only those files that
contain doxygen comments will be available for viewing. Each *.h or *.cpp with
doxygen comments will contain the \file command, indicating to doxygen that
comments should be parsed within the file and displayed.

Jira: GSE-1575
Tests: Run Doxygen

Signed-off-by: Kelly Ledford <kelly.ledford@intel.com>


csravelar:
Close and open this PRto check travis passes. Otherwise, looks good to merge

PR #421, opened 2018-08-03
Remove hack job to fix travisci

csravelar:
This patch removes a problematic/outdated line in travisci that
was causing false error fails to all pull requests.

Jira: GSE-1642
Tests: TravisCi passes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -161,10 +161,6 @@ script:
 #Build weston and weston plugin
 - git clone -b iahwc-plugin https://github.com/intel/external-weston.git ../weston
 - cd ../weston
-#HACK: Weston's paths are hardcoded to ../iahwc/..
-- cp -r ../IA-Hardware-Composer ../iahwc
-- cd -
-- os/linux/weston/build_script.sh --weston-dir=../weston --iahwc-dir=$(pwd) --build
 
 branches:
   only:


.travis.yml
```diff
@@ -161,10 +161,6 @@ script:
 #Build weston and weston plugin
 - git clone -b iahwc-plugin https://github.com/intel/external-weston.git ../weston
 - cd ../weston
-#HACK: Weston's paths are hardcoded to ../iahwc/..
-- cp -r ../IA-Hardware-Composer ../iahwc
-- cd -
-- os/linux/weston/build_script.sh --weston-dir=../weston --iahwc-dir=$(pwd) --build
```
harishkrupo: Doing this will disable the testing of weston and iahwc backend for weston.
The HACK line was written just as a hint that the weston's build system expects that iahwc is placed at ../iahwc relative to the build directory.
IIRC the build failure was due to compilation error in libva, which occurs much before this.

MicheleL-Intel:
@harishkrupo  Can you share the libva compile error?  I want to see if it's the same as the one I've observed recently.

stripes416:
@MicheleL-Intel if you click on 'details' to the right of travis build fail alert, you can view the job log and see what failed where.

csravelar:
@harishkrupo It actually seems to be the vaapi autogen right before the hack line

PR #422, opened 2018-08-05
Ensure last plane valid when calling IsVideoPlane

renchenglei:
This help fix the crash when running Antutu 7.x. If function
GetLastUsedOverlay didn't return a valid value, HWC would crash
with "null pointer dereference" when calling IsVideoPlane function

Jira: OAM-67193
Test: Antutu 7.x won't have HWC crash

### DIFF
--- a/common/display/displayplanemanager.cpp
+++ b/common/display/displayplanemanager.cpp
@@ -462,6 +462,8 @@ void DisplayPlaneManager::ValidateCursorLayer(
       }
 
       last_plane = GetLastUsedOverlay(composition);
+      if (last_plane == NULL)
+        break;
       bool reset_overlay = false;
       if (!last_plane->GetOffScreenTarget())
         reset_overlay = true;
@@ -498,7 +500,8 @@ void DisplayPlaneManager::ValidateCursorLayer(
       }
 
       last_plane = GetLastUsedOverlay(composition);
-      is_video = last_plane->IsVideoPlane();
+      if (last_plane)
+        is_video = last_plane->IsVideoPlane();
     }
 
     cursor_index++;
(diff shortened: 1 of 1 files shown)


js0701:
@renchenglei 
I think this NULL check need to be extended to all places  GetLastUsedOverlay is called?

renchenglei:
@js0701 
I think so. If we want to call IsVideoPlane, we should confirm last_plane is valid. If not, there will be "null pointer dereference" and HWC crashes.

js0701:
@renchenglei  Please do it in this PR to prevent later bugs

js0701:
@tpalli  we run klockwork. Can you specify which coverity tool?

tpalli:
@js0701 I mean Coverity by Synopsys (https://scan.coverity.com/), I would presume Klockwork to also detect usage of possibly NULL pointer though.

renchenglei:
Thanks @js0701 & @tpalli, I will take a look it.

kalyankondapally:
we do run coverity but is not automated in CI (which we should do).

common/display/displayplanemanager.cpp
```diff
@@ -498,7 +498,8 @@ void DisplayPlaneManager::ValidateCursorLayer(
       }
 
       last_plane = GetLastUsedOverlay(composition);
-      is_video = last_plane->IsVideoPlane();
+      if (last_plane)
```
kalyankondapally: While I understand the check, it doesn't make much sense that we are ending up with no last overlay unless for some reason we are just rendering cursor layer and nothing else. So, I guess something is broken earlier in the system, @renchenglei You mentioned about some under runs in the email, what's the fix for it ?

renchenglei: @kalyankondapally this is not for "under runs", it is fix for https://jira01.devtools.intel.com/browse/OAM-67193

renchenglei:
@js0701 @kalyankondapally @tpalli 
I have updated the PR, please help take a look. Thanks!

js0701:
LGTM

renchenglei:
@js0701 
Thanks Johnson, I have updated it!

js0701:
merged

PR #429, opened 2018-08-13
Temp PAVP Debug

harishkrupo:


harishkrupo:
@js0701 I have tested that the blending works using the first 3 commits. The fourth commit adds some state changes for handling the composition. Could you take a look at the patches?
There are a few places where I have marked with "reviewer". Those places I wasn't certain on how to handle, could you advise on that?

harishkrupo:
@kalyankondapally please review

common/compositor/va/varenderer.cpp
```diff
@@ -227,106 +228,127 @@ bool VARenderer::Draw(const MediaState& state, NativeSurface* surface) {
...
 
-  if (surface_region.width == 1920 && surface_region.height == 1080) {
-    // FIXME: WA for OAM-63127. Not sure why this is needed but seems
-    // to ensure we have consistent 60 fps.
-    vaSyncSurface(va_display_, surface_out);
   }
 
+  ret |= vaEndPicture(va_display_, va_context_);
+
+  // reviewer: How to resolve this?
```
js0701: OK, We can drop here. 
But when Android KPI is not meet, we should pick this

common/display/displayplanestate.cpp
```diff
@@ -469,9 +469,15 @@ bool DisplayPlaneState::HasCursorLayer() const {
 }
 
 bool DisplayPlaneState::IsVideoPlane() const {
-  return private_data_->type_ == DisplayPlanePrivateState::PlaneType::kVideo;
+  bool is_video = private_data_->type_ == DisplayPlanePrivateState::PlaneType::kVideo;
+  bool single_layer = private_data_->source_layers_.size() == 1;
+  return is_video && single_layer;
```
js0701: Does it mean that if single layer but not video layer, also a video plane?

common/compositor/va/varenderer.cpp
harishkrupo: Sure, will drop it.
BTW, this seems like a hack. Any idea how to properly resolve it?

js0701:
What means [TO BE REMOVED] in the commit messge?

harishkrupo:
@js0701 The above patch list is just for preliminary review (as an RFC). Few of the above commits is to ease the testing of the media backend and the state changes. These include enforcing gpu composition and forcing media composition for all layers. Once I get the test setup for protected content, I will test all these commits make the necessary changes and update the PR. The updated PR will not contain the [TO BE REMOVED] commits.

js0701:
@harishkrupo I am going to rebase and merge all PAVP related patches.
Can you rebase this PR to master latest?

harishkrupo:
Closing this.
Patches rebased on master, will open a new PR for the master branch.

PR #430, opened 2018-08-14
Add RBC enabled log.

YuanjunHuang:
This is for the RBC feature test case, providing information to
indicate if RBC is working.

Jira: https://jira01.devtools.intel.com/browse/OAM-65434
Tests: test on APL.

### DIFF
--- a/os/platformcommondrmdefines.cpp
+++ b/os/platformcommondrmdefines.cpp
@@ -79,6 +79,11 @@ int CreateFrameBuffer(
     ret = drmModeAddFB2WithModifiers(
         gpu_fd, iwidth, iheight, iframe_buffer_format, m_igem_handles,
         m_ipitches, m_ioffsets, modifiers, fb_id, DRM_MODE_FB_MODIFIERS);
+
+    if ((ret == 0) && ((modifier == I915_FORMAT_MOD_Y_TILED_CCS) ||
+        (modifier == I915_FORMAT_MOD_Yf_TILED_CCS))) {
+      ITRACE("RBC enabled. Create frame buffer with css modifier successfully.");
+    }
   } else {
     ret = drmModeAddFB2(gpu_fd, iwidth, iheight, iframe_buffer_format,
                         m_igem_handles, m_ipitches, m_ioffsets, fb_id, 0);


YuanjunHuang:
@js0701 PTAL.

os/platformcommondrmdefines.cpp
```diff
@@ -79,6 +79,10 @@ int CreateFrameBuffer(
     ret = drmModeAddFB2WithModifiers(
         gpu_fd, iwidth, iheight, iframe_buffer_format, m_igem_handles,
         m_ipitches, m_ioffsets, modifiers, fb_id, DRM_MODE_FB_MODIFIERS);
+
+    if (ret == 0) {
```
js0701: there are a few modificators that is not related with RBC as I know.
For example, Yf tile.
We need to check exactly the CCS modificator I think.

tpalli: IIRC Y-tiling is a requirement for RBC though so it is required and RBC will fail without it. What's the current set of modifiers in use?

YuanjunHuang: both I915_FORMAT_MOD_Y_TILED_CCS and I915_FORMAT_MOD_Yf_TILED_CCS modifier are possible. In current impl, when the plane initialize, property will be queried what modifier is supported and set it to be prefer modifier, with which later to allocate offscreen render target and frame buffer. If allocate fails, will add it into blacklist. so I think we dont have to worry about this.

js0701:
How about I915_FORMAT_MOD_Yf_TILED it is modificator, and not related with RBC?

os/platformcommondrmdefines.cpp
js0701: I915_FORMAT_MOD_Yf_TILED how about this one

YuanjunHuang: Only I915_FORMAT_MOD_Y_TILED_CCS and I915_FORMAT_MOD_Yf_TILED_CCS is applicable for RBC, for primary plane. I915_FORMAT_MOD_Yf_TILED is not taking account.

YuanjunHuang:
@js0701 update

js0701:
LGTM

js0701:
Closed as merged

PR #431, opened 2018-08-15
Interlace mode control

MicheleL-Intel:
GP2.0 Renderer interlaced video mode control

  Many interlaced video clips still exist despite the fact that
  most modern computer monitors do not support interlaced video.
  Provide the interface to deinterlace the video prior to rendering
  to enhance user experience. since this capability does not exist
  in Android and Linux OSes.  
    
  Jira: GSE-1618
  Test: To be tested on KabyLake NUC
    

js0701:
@MicheleL-Intel  It seems you need to rebase your PR.
Please cherry-pick your commit on top of master /latest

common/compositor/va/varenderer.cpp
```diff
@@ -156,6 +156,9 @@ bool VARenderer::GetVAProcDeinterlaceFlagFromVideo(HWCDeinterlaceFlag flag) {
   } else {
     // TODO:Need video buffer meta data to judge if the frame really need
     // Deinterlace.
+    if (OverlayBuffer::GetInterlaceMode() != kDeinterlaceNone) {
+      return true;
```
js0701: The logic is wrong here.
OverlayBuffer::GetInterlaceMode() shoud not be static function
We should able to track is interlaced fied regarding to each overlay buffer instance

js0701: It also can't be compiled

MicheleL-Intel:
@xuguangxin understood and thanks for sharing the deinterlace API document because it helps to clarify the interface.  

Can you please compare external/kernel-headers/original/uapi/drm/i915_drm.h with kernel/project-celadon/include/uapi/drm/i915_drm.h?   It appears those two files are out of sync and I'm wondering if that needs to be corrected.  I presume the former is for exposing the defines to external modules.

PR #432, opened 2018-08-20
Fix screen flicker when selecting option in setting menu

renchenglei:
This help fix screen flicker when selecting any option in the
setting menu screen with touch screen. We should return empty
rectangle when rmin <= lmax or bmin <= tmax.

Jira: OAM-67240
Test: Screen won't flicker when selecting option in setting menu


js0701:
Good. Thanks @renchenglei

js0701:
Closed as merged

PR #437, opened 2018-08-23
  GP2.0 Renderer interlaced video mode control

MicheleL-Intel:
Many interlaced video clips still exist despite the fact that
  most modern computer monitors do not support interlaced video.
  Since this capability does not currently exist in Android,
  provide the interface to deinterlace the video prior to
  rendering to enhance the user experience.

  Jira: GSE-1618
  Test: To be tested on KabyLake NUC

### DIFF
--- a/common/compositor/va/varenderer.cpp
+++ b/common/compositor/va/varenderer.cpp
@@ -150,23 +150,36 @@ bool VARenderer::SetVAProcFilterColorValue(HWCColorControl mode,
   }
 }
 
-bool VARenderer::GetVAProcDeinterlaceFlagFromVideo(HWCDeinterlaceFlag flag) {
+bool VARenderer::GetVAProcDeinterlaceFlagFromVideo(
+    const HWCDeinterlaceFlag flag, OverlayBuffer* buffer) {
+  if (buffer == NULL) {
+    ETRACE("Invalid buffer");
+    return false;
+  }
+
   if (flag != HWCDeinterlaceFlag::kDeinterlaceFlagAuto) {
     return false;
   } else {
-    // TODO:Need video buffer meta data to judge if the frame really need
-    // Deinterlace.
+    // Inspect metadata in current layer to determine if
+    // deinterlacing is required.
+    return buffer->GetInterlace();
   }
-  return false;
 }
-bool VARenderer::SetVAProcFilterDeinterlaceMode(
-    const HWCDeinterlaceProp& prop) {
+
+bool VARenderer::SetVAProcFilterDeinterlaceMode(const HWCDeinterlaceProp& prop,
+                                                OverlayBuffer* buffer) {
   VAProcDeinterlacingType mode;
+
+  if (buffer == NULL) {
+    ETRACE("Invalid buffer");
+    return false;
+  }
+
   if (prop.flag_ == HWCDeinterlaceFlag::kDeinterlaceFlagNone) {
     SetVAProcFilterDeinterlaceDefaultMode();
     return true;
   } else if (prop.flag_ == HWCDeinterlaceFlag::kDeinterlaceFlagForce ||
-             GetVAProcDeinterlaceFlagFromVideo(prop.flag_)) {
+             GetVAProcDeinterlaceFlagFromVideo(prop.flag_, buffer)) {
     switch (prop.mode_) {
       case HWCDeinterlaceControl::kDeinterlaceNone:
         mode = VAProcDeinterlacingNone;
@@ -280,7 +293,8 @@ bool VARenderer::Draw(const MediaState& state, NativeSurface* surface) {
   for (auto itr = state.colors_.begin(); itr != state.colors_.end(); itr++) {
     SetVAProcFilterColorValue(itr->first, itr->second);
   }
-  SetVAProcFilterDeinterlaceMode(state.deinterlace_);
+
+  SetVAProcFilterDeinterlaceMode(state.deinterlace_, buffer_in);
 
   if (!UpdateCaps()) {
     ETRACE("Failed to update capabailities. \n");
--- a/os/android/utils_android.h
+++ b/os/android/utils_android.h
@@ -251,6 +251,9 @@ static bool ImportGraphicsBuffer(HWCNativeHandle handle, int fd) {
   if (handle->meta_data_.format_ == DRM_FORMAT_YVU420_ANDROID)
     handle->meta_data_.format_ = DRM_FORMAT_YVU420;
 
+  // Store is_interlaced flag from Gralloc1
+  handle->meta_data_.is_interlaced_ = gr_handle->is_interlaced;
+
   return true;
 }
 #ifdef __cplusplus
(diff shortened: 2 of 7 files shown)


Android.common.mk
```diff
@@ -137,7 +137,7 @@ LOCAL_CPPFLAGS += \
 	-DENABLE_RBC
 endif
 
-ifneq ($(strip $(HWC_DISABLE_VA_DRIVER)), true)
```
js0701: No nessary change make file. On GP2.0 it DISABLE_VA_DRIVER is not defined.
Discard this change

Android.common.mk
```diff
@@ -137,7 +137,7 @@ LOCAL_CPPFLAGS += \
...
+ifneq ($(strip $(HWC_DISABLE_VA_DRIVER)), false)
```
js0701: Same obove

common/Android.mk
```diff
@@ -86,7 +86,7 @@ LOCAL_SRC_FILES := \
         utils/hwcutils.cpp \
```
js0701: Same above

common/Android.mk
```diff
@@ -86,7 +86,7 @@ LOCAL_SRC_FILES := \
...
         utils/disjoint_layers.cpp
 
-ifneq ($(strip $(HWC_DISABLE_VA_DRIVER)), true)
```
js0701: Same obove

common/compositor/va/varenderer.cpp
```diff
@@ -150,23 +150,40 @@ bool VARenderer::SetVAProcFilterColorValue(HWCColorControl mode,
...
+  }
+
   if (flag != HWCDeinterlaceFlag::kDeinterlaceFlagAuto) {
     return false;
   } else {
-    // TODO:Need video buffer meta data to judge if the frame really need
-    // Deinterlace.
+    // Inspect video buffer metadata in current layer to determine if
+    // this frame needs to be deinterlaced.
+    HWCDeinterlaceControl mode = layer->GetBuffer()->GetInterlaceMode();
```
js0701: I think we can change the function name to avoid confusion.
InterlacedMode is the interlace hahavior controller, which is defined in HWC service.
Here suggest to use IsInterlaced

common/compositor/va/varenderer.cpp
```diff
@@ -150,23 +150,40 @@ bool VARenderer::SetVAProcFilterColorValue(HWCColorControl mode,
...
+    DUMPTRACE("Interlace mode = (%d)", mode);
+    if (mode != HWCDeinterlaceControl::kDeinterlaceNone) {
```
js0701: Do not confuse IsInterlaced with Deinterlace controller. It is different. IsInterlaced is about the buffer itself, Deinterlace controller is tha Application intend

common/compositor/va/varenderer.cpp
```diff
@@ -280,7 +297,8 @@ bool VARenderer::Draw(const MediaState& state, NativeSurface* surface) {
   for (auto itr = state.colors_.begin(); itr != state.colors_.end(); itr++) {
     SetVAProcFilterColorValue(itr->first, itr->second);
   }
-  SetVAProcFilterDeinterlaceMode(state.deinterlace_);
+
+  SetVAProcFilterDeinterlaceMode(state.deinterlace_, state.layer_);
```
js0701: Suggest this function use OverlayBuffer as the second parameter, which is clearer

os/android/utils_android.h
```diff
@@ -251,6 +251,11 @@ static bool ImportGraphicsBuffer(HWCNativeHandle handle, int fd) {
   if (handle->meta_data_.format_ == DRM_FORMAT_YVU420_ANDROID)
     handle->meta_data_.format_ = DRM_FORMAT_YVU420;
 
+#ifdef USE_GRALLOC1
```
js0701: Why we need this USE_GRALLOC1 macro?

public/hwcbuffer.h
```diff
@@ -38,7 +37,10 @@ struct HwcBuffer {
   int prime_fds_[4];
   uint32_t num_planes_ = 0;
   uint32_t fb_modifiers_[8];
+  uint32_t is_interlaced_ = 0;
   hwcomposer::HWCLayerType usage_ = hwcomposer::kLayerNormal;
+  hwcomposer::HWCDeinterlaceControl interlace_mode_ =
```
js0701: As talked before, a buffer does not have interlaced_mode, but should have bool is_interlaced

js0701:
Still something is missing.
OverlayLayer::SetBuffer
this function should add SetInterlaced to drmBuffer
Because interlaced info can be changed while the buffer handle does not change

Android.common.mk
MicheleL-Intel: Ditto.

common/compositor/va/varenderer.cpp
MicheleL-Intel: The initial intent was to read the interlace mode from the active layer but it looks like it's not necessary.  I will change as suggested.

common/compositor/va/varenderer.cpp
MicheleL-Intel: Same comment as above: "The initial intent was to read the interlace mode from the active layer but it looks like it's not necessary."

common/compositor/va/varenderer.cpp
MicheleL-Intel: Agree this macro is not needed since HwcBuffer services both gralloc0 and gralloc1.

os/android/utils_android.h
MicheleL-Intel: Agree this macro is not needed since HwcBuffer services both gralloc0 and gralloc1.

public/hwcbuffer.h
MicheleL-Intel: I agree with your comment and had initially defined is_interlaced_ as bool since it requires only 8-bits.  However, the function you created, setInterlace() defines "uint32_t interlace" as the second parameter.  That's why it is now uint32_t, in order to be consistent with your setInterlace().  I'm changing to "bool is_interlaced" in HwcBuffer since that makes more sense but leaving setInterlace as is because I'm not aware of the reason behind it.

common/compositor/va/varenderer.cpp
MicheleL-Intel: Will do.

js0701:
Why deleting Android.common.mk

wsi/overlaybuffer.h
```diff
@@ -63,6 +63,10 @@ class OverlayBuffer {
 
   virtual uint32_t GetTilingMode() const = 0;
 
+  virtual bool GetInterlace() const = 0;
+
+  virtual void SetInterlace(bool isInterlaced) const = 0;
```
js0701: Why this function has to be const here

MicheleL-Intel: They should not be constant.   Thanks for the catch.

MicheleL-Intel:
@js0701 it's already done, just waiting for rebuild to complete.  Thanks.

js0701:
closed as merged

PR #438, opened 2018-08-24
Robustness for AaaG: when there is no pipe assigned,

js0701:
HWC should return false instead of crash

Jira: None
Tests: Android boot with no pipe assigned in AaaG


js0701:
Closed as merged

PR #444, opened 2018-08-24
TravisCI fix

stripes416:
Fully isolate travisCI builds

'sudo: required' fully seperates each build into their own, isolated & clean,
Google Compute Engine vm -- so no more shared state between builds.

Signed-off-by: Kelly Ledford <kelly.ledford@intel.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: cpp
 
-sudo: false
+sudo: required
 dist: trusty
 
 cache:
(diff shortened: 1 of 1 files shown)


stripes416:
I've applied this patch to all of my open PRs and it has fixed the travisCI issue we've been seeing.

The theory we're working with is that during the mesa portion of the builds, travis was reusing some internal state, and in some cases, failing, while passing in others.

MicheleL-Intel:
LGTM.

js0701:
LGTM

PR #446, opened 2018-08-27
  Fix for Antutu Regression

amuthuk1:
The issue here is canvas color was updated to kernel
  everytime during the display update eventhough there
  was no update in the color from the upper layer. Bcos,
  of this redundant ioctl calls to kernel, 3d scores
  regressed. The fix here is to call the ioctl only
  when there is a set/update in the canvas pipe color

  Signed-off-by: Kedar Karanje <kedar.j.karanje@intel.com>

  JIRA: https://jira01.devtools.intel.com/browse/OAM-65913


js0701:
LGTM. Thanks

amuthuk1:
I have incorporated the review comment provided. Changed Bcos to Because in the comment section of the gerrit

js0701:
closed as merged

PR #449, opened 2018-08-31
Update AliOS support

caiwenzh:
1. align with AliOS HAL and clean up the implementation
2. Fix compile issues on AliOS


wsi/drm/drmplane.cpp
```diff
@@ -514,10 +513,17 @@ bool DrmPlane::IsSupportedModifier(uint64_t modifier, uint32_t format) {
     const format_mods& obj = formats_modifiers_.at(i);
     if (obj.format == format) {
       std::vector<uint64_t>::const_iterator it;
+#ifdef OS_YUNOS
```
kalyankondapally: why is this alios specific ?

caiwenzh: AS YUNOS(alios) toolchain is very old there is compile error here.

kalyankondapally: hmm not sure why this is marked as resolved. I dont understand why this is YUNOS specific.

strassek: I'm a bit confused on this also, the auto keyword is newer than std::find... Maybe you just need to "#include <algorithm>" somewhere?

wsi/drm/drmplane.cpp
```diff
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 */
+#include <algorithm>
```
caiwenzh: add this to avoid compile error on AliOS.

wsi/drm/drmplane.cpp
caiwenzh: yes, after add “#include <algorithm>” can pass the build now. Thanks!

PR #453, opened 2018-09-19
Remove NON_BLOCKING flag when drm commit

js0701:
It will cause frame dropping in Android
It can be due to an issue of between AOSP and HWC
That this flag should be discard

Jira: ACI-5424
Tests: On Android, no frame droping with DialogTest.apk

### DIFF
--- a/wsi/Android.mk
+++ b/wsi/Android.mk
@@ -104,6 +104,8 @@ LOCAL_CPPFLAGS += \
         -DUSE_GL
 endif
 
+LOCAL_CPPFLAGS += -DUSE_BLOCKING_COMMIT
+
 LOCAL_C_INCLUDES += \
 	$(INTEL_MINIGBM)/cros_gralloc/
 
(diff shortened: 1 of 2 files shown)


kalyankondapally:
Not sure why you need this. Let's please fix it in right place than here.

js0701:
@kalyankondapally  As I mentioned, it is a issue between SurfaceFlinger and HWC synchronization.
We can fix the "janky frame" in this way.

kalyankondapally:
This is not a fix for me. Please figure out why we are dropping frames and see if we lack any synchronization else where.

js0701:
@kalyankondapally I checked the fences they are correct.
Google's framework may be designed in this way.
I am syncing with @txu4  too

js0701:
@kalyankondapally this "jankly frame" issue exist also for GP1.3.

js0701:
Referecing google's drm HWC
Google’s drm HWC https://android.googlesource.com/platform/external/drm_hwcompoer
It does not use this NONBLOCK flag. 
I think we better follow it

kalyankondapally:
As we discussed offline, Async flag is a supported option upstream.

tpalli:
This might cause performance regression. In any case, I think it would be interesting to measure this. We are using unblocking commit to be able to be faster. If we are not faster, then there is a problem somewhere else in the pipeline.

js0701:
@tpalli Yes it is interesting to me too.
Have are running PNP test. Will share the result.
However, with a few customer's application, the PR is the solution janky frame situation.

PR #454, opened 2018-09-20
Add support for SolidColor Type layer in HWC

js0701:
It is a layer without buffer handler,
it expected HWC can fill the layer with
color(0,0,0) and multiplay with an Alpha value

Jira: https://jir01.devtoosl.intel.com/browse/OAM-68857
Tests: SolidColor composition layer can be rendered


js0701:
@kalyankondapally  you may want to take a look

PR #457, opened 2018-09-28
Revert "Apply damage transformations in GetLayerDamage"

js0701:
Flicker is observed when start an APK after this patch.
for example use Car Launcher

This reverts commit 145bf9c5853cb5b0de0136a076d1a8b9b0b1c8dc.

Conflicts:
	common/core/hwclayer.cpp
	public/hwclayer.h
	public/hwcutils.h

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -621,6 +621,15 @@ void GpuDevice::DisableHDCPSessionForAllDisplays() {
   }
 }
 
+void GpuDevice::SetPAVPSessionStatus(bool enabled, uint32_t pavp_session_id,
+                                     uint32_t pavp_instance_id) {
+  size_t size = total_displays_.size();
+  for (size_t i = 0; i < size; i++) {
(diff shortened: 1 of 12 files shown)


js0701:
@harishkrupo 
I don't know why
It is better if you can root cause it

stripes416:
@js0701 Can you help me understand why these two commits are in the same PR? They seem to do quite different things and if they don't belong together, we should split them into separate PRs.

js0701:
@stripes416  I agree that should use 2 PRs for those commits
Any easy BKM that 2 PRs can be generated in the same time? Thanks

stripes416:
@js0701 You'll need to split the two commits onto two different branches and do a force push over this one with just commit 2a695dd44441cb475809a3096566ac17dd5ed88f (the one that matches this pr title) and then do a separate pr for commit b2d35e5147d549d827b7e76bb23eeb6c673b85b4.

js0701:
closed this only as the 1st commit merged will generate another one for the second

PR #466, opened 2018-10-15
Set HWC as DRM master when got exclusive lock of /vendor/hwc.lock.

shuangwan01:
This is to ensure HWC could take over the DRM master role when earlyEvs
exits and make the Android UI rendering correctly as expected.

Jira: OAM-70073
Tests: Boot into Android home screen

### DIFF
--- a/wsi/drm/drmdisplaymanager.cpp
+++ b/wsi/drm/drmdisplaymanager.cpp
@@ -445,6 +445,13 @@ void DrmDisplayManager::IgnoreUpdates() {
   }
 }
 
+void DrmDisplayManager::setDrmMaster() {
+  int ret = drmSetMaster(fd_);
+  if (ret) {
+    ETRACE("Failed to call drmSetMaster : %s", PRINTERROR());
+  }
+}
+
 void DrmDisplayManager::HandleLazyInitialization() {
   spin_lock_.lock();
   if (release_lock_) {
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -670,6 +670,8 @@ void GpuDevice::HandleRoutine() {
       ITRACE("Successfully grabbed the hwc lock.");
     }
 
+    display_manager_->setDrmMaster();
+
     close(lock_fd_);
     lock_fd_ = -1;
   }
--- a/wsi/displaymanager.h
+++ b/wsi/displaymanager.h
@@ -53,6 +53,8 @@ class DisplayManager {
   // manager until ForceRefresh is called.
   virtual void IgnoreUpdates() = 0;
 
+  virtual void setDrmMaster() = 0;
+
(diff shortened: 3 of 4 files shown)


js0701:
@shuangwan01  LGTM 
but seems some coding style problems in your PR

wsi/drm/drmdisplaymanager.cpp
```diff
@@ -445,6 +445,12 @@ void DrmDisplayManager::IgnoreUpdates() {
   }
 }
 
+void DrmDisplayManager::setDrmMaster() {
+
+  drmSetMaster(fd_);
```
harishkrupo: Here you are assuming that the call will succeed. What if it fails? What should HWC do in that case?

js0701: Good point. Maybe add some some error message here

harishkrupo: I have previously observed this happening. In such a case the whole rendering will fail. What should we do?
I believe the handlewait condition for the thread should be (get_exclusive_lock_on_file && drmSetMaster_succeded).

shuangwan01: @harishkrupo And for endless loop if get_exclusive_lock_on file not return, it's the default behavior currently. The flock is executed in blocking mode. The exceptional case found so far is HWC doesn't have file access privilege to /vendor/hwc.lock, the call will returns directly and I believe this case is a bug needs to be fixed in system configuration. For endless loop for drmSetMaster_, suggest not add this behavior since there is no corresponding condition to break the endless loop here.

harishkrupo: Sorry, let me re-phrase. What I meant was that when we fail to set master, we need to ensure that all the rendering is blocked (or we return with an assertion) as we will not be able to use the KMS apis without master privileges. Thus instead of each commit returning EPERM, we can ensure that android is notified of the failure (unless this behaviour is needed somewhere).
The other option would be to keep trying to get the privileges and block all rendering till we get it. This would ensure that we just don't wait for the lock file but also ensure that the other drm client app drops its master.

shuangwan01: @harishkrupo Understand your points and more appreciated the idea of second option. So how about we consider second one as an enhancement PR? Since we depend on this PR to be integrated for condition earlyEvs drop master and then HWC set itself as master for smooth bootsplash/early RVC -> Android UI boot transition.  This PR could cover this verified use case. For the exceptional case that drmSetMaster fails, we could submit another PR after do thorough testing locally. Thanks!

harishkrupo: Ok sure. LGTM.

js0701:
closed as merged

PR #471, opened 2018-10-17
Hdcp 2.2

harishkrupo:


stripes416:
@harishkrupo Your series failed Travis CI due to failing check_patch.py. Please run it locally, fix any errors, and resubmit for review.

js0701:
@harishkrupo  I am also submiiting PR for HDCP for Beta release.
may need your rebase.
however, I agree the basic logic in this PR

harishkrupo:
> clang-format-diff-3.9 is what I am using.

Okay will downgrade my tool and will test.
 
>@harishkrupo I am also submiiting PR for HDCP for Beta release.
may need your rebase.
however, I agree the basic logic in this PR

Sure will rebase once you merge it :)

stripes416:
Travis-Ci uses clang-format 3.8 I think.

harishkrupo:
hmm, there is a discrepancy between clang-format-diff 6.0 and 3.9. I have pushed the above patches checked with 3.9 as that is the version we use in travis.
Not sure whether to accept 6.0's output or 3.9's.

stripes416:
Yeah, you are right, it does use 3.9. The system that Travis is on runs Xenial 16.04, which uses 3.9. That is currently the recommend OS version for Celadon systems -- so we should stick with that version imo.

harishkrupo:
@js0701 @stripes416 Patches rebased. Requires another review.

wsi/drm/drmdisplay.cpp
```diff
@@ -178,26 +178,15 @@ bool DrmDisplay::ConnectDisplay(const drmModeModeInfo &mode_info,
     return false;
   }
 
+  uint32_t hdcp_id = 0, hdcp_cp_id = 0;
   int value = -1;
   GetDrmHDCPObjectProperty("Content Protection", connector, connector_props,
-                           &hdcp_id_prop_, &value);
+                           &hdcp_id, &value);
```
js0701: value is set by   GetDrmHDCPObjectProperty, but seems not used anywhere?

wsi/drm/drmdisplay.cpp
```diff
@@ -419,30 +407,150 @@ bool DrmDisplay::SetBroadcastRGB(const char *range_property) {
...
+  ScopedDrmObjectPropertyPtr cprops(drmModeObjectGetProperties(
+      gpu_fd_, connector_, DRM_MODE_OBJECT_CONNECTOR));
+  for (i = 0; i < cprops->count_props; i++) {
+    if (cprops->props[i] == hdcp_id_) {
+      val = cprops->prop_values[i];
+      break;
+    }
+  }
+
+  if ((current_state_ && val == 2) || (!current_state_ && val == 0))
```
js0701: Should we use MACRO instead of constant int

wsi/drm/drmdisplay.cpp
```diff
@@ -419,30 +407,150 @@ bool DrmDisplay::SetBroadcastRGB(const char *range_property) {
...
+    return true;
 
-  if (hdcp_id_prop_ <= 0) {
+  return false;
+}
+
+bool DrmDisplay::DrmHDCPThread::DrmConnectorSetupHDCP() {
```
js0701: Instead of doing this in HWC, HDCP Deamon may have similar logic,
I think we need to sync with the VPG CP team about this.
We should avoid duplidated logic

harishkrupo: Yes, I initially thought so too. But the compositor needs to be an independent entity. I believe contacting HDCP daemon would require a binder call? Then that would incur overhead. 
Also, HDCP daemon is available only on Android. What about the other OSes/distros we support?

stripes416: @harishkrupo  @js0701  Do you consider this issue resolved?

harishkrupo:
@js0701 @stripes416 

@js0701 Are you referring to [HDCP](https://github.com/intel/hdcp)?
I gave a quick glance through their code and found some issues (maybe?)
- When the property is set, they first try to set DRM master. I am not sure if that will be allowed because IAHWC doesn't provide the mechanism to drop the privileges. If that call fails they return EBUSY.
- They try to set the property 3 times which is okay but they don't wait for ~5 seconds before trying again. This time is mandated by the HDCP 2.2 spec.
- They don't have a polling thread to identify if the protection is down due to some failure.

I am not against using the Daemon. One upside of using it would be that we could completely remove the HDCP bits in IAHWC (SRM, protection, etc) and directly use that as they provide support for these. The downside would be that it will add another dependency to IAHWC.

js0701:
@harishkrupo 
Romli, Khairul in IOTG is owning HDCP daemon in Android.
We can sync with him for those issues

harishkrupo:
@js0701 Sure.

harishkrupo:
@js0701 I observed that the HDCP daemon runs a thread to continually monitor the HDCP status. However, they don't wait for ~15 seconds before ensuring that the connection is authenticated. I have pushed a patch removing all the thread related code and just simply set the connector property and wait for the said amount of time before declaring failure or success.

common/core/mosaicdisplay.cpp
```diff
@@ -504,12 +504,15 @@ bool MosaicDisplay::GetDisplayName(uint32_t *size, char *name) {
...
 }
 
-void MosaicDisplay::SetHDCPState(HWCContentProtection state,
+bool MosaicDisplay::SetHDCPState(HWCContentProtection state,
                                  HWCContentType content_type) {
   uint32_t size = physical_displays_.size();
+  bool ret = true;
   for (uint32_t i = 0; i < size; i++) {
-    physical_displays_.at(i)->SetHDCPState(state, content_type);
+    ret = ret && physical_displays_.at(i)->SetHDCPState(state, content_type);
```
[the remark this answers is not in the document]
harishkrupo: Agreed. will change it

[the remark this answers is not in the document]
harishkrupo: I guess similar thing applies here too?
https://github.com/intel/IA-Hardware-Composer/pull/471/files#diff-a831b063d6a07594bea36549d019ecadR231

wsi/drm/drmdisplay.cpp
```diff
@@ -419,30 +408,81 @@ bool DrmDisplay::SetBroadcastRGB(const char *range_property) {
...
+    ETRACE("Failed to commit HDCP ret %d errno %d : %s\n", ret, errno,
+           PRINTERROR());
+    return false;
   }
 
-  drmModeConnectorSetProperty(gpu_fd_, connector_, hdcp_id_prop_, value);
-  ETRACE("Ignored Content type. \n");
+  /* Wait for Enable should be 5.1 Sec * 3(kernel reattempt cnt) */
+  for (uint8_t i = 0; i < 160; i++) {
+    if (CheckHDCPStatus()) {
```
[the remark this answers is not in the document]
js0701: Please hold on merge. This patch can impact HDCP integration with HDCP daemon on Android

stripes416:
Wait for @js0701 to merge (note to self!)

js0701:
@harishkrupo 
Current the no checking status mechanism is working well.
We had checked with VPG CP team and IOTG that HWC should only do simple mode settings.
Instead of anything about HDCP protocal itself. The protocal implementatinon should be through
daemon and kernel

harishkrupo:
@js0701 I agree that it is working correctly. It works for me too without any waits. But we shouldn't assume about the type of the connector/sink. We must wait for the mandated time. If not here then it should be done at the daemon. 
I didn't see the wait in the daemon so I added the wait here. If I am to remove this, then it is a fair expectation that a corresponding change must be done on the Daemon side.

wsi/drm/drmdisplay.cpp
```diff
@@ -419,30 +408,81 @@ bool DrmDisplay::SetBroadcastRGB(const char *range_property) {
...
+      gpu_fd_, connector_, DRM_MODE_OBJECT_CONNECTOR));
+  for (i = 0; i < cprops->count_props; i++) {
+    if (cprops->props[i] == hdcp_id_prop_) {
+      val = cprops->prop_values[i];
+      break;
+    }
+  }
+
+  if ((current_protection_support_ && val == HWCContentProtection::kEnabled) ||
+      (!current_protection_support_ && val == HWCContentProtection::kDesired))
```
YuanjunHuang: when and how the status will change from kDesired to kEnabled? Is it possible in the sleep time frame it is already changed to kEnabled, you can not get the kDesired state.

harishkrupo: sorry, could you please explain more about the sleep time frame?

YuanjunHuang: I mean in the SetHDCPState() there is  'usleep(100 * 1000);' for every iteration. is it possible in this time frame, the state changes from kUnDisired->kDesired->kEnabled? for such case, your logic in Check state will not return true.

js0701:
@kromli 
Can you check HDCP daemon has wait logic or not? As we talked in the email, the HWC should not worry about HDCP protocal itself
However, we still have the consern @harishkrupo  raised here

wsi/drm/drmdisplay.cpp
```diff
@@ -419,30 +408,81 @@ bool DrmDisplay::SetBroadcastRGB(const char *range_property) {
...
+                            NULL);
+  if (ret) {
+    ETRACE("Failed to commit HDCP ret %d errno %d : %s\n", ret, errno,
+           PRINTERROR());
+    return false;
   }
 
-  drmModeConnectorSetProperty(gpu_fd_, connector_, hdcp_id_prop_, value);
-  ETRACE("Ignored Content type. \n");
+  /* Wait for Enable should be 5.1 Sec * 3(kernel reattempt cnt) */
```
YuanjunHuang: possible to wait up to 15s  blocking hwc service such long time seems not acceptable. I would suggest not blocking in hwc.

harishkrupo: Yes, agreed. This should be done by the hdcp daemon. I had previously raised it as a concern.
Once the daemon does it, We can remove this here.

PR #475, opened 2018-10-24
Fixed the crash due to layer->GetNativeHandle returns a null pointer.

shuangwan01:
The fix is to skip the layer handling if the layer buffer is NULL.

Jira: OAM-70263
Tests: HWC doesn't crash if handling layer with no buffer attached.

### DIFF
--- a/common/display/virtualdisplay.cpp
+++ b/common/display/virtualdisplay.cpp
@@ -198,6 +198,13 @@ bool VirtualDisplay::Present(std::vector<HwcLayer *> &source_layers,
 
       const HwcRect<int> &display_frame = layer->GetDisplayFrame();
       sf_handle = layer->GetNativeHandle();
+
+      if (NULL == sf_handle) {
+        ITRACE("Skip layer index: %u for Hyper DMA buffer sharing",
+               layer_index);
+        continue;
+      }
+
       std::shared_ptr<OverlayBuffer> buffer(NULL);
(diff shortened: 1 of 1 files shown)


js0701:
LGTM

common/display/virtualdisplay.cpp
```diff
@@ -198,6 +198,13 @@ bool VirtualDisplay::Present(std::vector<HwcLayer *> &source_layers,
...
       sf_handle = layer->GetNativeHandle();
+
+      if (NULL == sf_handle) {
+        ETRACE("layer->GetNativeHandle returns NULL, skip layer_index: %u.",
+               layer_index);
+        continue;
+      }
+
       std::shared_ptr<OverlayBuffer> buffer(NULL);
       uint32_t gpu_fd = resource_manager_->GetNativeBufferHandler()->GetFd();
```
js0701: @shuangwan01  Please use ITRACE here since it is an error in HWC. the log here is just for information. The log should be "... skip layer_index: %u for hyper dma buffer sharing"

stripes416: Once these changes are made, this should be good to merge.

shuangwan01:
Thanks all for comments, the warning message has been revised accordingly.

stripes416:
Thanks for updating @shuangwan01!

PR #476, opened 2018-10-25
SolidColor type: as the layer is a render target of GPU, we should still

js0701:
set the release fence of it. SurfaceFlinger need the fence to sync.
The acquire fence is not necessary since the layer has no buffer at all

Jira: https://jira01.devtools.intel.com/browse/OAM-70347
Tests: Android UI normal and no fd leak

### DIFF
--- a/common/core/hwclayer.cpp
+++ b/common/core/hwclayer.cpp
@@ -178,14 +178,18 @@ void HwcLayer::SetReleaseFence(int32_t fd) {
 }
 
 int32_t HwcLayer::GetReleaseFence() {
-  if (!sf_handle_)
-    return -1;
   int32_t old_fd = release_fd_;
   release_fd_ = -1;
   return old_fd;
 }
 
 void HwcLayer::SetAcquireFence(int32_t fd) {
+  if (!sf_handle_) {
+    if (fd > 0) {
+      close(fd);
+    }
+    acquire_fence_ = -1;
+  }
   if (acquire_fence_ > 0) {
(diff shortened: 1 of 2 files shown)


MicheleL-Intel:
I general, I think you may need to take a look at possible synchronization issues in the accessor functions.  For example, there may be some delay before the fd is released or acquired.  The code assumes the file descriptors are already in the proper state, which may not be the case.

Also, the title of this PR is either incomplete or does not match the description of the patch.

common/core/hwclayer.cpp
```diff
@@ -178,14 +178,18 @@ void HwcLayer::SetReleaseFence(int32_t fd) {
...
 }
 
 void HwcLayer::SetAcquireFence(int32_t fd) {
+  if (!sf_handle_) {
+    if (fd > 0) {
+      close(fd);
+    }
+    acquire_fence_ = -1;
+  }
   if (acquire_fence_ > 0) {
```
MicheleL-Intel: In GetAcquireFence(), if sf_handle_ is invalid, does it need to close fd before exiting?

int32_t HwcLayer::GetAcquireFence() {
  if (!sf_handle_) {
    if (fd > 0) {
      close(fd);
    }
    return -1;
}

js0701: no need

js0701:
@stripes416 
Thanks for the comment. Updated

js0701:
Close as merged

PR #480, opened 2018-10-29
Add support for the Color Depth control

Kishore409:
By default the Kernel sets the bpc for a Pipe based on EDID
while considering the range of bpc supported by the pipe .
In this patch, Userspace requests the kernel for a specific bpc
based on a particular resolution.

Jira: None
Tests: Android UI starts normally and bpc request to kernel goes fine
       leaving the kernel to decide for the pipe.


Kishore409:
@stripes416 I will fix and update

PR #481, opened 2018-10-30
Add support for the Color Depth control

Kishore409:
By default the Kernel sets the bpc for a Pipe based on EDID
while considering the range of bpc supported by the pipe .
In this patch, Userspace requests the kernel for a specific bpc
based on a particular resolution.

Jira: None
Tests: Android UI starts normally and bpc request to kernel goes fine
       leaving the kernel to decide for the pipe.

### DIFF
--- a/wsi/drm/drmdisplay.cpp
+++ b/wsi/drm/drmdisplay.cpp
@@ -132,6 +132,10 @@ void DrmDisplay::DrmConnectorGetDCIP3Support(
   }
 
   edid = (uint8_t *)blob->data;
+  if (!edid) {
+    return;
+  }
+
   blocks = FindExtendedBlocksForTag(edid, CTA_EXTENDED_TAG_CODE);
 
   for (uint8_t *ext_block : blocks) {
--- a/common/core/overlaylayer.cpp
+++ b/common/core/overlaylayer.cpp
@@ -416,8 +416,8 @@ void OverlayLayer::ValidatePreviousFrameState(OverlayLayer* rhs,
     if (!rhs->imported_buffer_.get()) {
       state_ |= kNeedsReValidation;
       return;
-    } else if (buffer->GetFormat() !=
-               rhs->imported_buffer_->buffer_->GetFormat()) {
+    } else if (buffer && (buffer->GetFormat() !=
+                          rhs->imported_buffer_->buffer_->GetFormat())) {
       state_ |= kNeedsReValidation;
       return;
     }


wsi/drm/drmdisplay.cpp
```diff
@@ -205,12 +205,15 @@ bool DrmDisplay::ConnectDisplay(const drmModeModeInfo &mode_info,
...
   GetDrmObjectProperty("Broadcast RGB", connector_props, &broadcastrgb_id_);
   GetDrmObjectProperty("DPMS", connector_props, &dpms_prop_);
+  GetDrmObjectProperty("max bpc", connector_props, &max_bpc_prop_);
 
   DrmConnectorGetDCIP3Support(connector_props);
-  if (dcip3_)
+  if (dcip3_) {
     ITRACE("DCIP3 support available");
-  else
+    SetPipeMaxBpc(PIPE_BPC_TWELVE);
```
js0701: Why it is 12, not 16 here

Kishore409: On HDMI, the BPC that a pipe support is ranging from 8 to 12
So the maximum it can be set is 12

js0701: OK. How about DP? and other interfaces
What should be the result if we leave things unset?

Kishore409: Even DP also ranges from 6 to 12(max)

js0701: bpc here means byte per component, right? If it strange that userpace should the max value of it, as we are using a fixed constant here. If hardware limitation, kernel should be able to set it automatically?

Kishore409: here the property "max bpc" is used to set the bpp of a pipe.
Not always the Kernel sets the Max bpp for a pipe.
But in case of HDR, the user space ensure toe set max bpp.
Hope I am clear here

MicheleL-Intel:
How are the alpha bits stored?

stripes416:
@Kishore409 just for future reference, when changes need to be made to your patch series, you can do a force push from your local branch to your github branch and it will automatically update an existing PR from that branch. You don't have to close the PR and reopen a new one. Its not something that everyone knows, so I thought I could save you some trouble the next time!

Kishore409:
> @Kishore409 just for future reference, when changes need to be made to your patch series, you can do a force push from your local branch to your github branch and it will automatically update an existing PR from that branch. You don't have to close the PR and reopen a new one. Its not something that everyone knows, so I thought I could save you some trouble the next time!

I realized later the older one where you commented got closed. Sure I will do force push as mentioned

Kishore409:
> How are the alpha bits stored?

Connector's don't have Alpha bits

wsi/drm/drmdisplay.h
```diff
@@ -160,6 +171,7 @@ class DrmDisplay : public PhysicalDisplay {
   uint32_t canvas_color_prop_ = 0;
   uint32_t connector_ = 0;
   bool dcip3_ = false;
+  uint32_t max_bpc_prop_ = 0;
```
MicheleL-Intel: Does this need to be an "uint32_t"?  "uint8_t" is sufficient don't you think?

wsi/drm/drmdisplay.cpp
```diff
@@ -205,12 +205,15 @@ bool DrmDisplay::ConnectDisplay(const drmModeModeInfo &mode_info,
   GetDrmObjectProperty("CRTC_ID", connector_props, &crtc_prop_);
   GetDrmObjectProperty("Broadcast RGB", connector_props, &broadcastrgb_id_);
   GetDrmObjectProperty("DPMS", connector_props, &dpms_prop_);
+  GetDrmObjectProperty("max bpc", connector_props, &max_bpc_prop_);
```
MicheleL-Intel: [MINOR] For consistency, suggest "max_bpc" to be in all caps.

wsi/drm/drmdisplay.cpp
```diff
@@ -778,6 +781,14 @@ void DrmDisplay::SetPipeCanvasColor(uint16_t bpc, uint16_t red, uint16_t green,
                            canvas_color_prop_, canvas_color);
 }
 
+void DrmDisplay::SetPipeMaxBpc(uint16_t max_bpc) const {
+  if (max_bpc_prop_ == 0)
+    return;
+
+  drmModeConnectorSetProperty(gpu_fd_, connector_, max_bpc_prop_,
```
MicheleL-Intel: drmModeConnectorSetProperty() does not check to ensure gpu_fd_ is valid which can be a problem if gpu_fd_ were to be invalid.

For now, check gpu_fd_ here or do it the correct way by checking gpu_fd_ in drmModeConnectorSetProperty().

wsi/drm/drmdisplay.cpp
Kishore409: We can't change this , the convention is as per Kernel implementation

wsi/drm/drmdisplay.h
Kishore409: Probably, but here I choose earlier implementations..

wsi/drm/drmdisplay.cpp
Kishore409: I think the fd check is taken care in libdrm.

MicheleL-Intel: Hopefully so.  Can you please confirm?  This is important for stability.

harishkrupo: The gpu_fd_ is set during initialization of drmdisplay, which typically would never be wrong.
My suggestion would be to return bool instead of void and check the return value from drmModeConnectorSetProperty instead of assuming that it would always succeed.

MicheleL-Intel: @harishkrupo That sounds proper.

wsi/drm/drmdisplay.cpp
harishkrupo: > @harishkrupo I do have some comments regarding this patch:
> 
> 1. The purpose of the patch is to crop the max_bpc which in essence will reduce the color gamut but it is a workaround for faulty dongles.  Upon setting the max_bpc, I don't see how these two variables are getting updated:
>    ```
>       old_connector_state->max_requested_bpc
>       new_connector_state->max_requested_bpc
>    ```

@MicheleL-Intel The patch that I shared is a bit old. The latest atm is v13:
https://patchwork.freedesktop.org/patch/256122/

IIRC, the new state would be set here:
```
diff --git a/drivers/gpu/drm/drm_atomic_uapi.c b/drivers/gpu/drm/drm_atomic_uapi.c
index a22d6f269b07..7492819c81ce 100644
--- a/drivers/gpu/drm/drm_atomic_uapi.c
+++ b/drivers/gpu/drm/drm_atomic_uapi.c
@@ -761,6 +761,8 @@  static int drm_atomic_connector_set_property(struct drm_connector *connector,
 
 		return set_out_fence_for_connector(state->state, connector,
 						   fence_ptr);
+	} else if (property == connector->max_bpc_property) {
+		state->max_requested_bpc = val;
 	} else if (connector->funcs->atomic_set_property) {
 		return connector->funcs->atomic_set_property(connector,
 				state, property, val);
```
Then it would be compared with the old state (i.e. the current state before we finish the atomic commit)

> 2. Just the basics ... there are unchecked pointers not in the patch but elsewhere.  For example, local variables created in drm_atomic_check_only() are not validated before they are passed into drm_atomic_plane_check() and drm_atomic_connector_check().

I am not sure about this. Probably the same reason we don't check gpu_fd_ everywhere we use it. You could reach out to the ML if you feel that the checks are required :)

stripes416:
@Kishore409 @js0701 Is this good to go or are we waiting for something? Also, @Kishore409 don't forget to rebase on top of latest master to enable an updated run through Travis.

js0701:
@harishkrupo please comment is this look good for you or not

Kishore409:
@js0701  please reopen the PR

PR #484, opened 2018-11-12
Separate disable_overlay and disable_explictsync

js0701:
disable_overlay means do not use HW overly.
It is beneficial for overlay debug.
disable_explictsync is Android framework feature about fence synchronization

Tracked-On: None
Tests: Compliation on Android pass


js0701:
@stripes416  done

PR #487, opened 2018-11-14
Do not need change size from source crop to display rect

yhe39:
When source crop begin from (0, 0), the source crop size
change will cause translated damage region doesn't correct.

Change-Id: Iae3143ac9be38b35c1e6649e7721469e6f39a790
Tracked-On: https://jira01.devtools.intel.com/browse/OAM-71102
Tests: AutoBots test don't have flick/half screen, the HAVC working ok.

### DIFF
--- a/common/core/hwclayer.cpp
+++ b/common/core/hwclayer.cpp
@@ -211,24 +211,14 @@ void HwcLayer::SufaceDamageTransfrom() {
   HwcRect<int> translated_damage =
       TranslateRect(surface_damage_, -source_crop_.left, -source_crop_.top);
 
-  int display_width = display_frame_.right - display_frame_.left;
-  int display_height = display_frame_.bottom - display_frame_.top;
-  int source_width = source_crop_.right - source_crop_.left;
-  int source_height = source_crop_.bottom - source_crop_.top;
+  // From observation: In Android, when the source crop coordinate
+  // is (0, 0), the surface damage is already translated to global display
+  // coordinate. Therefore, no translation is needed.
+  // The rotation scenario is not verified as the rotation is not supported on P
+  // Leave the rotation scenario code here temporary.
 
-  // From observation: In Android, when the source crop doesn't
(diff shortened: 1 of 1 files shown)


yhe39:
@js0701 @jennycao , please help to review. Thanks!

yhe39:
Thanks @js0701 , please help to check again.

js0701:
Missing Change-Id

common/core/hwclayer.cpp
```diff
@@ -211,24 +211,11 @@ void HwcLayer::SufaceDamageTransfrom() {
   HwcRect<int> translated_damage =
       TranslateRect(surface_damage_, -source_crop_.left, -source_crop_.top);
 
-  int display_width = display_frame_.right - display_frame_.left;
-  int display_height = display_frame_.bottom - display_frame_.top;
-  int source_width = source_crop_.right - source_crop_.left;
-  int source_height = source_crop_.bottom - source_crop_.top;
-
   // From observation: In Android, when the source crop doesn't
```
MicheleL-Intel: The proposed code change does not seem to match the comment.  If I understand the change correctly, if the surface_damage_ is populated with data, and the source_crop coordinate is (0,0), no translation is needed.  Therefore, should the preceding comment be changed to the following?

  // From observation: In Android, when the source crop coordinate
  // is (0, 0), the surface damage is already translated to global display
  // coordinate.  Therefore, no translation is needed.

yhe39:
Thanks @js0701 , please help to check!

common/core/hwclayer.cpp
yhe39: Hi @MicheleL-Intel, you are right, i will modify the comment according. 
Besides, the rotation situation is not verified so far as the rotation is not supported on P. Just leave the rotation situation here temporary. 
Thanks for your advise.

js0701:
closed as merged

PR #488, opened 2018-11-14
SetProtected: fix logic error

js0701:
Should only allow video and protected layer switch

Change-Id: I4633725277e0cee125c24b33a42fcd481f9f5990
Tracked-On: None
Tests: Video playback on Android

### DIFF
--- a/common/core/overlaylayer.h
+++ b/common/core/overlaylayer.h
@@ -185,9 +185,9 @@ struct OverlayLayer {
   }
 
   void SetProtected(bool isProtected) {
-    if (isProtected && type_ == kLayerVideo)
+    if (isProtected && (type_ == kLayerVideo || type_ == kLayerProtected))
       type_ = kLayerProtected;
-    else if (!isProtected && type_ == kLayerProtected)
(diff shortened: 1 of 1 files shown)


js0701:
@xuguangxin

common/core/overlaylayer.h
```diff
@@ -185,9 +185,9 @@ struct OverlayLayer {
   }
 
   void SetProtected(bool isProtected) {
-    if (isProtected && type_ == kLayerVideo)
+    if (isProtected && (type_ == kLayerVideo || type_ == kLayerProtected))
```
YuanjunHuang: is the 'protected' only for video layer? If so kLayerProtectedVideo may be better than kLayerProtected.

js0701: yes, it is only for video now.

js0701: @YuanjunHuang 
Let's keep this PR just for logic fix not anything else.
If you propose better name or refactory, you can give another PR

YuanjunHuang: ok sure.

stripes416:
closed as merged.

PR #492, opened 2018-11-26
Fix two marjor issues blocking PAVP playback in HWC

js0701:
1) When video is on top layers, there is no ganrantee that video get a separate plane.
Video can be assign to be composited. Add logic to squash non-video planes to make video
to get a seperate plane
2) When layers are scanout, or be processed as video plane, there is no necessary to
getGPUResources from them. It is a reductant operation and may result in error then video is
protected

Change-Id: I379ce5ecf3944230580734ac8317e6c6e3f7f8e1
Tracked-On: https://jira01.devtools.intel.com/browse/OAM-71887


js0701:
Closed as merged

PR #498, opened 2018-11-28
Do not need change surface damage rect

yhe39:


common/core/hwclayer.cpp
```diff
@@ -215,6 +215,11 @@ void HwcLayer::SufaceDamageTransfrom() {
   HwcRect<int> translated_damage =
       TranslateRect(surface_damage_, -source_crop_.left, -source_crop_.top);
 
+// On GP2.0 the transform is not necessary as the Frame work translate already.
```
js0701: Instead of use this maro inside SurfaceDamageTransform,
You should not call this function at all
This SufaceDamageTransform function should be include in SURFACE_DAMAGE_TRANSFORM macro

yhe39: hi @js0701, if include the SufaceDamageTransform  in SURFACE_DAMAGE_TRANSFORM  macro, will has compile error, as the SurfaceDamageTransfrom function is called in Validate function.

yhe39: Include the SufaceDamageTransform in caller function in macro too.

yhe39:
@js0701 ,please help to check. Thanks!

js0701:
don't we need to change header file?

js0701:
Closed as merged.

PR #500, opened 2018-11-30
Fix Alpha calculation error

Shao-Feng:
The Alpha of layer should be calculated with texture Alpha alone.
Instead of plus with RGB.

Change-Id: I848e0f1060a87e1a2b54f07a4f353cf0b4e453b4
Tracked-On: https://jira01.devtools.intel.com/browse/OAM-72398
Tests: Compile sucessful for Android. Check the issue is fixed


js0701:
Closed as merged

PR #506, opened 2018-12-10
Fixed the HWC crash caused by map iterator erase.

shuangwan01:
Erase map iterator during iterating a map caused the crash.
The fix is to avoid erase itator in the interating loop and clear
the iterator after the loop end.

Tracked-On: https://jira01.devtools.intel.com/browse/OAM-71721
Tests: Trigger VirtualDisplay destructor condition and no HWC crash.


shuangwan01:
Just updated the patch commit message with full issue URL included. Thanks!

shuangwan01:
Okay, the change-id is added into commit message.

js0701:
Closed as merged

PR #509, opened 2018-12-13
optimize the 1st display frame commit 

dyang23:
After the earlyEvs exit, the HWC will get the hwc.lock
and begin rendering, but the initialize for the
1st frame will use about 150ms, this patch will let
the initialize done before hwc.lock.

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -649,7 +649,9 @@ void GpuDevice::HandleRoutine() {
       ITRACE("Successfully grabbed the hwc lock.");
     }
 
+	ETRACE("hwc before  set master.");
     display_manager_->setDrmMaster();
+	ETRACE("hwc after  set master.");
 
     close(lock_fd_);
     lock_fd_ = -1;
--- a/wsi/drm/drmdisplay.cpp
+++ b/wsi/drm/drmdisplay.cpp
@@ -37,6 +37,7 @@
 #define CTA_EXTENSION_TAG 0x02
 #define CTA_EXTENDED_TAG_CODE 0x07
 #define CTA_COLORIMETRY_CODE 0x05
+int FirstFrame=1;
 
 namespace hwcomposer {
 
@@ -590,6 +591,13 @@ bool DrmDisplay::CommitFrame(
   if (ret) {
     ETRACE("Failed to commit pset ret=%s\n", PRINTERROR());
     return false;
+  }
+  else
+  {
+    if (FirstFrame) {
+      FirstFrame = 0;
+	  ETRACE(" hwc first frame commited \n");
+    }
   }
 
   return true;
--- a/wsi/drm/drmdisplaymanager.cpp
+++ b/wsi/drm/drmdisplaymanager.cpp
@@ -365,9 +365,7 @@ void DrmDisplayManager::NotifyClientsOfDisplayChangeStatus() {
 
   for (auto &display : displays_) {
     display->NotifyDisplayWA(disable_last_plane_usage);
-    if (!ignore_updates_) {
-      display->ForceRefresh();
-    }
+    display->ForceRefresh();
   }
 
   for (auto &display : displays_) {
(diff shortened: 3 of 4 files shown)


common/core/gpudevice.cpp
```diff
@@ -649,7 +649,9 @@ void GpuDevice::HandleRoutine() {
       ITRACE("Successfully grabbed the hwc lock.");
     }
 
+	ETRACE("hwc before  set master.");
```
YuanjunHuang: use ITRACE instead of ETRACE if it is not error message

wsi/drm/drmdisplay.cpp
```diff
@@ -591,6 +592,13 @@ bool DrmDisplay::CommitFrame(
     ETRACE("Failed to commit pset ret=%s\n", PRINTERROR());
     return false;
   }
+  else
+  {
```
YuanjunHuang: not to turn to next line for left {

common/core/gpudevice.cpp
js0701: Suggest just to remove it

common/core/gpudevice.cpp
```diff
@@ -649,7 +649,9 @@ void GpuDevice::HandleRoutine() {
...
     display_manager_->setDrmMaster();
+	ETRACE("hwc after  set master.");
```
js0701: The same as above, it is debug message.

wsi/drm/drmdisplaymanager.cpp
```diff
@@ -446,10 +444,14 @@ void DrmDisplayManager::IgnoreUpdates() {
 }
 
 void DrmDisplayManager::setDrmMaster() {
-  int ret = drmSetMaster(fd_);
-  if (ret) {
-    ETRACE("Failed to call drmSetMaster : %s", PRINTERROR());
-  }
+    int ret = drmSetMaster(fd_);
+    while (ret) {
+        usleep(10000);
```
js0701: Why usleep(10000)
Can we avoid to do so

wsi/drm/drmdisplay.cpp
```diff
@@ -591,6 +592,13 @@ bool DrmDisplay::CommitFrame(
...
+    if (FirstFrame) {
+      FirstFrame = 0;
+	  ETRACE(" hwc first frame commited \n");
```
YuanjunHuang: ITRACE

common/core/gpudevice.cpp
dyang23: Done

wsi/drm/drmdisplay.cpp
dyang23: Done

wsi/drm/drmdisplay.cpp
```diff
@@ -37,6 +37,7 @@
 #define CTA_EXTENSION_TAG 0x02
 #define CTA_EXTENDED_TAG_CODE 0x07
 #define CTA_COLORIMETRY_CODE 0x05
+int FirstFrame=1;
```
YuanjunHuang: what is this used for? just see it set to 0 and print the log.

common/core/gpudevice.cpp
dyang23: since we are do the seamless translate with other team, this log can provide the time which we get the master. so we can keep it for a short time, and remove it later, once the task force done.

wsi/drm/drmdisplay.cpp
dyang23: only the first successful commit will be logged, used to check the 1st frame commit time.
This can be remove later, once the seamless work done.

wsi/drm/drmdisplay.cpp
dyang23: Done

wsi/drm/drmdisplaymanager.cpp
dyang23: If the earlyEvs can drop the master successfully, it's no need for HWC to get the master again, but during my debug process, I saw due to the permission issue, the earlyEvs may drop the master failed, and the HWC failed to get the master, so the display screen will keep black forever.  
The master is critical for HWC, and this logic is a protection,  in case some error in earlyEvs or system, and HWC cannot get master currently, but may get it later, so it won't show black screen forever.

js0701:
@dyang23  this should be closed right? Please close

PR #510, opened 2018-12-13
HWC: optimize the 1st display frame commit

dyang23:
After the earlyEvs exit, the HWC will get the hwc.lock
and begin rendering, but the initialize for the
1st frame will take ~150ms, this patch will let
the initialize done before taken hwc.lock.

### DIFF
--- a/common/display/displayqueue.cpp
+++ b/common/display/displayqueue.cpp
@@ -919,9 +919,10 @@ bool DisplayQueue::QueueUpdate(std::vector<HwcLayer*>& source_layers,
 
   int32_t fence = 0;
   bool fence_released = false;
-  composition_passed = display_->Commit(
-      current_composition_planes, previous_plane_state_, disable_explictsync,
-      kms_fence_, &fence, &fence_released);
+  if (!IsIgnoreUpdates())
+    composition_passed = display_->Commit(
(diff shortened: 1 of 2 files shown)


stripes416:
I'm confused here ... these are the same changes from PR #509? One of these should be closed to avoid confusion. Just a reminder: you can just force push to the same branch with changes and it will update the open PR associated with it (instead of opening a new one).

js0701:
@dyang23 
Please refer to other commits about the commit message. We require defined format for it

wsi/drm/drmdisplay.cpp
```diff
@@ -585,11 +585,16 @@ bool DrmDisplay::CommitFrame(
     *previous_fence_released = true;
   }
 #endif
-
+  static int FirstFrame = 1;
```
js0701: static bool isFirstFrame = true?

js0701: Actually this firstFrame logic is fore debug right
We should remove it for cleancode

dyang23: Done, removed the FirstFrame indicator.

js0701:
@dyang23 
Please update commit message, following convention of other commits

js0701:
Closed as merged. Thanks for working on this

PR #512, opened 2018-12-20
Zoom in or out the rendering damage for different display

Shao-Feng:
Add the "zoom in and out" code back for connecting 2 display with different
resolution. In the bug case, 2K and 1080P display are connected. the Rect
should be enlarged for 2K when the rect is calculated with 1080P
coordinates.

Change-Id: I60cdd681895e8fa4907a89c2450c6da410ba208b
Tests: Verified on Celadon with 2 displays and mosaic on GP
Tracked-On: https://jira.devtools.intel.com/browse/OAM-73581

### DIFF
--- a/common/core/hwclayer.cpp
+++ b/common/core/hwclayer.cpp
@@ -19,6 +19,7 @@
 #include <cmath>
 
 #include <hwcutils.h>
+#include "hwctrace.h"
 
 namespace hwcomposer {
 
@@ -115,9 +116,12 @@ void HwcLayer::SetSurfaceDamage(const HwcRegion& surface_damage) {
       UpdateRenderingDamage(rect, rect, true);
       surface_damage_.reset();
       return;
+    } else {
+      state_ |= kSurfaceDamageChanged;
     }
   } else if (rects == 0) {
     rect = display_frame_;
+    state_ &= ~kSurfaceDamageChanged;
   }
 
   if ((surface_damage_.left == rect.left) &&
@@ -126,8 +130,6 @@ void HwcLayer::SetSurfaceDamage(const HwcRegion& surface_damage) {
       (surface_damage_.bottom == rect.bottom)) {
     return;
   }
-
-  state_ |= kSurfaceDamageChanged;
 
   UpdateRenderingDamage(surface_damage_, rect, false);
   surface_damage_ = rect;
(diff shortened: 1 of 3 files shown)


common/core/hwclayer.cpp
```diff
@@ -229,6 +229,41 @@ void HwcLayer::SufaceDamageTransfrom() {
...
+                     (source_crop_.bottom - source_crop_.top));
+#endif
+    int display_width = display_frame_.right - display_frame_.left;
+    int display_height = display_frame_.bottom - display_frame_.top;
+    int source_width = source_crop_.right - source_crop_.left;
+    int source_height = source_crop_.bottom - source_crop_.top;
+    int surface_width = surface_damage_.right - surface_damage_.left;
+    int surface_height = surface_damage_.bottom - surface_damage_.top;
+
+    if (display_width != surface_width && display_height != surface_height) {
```
js0701: This judge is not about the actually display resolution, right?
According to your description,  We need to get the real resolution?

Shao-Feng: In the bug case, the surface Rect is calculated with 1080P coordination when calculating the rect size for 4k/2k display. the size of surface Rect should be enlarged from the source size (1080P) to the display size (2k/4k).

js0701: @Shao-Feng  How about if display is smaller than 1080P?

Shao-Feng: That is why we use “!=” rather than ">" here. when the surface size is not equal with the display, it will be enlarged or reduced.

js0701: This patch is in-consistent with commit description.
Acording to this patch, the damage rect is not zoomed to display frame in SF. Instead HWC shuld zoom it, it is not about the display resolution, 2K or 4K right
@yhe39  Please also comment
Let's test the cases we know to make sure no reg

Shao-Feng: In playing 480P or 4K video with mosaic mode, the display size is already 3840*1080, and the surface is also calculated based on  3840*1080. for 480p, the size of video surface is almost 1920*1080. with this judge, the size will not be changed.

Shao-Feng: Chunge is running a reg scope testing for all 7 bugs which patch is related with this method. we can wait for her result

yhe39: LGTM, i think this patch could be merge if no regression is found. thanks!

Shao-Feng:
Updated to only do the transfer for partial refresh

common/core/hwclayer.cpp
```diff
@@ -115,20 +116,22 @@ void HwcLayer::SetSurfaceDamage(const HwcRegion& surface_damage) {
...
   } else if (rects == 0) {
     rect = display_frame_;
+    state_ &= ~kSurfaceDamageChanged;
   }
 
   if ((surface_damage_.left == rect.left) &&
       (surface_damage_.top == rect.top) &&
       (surface_damage_.right == rect.right) &&
       (surface_damage_.bottom == rect.bottom)) {
+    state_ &= ~kSurfaceDamageChanged;
```
js0701: I don't think this line is necessary?

Shao-Feng: removed

Shao-Feng: rebased

PR #513, opened 2018-12-26
Correct overlay alpha blending value setting

YuanjunHuang:
Correct Alpha value setting to align with i915 kernel driver
in which overlay Alpha blending is first time being enabled
from kernel 4.20.

In driver it takes high 8bit of 16bit value.

Tracked-On: None
Test: boot to Android UI with kernel 4.20

### DIFF
--- a/wsi/drm/drmplane.cpp
+++ b/wsi/drm/drmplane.cpp
@@ -269,7 +269,7 @@ bool DrmPlane::Initialize(uint32_t gpu_fd, const std::vector<uint32_t>& formats,
 bool DrmPlane::UpdateProperties(drmModeAtomicReqPtr property_set,
                                 uint32_t crtc_id, const OverlayLayer* layer,
                                 bool test_commit) const {
-  uint64_t alpha = 0xFF;
+  uint32_t alpha = 0xFFFF;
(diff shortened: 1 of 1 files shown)


wsi/drm/drmplane.cpp
```diff
@@ -283,8 +283,9 @@ bool DrmPlane::UpdateProperties(drmModeAtomicReqPtr property_set,
     fence = layer->GetAcquireFence();
   }
 
+   // i915 driver reads high 8bit of 16bit value
   if (layer->GetBlending() == HWCBlending::kBlendingPremult)
-    alpha = layer->GetAlpha();
+    alpha = layer->GetAlpha() << 8;
```
js0701: Is it possible to out of scope here

YuanjunHuang: GetAlpha returns 8bit. so it is safe to left shift 8 bit

js0701: it is not safe, alpha of layer is in type uint8_t

MicheleL-Intel: It doesn't hurt to AND with the masking bits.

js0701:
Closed as merged. Thanks

PR #516, opened 2018-12-29
HWC, modify the connected display calculate logic

dyang23:
The conneted display count should be calculated before the
drmdisplaymanager try to connect the physical display,
the original method will count a logical connected display,
only already conneted by drmdisplaymanager will be counted,
but since the number will be used during the diaplay
connection, so it will get the wrong display number.

Change-Id: I9f0a96bd34b7c3ef4ea040c798cebfa0f332796a
Tracked-On: https://jira.devtools.intel.com/browse/OAM-73655

### DIFF
--- a/os/android/gralloc1bufferhandler.cpp
+++ b/os/android/gralloc1bufferhandler.cpp
@@ -149,18 +149,17 @@ bool Gralloc1BufferHandler::CreateBuffer(uint32_t w, uint32_t h, int format,
 
   set_format_(gralloc1_dvc, temp->gralloc1_buffer_descriptor_t_, pixel_format);
 #ifdef ENABLE_RBC
+  uint64_t modifier = 0;
   if (set_modifier_) {
-    uint64_t modifier = 0;
     if (preferred_modifier != -1) {
       modifier = preferred_modifier;
     } else {
       modifier = choose_drm_modifier(format);
     }
-
     set_modifier_(gralloc1_dvc, temp->gralloc1_buffer_descriptor_t_, modifier);
   }
 
-  if (modifier_used) {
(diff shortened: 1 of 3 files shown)


js0701:
@dyang23  Seems this getConnectedPhysicalDisplayCount does not deal with hotplug yet

dyang23:
The UpdateDisplayState() will be called also in hotplug event-> HotPlugEventHandler(), so in theory it should be OK.
The EB in under build, https://buildbot.sh.intel.com/absp/builders/f_p_opensourcegfx_master-engineering/builds/88,
we can double confirm this scenario by testing this case.

Thanks,

wsi/drm/drmplane.cpp
```diff
@@ -204,6 +204,10 @@ bool DrmPlane::Initialize(uint32_t gpu_fd, const std::vector<uint32_t>& formats,
 
   // query and store supported modifiers for format, from in_formats
   // property
+  if (!use_modifier) {
```
js0701: Is it important to avoid modificator initialization

dyang23:
________________________________

In wsi/drm/drmplane.cpp<https://github.com/intel/IA-Hardware-Composer/pull/516#discussion_r244445220>:

> @@ -204,6 +204,10 @@ bool DrmPlane::Initialize(uint32_t gpu_fd, const std::vector<uint32_t>& formats,



   // query and store supported modifiers for format, from in_formats

   // property

+  if (!use_modifier) {

Is it important to avoid modificator initialization

I can modify this, to keep it as original, if fact this modification is not related to this bug fix

YuanjunHuang:
LGTM.

js0701:
Closed as merged. Thanks

PR #517, opened 2018-12-29
Remove choose_drm_modifier

js0701:
We should not relay on a simple util functions to choose CCS
modificators. That could be risky if one drm plane does not support
CCS modificator. We should let drmplane tell us which kind of modifier
the plane preferred

Change-Id: Ia9b3c9d0a7a1f05694635e12aaa97a70b880cb9b
Tracked-On: None
Tests: Android UI normal and RBC working

### DIFF
--- a/os/android/gralloc1bufferhandler.cpp
+++ b/os/android/gralloc1bufferhandler.cpp
@@ -153,8 +153,6 @@ bool Gralloc1BufferHandler::CreateBuffer(uint32_t w, uint32_t h, int format,
     uint64_t modifier = 0;
     if (preferred_modifier != -1) {
       modifier = preferred_modifier;
-    } else {
-      modifier = choose_drm_modifier(format);
     }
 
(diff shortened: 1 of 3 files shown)


os/android/gralloc1bufferhandler.cpp
```diff
@@ -153,8 +153,6 @@ bool Gralloc1BufferHandler::CreateBuffer(uint32_t w, uint32_t h, int format,
     uint64_t modifier = 0;
```
YuanjunHuang: better initialize as: uint64_t modifier = DRM_FORMAT_MOD_NONE;

js0701: this may require to include drm header, still need to consider it

js0701: @YuanjunHuang Done

YuanjunHuang: LGTM

js0701:
Closed as merged

PR #523, opened 2019-01-17
Trying to avoid unecessary composition on 2 planes

js0701:
We observed this situation in ACRN
When the displayframe of 2 planes overlaps
composition on separate plane is not necessary

Change-Id: Iefa948785191ecd6516a7503cd449f4d2e26696d
Tracked-On: None
Tests: Android UI normal


js0701:
closed as merged

PR #526, opened 2019-01-21
PAVP:  tear down discard layer before import buffers

js0701:
So the meta flags after import is not set yet, it
should not be used to judge the original buffer is
protected or not. Instead, use IsBufferProtected to
judge the original buffer is protected or not

Change-Id: Ibe38def41c9bfb56a02cac398132776e8cfe988e
Tracked-On: None
Tests: Android UI normal


js0701:
closed as merged

PR #527, opened 2019-01-21
Calculate damage rect for secondary display

Shao-Feng:
The coordinate of damage rect should be calculated according to the
display order in overlaylayer.

Change-Id: I5f7a9a6cf471ee081d86877b6bff830b896662e8
Tests: Compile sucessful for Android.
Tracked-On: https://jira.devtools.intel.com/browse/OAM-74960

### DIFF
--- a/common/core/overlaylayer.cpp
+++ b/common/core/overlaylayer.cpp
@@ -290,6 +290,13 @@ void OverlayLayer::InitializeState(HwcLayer* layer,
     }
 
     // Handle case where we might be using logical and Mosaic together.
+    IMOSAICDISPLAYTRACE("-------Layer[%d]", z_order_);
+    IMOSAICDISPLAYTRACE(
+        "Original display_frame_ %d %d %d %d  left_source_constraint: %d "
+        "left_constraint: %d \n",
+        display_frame_.left, display_frame_.right, display_frame_.top,
+        display_frame_.bottom, left_source_constraint, left_constraint);
+
     display_frame_.left =
         (display_frame_.left - left_source_constraint) + left_constraint;
     display_frame_.right =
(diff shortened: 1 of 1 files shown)


common/core/overlaylayer.cpp
```diff
@@ -300,11 +307,22 @@ void OverlayLayer::InitializeState(HwcLayer* layer,
...
+        "left_constraint: %d \n",
+        surface_damage_.left, surface_damage_.right, surface_damage_.top,
+        surface_damage_.bottom, left_source_constraint, left_constraint);
+
     display_frame_.bottom =
         std::min(max_height, static_cast<uint32_t>(display_frame_.bottom));
     display_frame_width_ = display_frame_.right - display_frame_.left;
     display_frame_height_ = display_frame_.bottom - display_frame_.top;
 
+    surface_damage_.left =
```
js0701: Can you explain now surface_damage_ is in the scale of displayframe?
Why don't we need some scaling

Shao-Feng: surface_damage_ and display_frame are in the same scale here. 
But the display_frame is re-calculated for the physical display in this function as 
display_frame_.left =
        (display_frame_.left - left_source_constraint) + left_constraint;
display_frame_.right =
        (display_frame_.right - left_source_constraint) + left_constraint;

so surface_damage_ need to be handled in the same way.

js0701: @Shao-Feng 
OK.
I don't understand why this is a regression of your last partial update PR

Shao-Feng: In last patch, we only transfer the surface_damage_ for partial refresh. before that patch, surface_damage will always be transferred, that is, below codes will be run on surface_damage even if the surface_damage_ is assigned with display_frame_. (In almost case, it is OK, because ox is 0. but in mosaic video case, it will make surface_damage be moved to right for half screen, and then be handled incorrectly.)
else if (transform_ == 0) {
      ox = display_frame_.left;
      oy = display_frame_.top;
      current_rendering_damage_.left = ox + translated_damage.left;
      current_rendering_damage_.top = oy + translated_damage.top;
      current_rendering_damage_.right = ox + translated_damage.right;
      current_rendering_damage_.bottom = oy + translated_damage.bottom;
    }
With previous patch, above code will only be invoked for partial refresh. the logical is correct. and the Rect of surface_damage_ and display_frame_ are correct too with the patch.

js0701:
Closed as merged

PR #530, opened 2019-01-24
Disable Modifier for plane which has video layer

Shao-Feng:
[DO NOT MERGE] Just checking Linux compiling


common/display/displayplanemanager.cpp
```diff
@@ -800,7 +800,7 @@ void DisplayPlaneManager::EnsureOffScreenTarget(DisplayPlaneState &plane) {
   }
 
   uint64_t modifier = plane.GetDisplayPlane()->GetPreferredFormatModifier();
-  if (plane.IsVideoPlane())
+  if (plane.HasVideoLayer())
```
js0701: If a plane has video layer, then it should be a video plane when going to VPP
Have you debug whey plane .IsVideoPlane is not true?

js0701: We called SetVideoPlane(true) when ForceVPP is called

Shao-Feng: Not clear yet, but "SetVideoPlane" only work for extend display.

js0701: Then we can fix ResetLayers logic,
without checking the layer count to be 1

PR #531, opened 2019-01-25
Set the active config as the fake config when no display is connected

Shao-Feng:
When no display is connected, the fake config id(0) must be set as
active config. otherwise SF will crash when invoking GetActiveConfig
SF only invoke SetActiveConfig when configs.size > 1. In no-display case,
only 1 fake config is return.
1 known issue: if the primary display(2k/4k) is not 1080p, and no
connected in booting. once the primary display is hotpluged. only
1080p will be shown.

Change-Id: I55b6ed900fe77339a8fc4bbbf5435ed9321e76b2
Tests: Compile done. No SF crash happen with no-display, and hotplug

### DIFF
--- a/wsi/physicaldisplay.cpp
+++ b/wsi/physicaldisplay.cpp
@@ -634,7 +634,8 @@ bool PhysicalDisplay::GetDisplayConfigs(uint32_t *num_configs,
     return false;
   *num_configs = 1;
   if (configs) {
-    configs[0] = 1;
+    configs[0] = FAKE_DISPLAY_CONFIG_ID;
+    config_ = FAKE_DISPLAY_CONFIG_ID;
   }
   return true;
 }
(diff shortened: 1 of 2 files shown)


wsi/physicaldisplay.cpp
```diff
@@ -634,7 +634,8 @@ bool PhysicalDisplay::GetDisplayConfigs(uint32_t *num_configs,
     return false;
   *num_configs = 1;
   if (configs) {
-    configs[0] = 1;
+    configs[0] = FAKE_DISPLAY_CONFIG_ID;
+    config_ = FAKE_DISPLAY_CONFIG_ID;
```
js0701: Why in a Get function, member variable config_ need to be change?
Seem strange

Shao-Feng: SF only invoke SetActiveConfig when configs.size > 1. 
In no-display case, only 1 fake config is return by hwc.
and then SF will not invoke SetActiveConfig to set this member.

Another way here, we can return 2 configs as 
configs[0] = 0;
configs[1] = 1;
then no need to set the member

PR #532, opened 2019-01-25
ActiveConfig member should not be reset as 0 in ConnectDisplay

Shao-Feng:
ActiveConfig is sent from SF. should not be reset as 0 in connecting
display

Change-Id: I4517a715d68fd4c9da227a17937da91db442f5df
Tests: Compile sucessful for Android. No crash in headless and hotplug


Shao-Feng:
Update for moving config_ as private

PR #533, opened 2019-01-26
ActiveConfig member should not be reset in ConnectDisplay

Shao-Feng:
ActiveConfig is sent from SF. should not be reset in connecting
display. The default config will be used as Fake config ID too

Known issue: SF expects primary display to be always connected.
Once the primary display is set with fake(1080p). it can not be
updated with disconnect-connect hotplug event again. otherwise, SF
will crash with "null point" when invoking the erased display.
no way to re-fresh the pri-display config except reboot.

Change-Id: I4517a715d68fd4c9da227a17937da91db442f5df
Tests: Compile sucessful for Android. No crash in headless and hotplug

### DIFF
--- a/wsi/drm/drmdisplay.cpp
+++ b/wsi/drm/drmdisplay.cpp
@@ -172,8 +172,18 @@ bool DrmDisplay::ConnectDisplay(const drmModeModeInfo &mode_info,
   connector_ = connector->connector_id;
   mmWidth_ = connector->mmWidth;
   mmHeight_ = connector->mmHeight;
+
+#ifdef ENABLE_ANDROID_WA
+  if (ordered_display_id_ == 0 && IsFakeConnected()) {
+    SetFakeAttribute(mode_info);
+  } else {
+    SetDisplayAttribute(mode_info);
+    config_ = config;
+  }
+#else
   SetDisplayAttribute(mode_info);
   config_ = config;
+#endif
 
   ScopedDrmObjectPropertyPtr connector_props(drmModeObjectGetProperties(
       gpu_fd_, connector_, DRM_MODE_OBJECT_CONNECTOR));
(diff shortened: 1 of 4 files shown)


wsi/drm/drmdisplay.cpp
```diff
@@ -393,7 +391,9 @@ void DrmDisplay::UpdateDisplayConfig() {
     return;
   }
   flags_ |= DRM_MODE_ATOMIC_ALLOW_MODESET;
-  SetDisplayAttribute(modes_[config_]);
+  uint32_t activeconfig = 0;
+  PhysicalDisplay::GetActiveConfig(&activeconfig);
+  SetDisplayAttribute(modes_[activeconfig]);
```
js0701: Why not just use  SetDisplayAttribute(config_)

wsi/drm/drmdisplay.cpp
```diff
@@ -393,7 +391,9 @@ void DrmDisplay::UpdateDisplayConfig() {
     return;
   }
   flags_ |= DRM_MODE_ATOMIC_ALLOW_MODESET;
-  SetDisplayAttribute(modes_[config_]);
+  uint32_t activeconfig = 0;
+  PhysicalDisplay::GetActiveConfig(&activeconfig);
```
js0701: this line is just to get activeconfig=config_?

Shao-Feng: config_ is moved into private in PhysicalDisplay

wsi/physicaldisplay.h
```diff
@@ -287,11 +293,6 @@ class PhysicalDisplay : public NativeDisplay, public DisplayPlaneHandler {
   };
 
   uint32_t pipe_;
-#ifdef ENABLE_ANDROID_WA
```
YuanjunHuang: What was this WA for?

Shao-Feng:
Update for setting mode_info in ConnectDisplay.
No need to consider if SetActiveConfig is invoked or not for ConnectDisplay. Because it is always false.
For primary display, the hotplug event only be sent to SF one time after booting. and never changed except reboot. so SetActiveConfig will be invoked one (primary display is connected) or zero time (headless case).  and ConnectDisplay will be invoked one time too. In primary connected, it is invoked in startHotplugMonitor. In headless case, it is invoked in hotplug connected with Fake config.
for extend display, once the display is re-connected, the display config will be erased and created again on SF, so SF will always invoke SetActiveConfig later. so it is OK to set preferred config id here. 

The mode_info and fake attribute should be set here for fake config too

js0701:
Please resolve confict (rebase) @Shao-Feng

Shao-Feng:
Updated for conflicts

PR #536, opened 2019-01-31
Fix cros gralloc handle const cast warnings

MicheleL-Intel:
TITLE: Suppress const cast warnings restricted to source line number

Every time an Android build is invoked, the following warning plus its corresponding details are generated fourteen (14) times, therefore obscuring the compile error that is of interest:

       "warning: cast from 'const native_handle *' to 'cros_gralloc_handle *' drops const qualifier"

This patch cleans up the build output and restrict suppression of the warning on the line number where the warning is being generated in the source file.


MicheleL-Intel:
If you happen to be reviewing this code, please ignore all *.mk file changes as it is part of another commit 67663a9e .  This commit is currently ahead of HEAD; i.e. HEAD needs to catch up. 

 Review only: os/android/platformdefines.h.  

 Thanks!

PR #539, opened 2019-02-12
Fix the Rect size in play video with overlay plane

Shao-Feng:
With overlay plane, the output region need to be same with displayframe
for normal layer. and need to be assigned with source_crop for video layer

Change-Id: If6e021653671b8b202bda9ae2a10bed7009c00da
Tests: Play video correct with/without overlay plane
Tracked-On: None

### DIFF
--- a/common/compositor/va/varenderer.cpp
+++ b/common/compositor/va/varenderer.cpp
@@ -230,6 +230,76 @@ unsigned int VARenderer::GetVAProcFilterScalingMode(uint32_t mode) {
     return VA_FILTER_SCALING_FAST;
 }
 
+bool VARenderer::GetVideoLayerRect(OverlayLayer* layer_out,
+                                   OverlayLayer* layer_in,
+                                   VARectangle* surface_region,
+                                   VARectangle* output_region) {
+  // Get Input Surface.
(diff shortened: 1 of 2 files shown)


js0701:
Can you explain why video output should use source_crop instead of displayframe

Shao-Feng:
when playing a scaled video (for instance, 480P), the displayframe for video layer is scaled for display. and the source_crop isn't. In scaled video case, if the scaled displayframe is assigned to VA-API, the video buffer will be re-scaled again, and only part of video layer will be shown. This issue only happen when overlay plane is available

js0701:
@Shao-Feng  Understand.

js0701:
One solution is that set the correct source_crop of output layer for video when overlay is used.
The source_crop should be equal to display frame in this case. @Shao-Feng  What do you think

Shao-Feng:
The displayframe and source_crop is assigned by SF, not sure if they can be re-calculated again in hwc. I will take a try and debugging when I come back.

js0701:
The output layer of video renderer is controled by HWC
@Shao-Feng

js0701:
UsePlaneScalar
We should dislabe use of plane scaler if the plane is video, perhaps that is what we should do

PR #540, opened 2019-02-14
Disable PlaneScale for video plane

Shao-Feng:
Video plane can be scaled by VAAPI. set false in UsePlaneScale.

Change-Id: I305ac6c419c8bcd7c86a3eafe60481cdc15baf01
Tests: Play video correct with/without overlay plane
Tracked-On: None

### DIFF
--- a/common/compositor/va/varenderer.cpp
+++ b/common/compositor/va/varenderer.cpp
@@ -249,11 +249,16 @@ bool VARenderer::Draw(const MediaState& state, NativeSurface* surface) {
   HwcRect<int> layer_out_disp_frame = layer_out->GetDisplayFrame();
   int xtranslation = layer_out_disp_frame.left;
   int ytranslation = layer_out_disp_frame.top;
+  HwcRect<float> source_rect;
+  source_rect.top = source_rect.left = 0.0;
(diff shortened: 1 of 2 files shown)


js0701:
It is better to split into 2 commits altought it is short,
It actually do 2 jobs
1) video will not use plane scaler
2) Need to translate rect in va renderer( I guess for mosaic mode???)

common/compositor/va/varenderer.cpp
```diff
@@ -246,14 +246,13 @@ bool VARenderer::Draw(const MediaState& state, NativeSurface* surface) {
 
   // Get Output Surface.
   OverlayLayer* layer_out = surface->GetLayer();
-  HwcRect<int> layer_out_disp_frame = layer_out->GetDisplayFrame();
-  int xtranslation = layer_out_disp_frame.left;
-  int ytranslation = layer_out_disp_frame.top;
-
+  HwcRect<int> layer_out_display_frame = layer_out->GetDisplayFrame();
```
js0701: Those lines are not changed, you don't need change?

Shao-Feng: Updated

js0701:
closed as merged

PR #543, opened 2019-02-18
Reuse last overlay plane on ACRN when multi monitors connected

renchenglei:
The last overlay plane is disabled in past due to APL not
reboot when multi monitors are connected. Now this issue
is not reproduced any more. We add the last overlay plane
back, which could help improve browser kinetic panning frame
rate with 2 displays.

Change-Id: I6a27202d822dc80433f4b5d20048472cf37f4ed8
Tracked-On: OAM-76190
Signed-off-by: Chenglei Ren <chenglei.ren@intel.com>

### DIFF
--- a/common/display/displayplanemanager.cpp
+++ b/common/display/displayplanemanager.cpp
@@ -48,17 +48,11 @@ void DisplayPlaneManager::ResizeOverlays() {
     total_overlays_ = overlay_planes_.size();
     if (total_overlays_ > 1) {
       cursor_plane_ = overlay_planes_.back().get();
-      bool needs_cursor_wa = false;
-#ifdef DISABLE_CURSOR_PLANE
-      needs_cursor_wa = overlay_planes_.size() > 3;
-#endif
       // If this is a universal plane, let's not restrict it to
       // cursor usage only.
-      if (!needs_cursor_wa && cursor_plane_->IsUniversal()) {
+      if (cursor_plane_->IsUniversal()) {
         cursor_plane_ = NULL;
-      }
-
-      if (needs_cursor_wa || (cursor_plane_ && !cursor_plane_->IsUniversal())) {
+      } else {
         total_overlays_--;
       }
     }
@@ -163,13 +157,7 @@ bool DisplayPlaneManager::ValidateLayers(
   if (layer_begin != layer_end) {
     auto overlay_end = overlay_planes_.end();
     if (cursor_plane_) {
-#ifdef DISABLE_CURSOR_PLANE
       overlay_end = overlay_planes_.end() - 1;
-#else
-      if (!cursor_plane_->IsUniversal()) {
-        overlay_end = overlay_planes_.end() - 1;
-      }
-#endif
     }
 
     // Handle layers for overlays.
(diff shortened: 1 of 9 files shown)


js0701:
@renchenglei 
You revert the original patch in this PR
Please specify the origin commit id in the message

renchenglei:
@js0701, done! Thanks a lot.

js0701:
@renchenglei  Can you test this PR in native also when multiple connectors connected?

renchenglei:
@js0701 Tried on Native, and the reboot issue can't be reproduced.

common/display/displayplanemanager.cpp
```diff
@@ -53,11 +48,17 @@ void DisplayPlaneManager::ResizeOverlays() {
     total_overlays_ = overlay_planes_.size();
     if (total_overlays_ > 1) {
       cursor_plane_ = overlay_planes_.back().get();
+      bool needs_cursor_wa = false;
+#ifdef DISABLE_CURSOR_PLANE
+      needs_cursor_wa = overlay_planes_.size() > 3;
```
Shao-Feng: I am worrying if 1 or more plane are reserved, needs_cursor_wa will not be assigned correct here.

js0701: @Shao-Feng 
Can you see when this DISABLE_CURSOR_PLANE is defined?
can we remove this MACRO

Shao-Feng: It is defined in BoardConfig.mk under device/intel for different board as "BOARD_CURSOR_WA := true/false".
It could be true in future board

common/display/displayplanemanager.cpp
```diff
@@ -163,13 +157,7 @@ bool DisplayPlaneManager::ValidateLayers(
   if (layer_begin != layer_end) {
     auto overlay_end = overlay_planes_.end();
     if (cursor_plane_) {
-#ifdef DISABLE_CURSOR_PLANE
       overlay_end = overlay_planes_.end() - 1;
```
js0701: If DISABLE_CURSOR_PLANE is not defined, seems this line should be ommited?

js0701: Instead line 169 to 171 shoud remain

common/display/displayplanemanager.cpp
```diff
@@ -467,11 +455,9 @@ void DisplayPlaneManager::ValidateCursorLayer(
     overlay_begin = overlay_planes_.begin() + composition.size();
   }
 
-#ifdef DISABLE_CURSOR_PLANE
   overlay_end = overlay_planes_.end() - 1;
   if (total_size == 1)
     overlay_begin = overlay_planes_.begin() + composition.size();
-#endif
```
js0701: Why lines in DISABLE_CURSOR_PLANE needs to remain

common/display/displayplanemanager.cpp
Shao-Feng: 165-167 is same with 169-171 since enable_last_plane_  has been removed

js0701: Ok.

common/display/displayplanemanager.cpp
Shao-Feng: Yes, these lines should be removed. I will update soon

js0701:
@Shao-Feng 
Please go with merge if review and validation process done

Shao-Feng:
Validated

PR #547, opened 2019-02-22
Delete the output buffer when it is same as output_handle_

yhe39:
In ACRN virtualdisplay setoutput buffer, if the buffer is same as output_handle_
and output_handle_ is not null, delete the buffer to prevent memory leak.

Change-Id: I558f31aea3315481f3a00c1fef1202bb7eb124f1
Tracked-On: None
Tests: Compile sucess for ACRN and baremetal. System without issue.

### DIFF
--- a/common/display/virtualdisplay.cpp
+++ b/common/display/virtualdisplay.cpp
@@ -436,6 +436,8 @@ void VirtualDisplay::SetOutputBuffer(HWCNativeHandle buffer,
     if (output_handle_) {
       handler->CopyHandle(output_handle_, &handle_);
     }
+  } else {
+    delete buffer;
   }
 
   if (acquire_fence_ > 0) {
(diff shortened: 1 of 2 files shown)


yhe39:
@js0701 @shuangwan01 @Shao-Feng please help to review. Thanks

common/display/virtualdisplay.cpp
```diff
@@ -420,6 +420,10 @@ void VirtualDisplay::SetOutputBuffer(HWCNativeHandle buffer,
   }
 #endif
 
+  if (output_handle_ && output_handle_ == buffer) {
```
Shao-Feng: This situation is same with the negation of next if statement. It is better adding a else statement instead of add a new if segment.

common/display/virtualpanoramadisplay.cpp
```diff
@@ -396,6 +396,10 @@ void VirtualPanoramaDisplay::CreateOutBuffer() {
 
 void VirtualPanoramaDisplay::SetOutputBuffer(HWCNativeHandle buffer,
                                              int32_t acquire_fence) {
+  if (output_handle_ && output_handle_ == buffer) {
```
Shao-Feng: same with above

yhe39:
Thanks @Shao-Feng , modified accordingly.

js0701:
Thanks.
@Shao-Feng 
Please merge it if no more comments

PR #550, opened 2019-02-25
Added the panorama dynamic turn on/off feature.

shuangwan01:
Added enable/disable panorama interfaces in HWC service API for
it be able to controlled by other process for dynamic turning
on/off panorama view mode in ACRN AaaG environment.

Change-Id: Ic3e5f22d65a533132adf6752a4f93b060476d6ca
Tracked-On: https://jira.devtools.intel.com/browse/OAM-76498
Tests: Turn on/off Panorama mode dynamically on ACRN AaaG.

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -258,6 +258,7 @@ void GpuDevice::InitializePanorama(
         panorama_sos_displays.at(0).at(sos_it));
     virtualdisp->InitVirtualDisplay(1920, 1080);
     i_available_panorama_displays.emplace_back(virtualdisp);
+    virtual_panorama_displays_.emplace_back(virtualdisp);
   }
 
   // Add the native displays
@@ -281,6 +282,8 @@ void GpuDevice::InitializePanorama(
             // Skip the disconnected display here
             i_available_panorama_displays.emplace_back(
                 temp_displays.at(panorama_displays.at(m).at(l)));
+            physical_panorama_displays_.emplace_back(
+                temp_displays.at(panorama_displays.at(m).at(l)));
             // Add tag for panorama-ed displays
             available_displays.at(panorama_displays.at(m).at(l)) = false;
           }
@@ -297,12 +300,29 @@ void GpuDevice::InitializePanorama(
     std::unique_ptr<MosaicDisplay> panorama(
         new MosaicDisplay(i_available_panorama_displays));
     panorama->SetPanoramaMode(true);
-    panorama->SetExtraDispInfo((int)panorama_displays.size(),
-                               (int)panorama_sos_displays.size());
+    panorama->SetExtraDispInfo(&virtual_panorama_displays_,
+                               &physical_panorama_displays_);
+    ptr_mosaicdisplay = (MosaicDisplay *)panorama.get();
     panorama_displays_.emplace_back(std::move(panorama));
     // Save the panorama to the final displays list
     total_displays_.emplace_back(panorama_displays_.back().get());
   }
+}
+
+bool GpuDevice::TriggerPanorama(uint32_t hotplug_simulation) {
+  bool ret = false;
+  if (ptr_mosaicdisplay) {
+    ret = ptr_mosaicdisplay->TriggerPanorama(hotplug_simulation);
+  }
+  return ret;
+}
+
+bool GpuDevice::ShutdownPanorama(uint32_t hotplug_simulation) {
+  bool ret = false;
+  if (ptr_mosaicdisplay) {
+    ret = ptr_mosaicdisplay->ShutdownPanorama(hotplug_simulation);
+  }
+  return ret;
 }
 
 #endif
--- a/common/core/mosaicdisplay.cpp
+++ b/common/core/mosaicdisplay.cpp
@@ -24,6 +24,11 @@
 
 #include "hwctrace.h"
 
+#ifdef ENABLE_PANORAMA
+#include "displaymanager.h"
+#include "virtualpanoramadisplay.h"
+#endif
+
 namespace hwcomposer {
 
 class MDVsyncCallback : public hwcomposer::VsyncCallback {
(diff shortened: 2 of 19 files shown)


js0701:
@Shao-Feng @YuanjunHuang @yhe39 
Please help review the source code

common/core/mosaicdisplay.cpp
```diff
@@ -542,13 +609,96 @@ bool MosaicDisplay::ContainConnector(const uint32_t connector_id) {
...
+  for (uint32_t i = 0; i < physical_panorama_displays->size(); i++) {
+    real_physical_displays_.emplace_back(physical_panorama_displays->at(i));
+  }
+}
+
+bool MosaicDisplay::EnablePanorama() {
+  if (!panorama_mode_)
+    return false;
+
+  if (!panorama_enabling_state_) {
```
js0701: may need to lock to protect this panorama_enabling_state_
suppose this function is in the HWC.info binder thread
and obvious this state_ should be protected as functions in other thread may need to access this too
So this state need to be protected by lock

os/android/hwcservice.h
```diff
@@ -97,6 +97,11 @@ class HwcService : public BnService {
 
     status_t DisableHDCPSessionForAllDisplays();
 
+#ifdef ENABLE_PANORAMA
+    status_t EnablePanorama(uint32_t hotplug_simulation);
```
YuanjunHuang: suggest change the name enable/disable to Trigger/Shutdown if you have to keep two APIs.

os/android/libhwcservice/test/hwcservice_test.cpp
```diff
@@ -61,8 +67,16 @@ int main(int argc, char** argv) {
   bool disable_hdcp_for_display = false;
   bool disable_hdcp_for_all_display = false;
   bool restore = false;
+#ifdef ENABLE_PANORAMA
+  bool enable_panorama = false;
```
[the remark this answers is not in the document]
shuangwan01: Okay, will change the variable names accordingly. Prefer to keep the 2 variables since the test app is not long last daemon and no way to keep current panorama status. Using separated variable will make the code simpler.

common/core/mosaicdisplay.cpp
shuangwan01: Okay, will revise the code pieces to have panorama_enabling_state_ lock protected.

common/display/virtualpanoramadisplay.cpp
```diff
@@ -492,6 +522,24 @@ int VirtualPanoramaDisplay::GetDisplayPipe() {
   return -1;
 }
 
+#ifdef HYPER_DMABUF_SHARING
+bool VirtualPanoramaDisplay::SetHyperDmaBufMode(uint32_t mode) {
+  if (hyper_dmabuf_mode_ != mode) {
+    hyper_dmabuf_mode_ = mode;
+    if (!hyper_dmabuf_mode_) {
```
YuanjunHuang: Is this expected? if so need more explain/comment here.

shuangwan01: Yes, this is expected. And just revised the code comments with more details. Thanks!

os/android/hwcservice.h
shuangwan01: The API naming has been revised accordingly in latest commit.

common/display/virtualpanoramadisplay.cpp
```diff
@@ -492,6 +522,29 @@ int VirtualPanoramaDisplay::GetDisplayPipe() {
...
```
YuanjunHuang: why not if (hyper_dmabuf_mode_) so that else is not needed?

shuangwan01: Reasonable, updated the PR accordingly :-)

common/core/mosaicdisplay.h
```diff
@@ -141,9 +150,20 @@ class MosaicDisplay : public NativeDisplay {
   bool pending_vsync_ = false;
   bool update_connected_displays_ = true;
 #ifdef ENABLE_PANORAMA
+  std::vector<NativeDisplay *> *virtual_panorama_displays_;
+  std::vector<NativeDisplay *> *physical_panorama_displays_;
+  std::vector<NativeDisplay *> real_physical_displays_;
+  DisplayManager *displaymgr_ptr_ = NULL;
```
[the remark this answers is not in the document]
shuangwan01: Yes, reasonable, will remove it accordingly.

common/core/mosaicdisplay.cpp
```diff
@@ -170,10 +196,33 @@ bool MosaicDisplay::Present(std::vector<HwcLayer *> &source_layers,
                             int32_t *retire_fence,
                             PixelUploaderCallback *call_back,
                             bool /*handle_constraints*/) {
-  if (power_mode_ != kOn)
+  if (power_mode_ != kOn) {
+#ifdef ENALBE_PANORAMA
+    if (skip_update_) {
+      event_.Signal();
```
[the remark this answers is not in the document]
shuangwan01: Because the trigger/shutting is triggered by app, so even when the mode is not On but it doesn't block end user to do such action.

common/core/gpudevice.cpp
```diff
@@ -297,14 +300,30 @@ void GpuDevice::InitializePanorama(
...
+    ptr_mosaicdisplay->TriggerPanorama(hotplug_simulation);
+  }
+  return true;
+}
+
+bool GpuDevice::ShutdownPanorama(uint32_t hotplug_simulation) {
+  if (ptr_mosaicdisplay) {
+    ptr_mosaicdisplay->ShutdownPanorama(hotplug_simulation);
+  }
+  return true;
```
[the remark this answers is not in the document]
shuangwan01: Reasonable, will revise the PR accordingly.

common/core/mosaicdisplay.h
```diff
@@ -141,9 +150,20 @@ class MosaicDisplay : public NativeDisplay {
...
+  SpinLock panorama_lock_;
   bool panorama_mode_ = false;
+  bool panorama_enabling_state_ = false;
+  bool skip_update_ = false;
```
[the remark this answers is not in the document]
shuangwan01: DRM displays are just part of child displays of mosaic display, so they are different in general. Anyway, since we have binder API for other process to control the HWC updating mechanism, we need to consider how to handle it properly for all display modes. We could address this in a separated PR. Thanks!

PR #553, opened 2019-03-01
An API for dropping DrmMaster and ignoring DRM commit

Shao-Feng:
New API 'ResetDrmMaster' is added for dropping DrmMaster and
ignoring DRM commit when the parameter is true. otherwise, setting
DrmMaster and force refreshing. After dropping DrmMaster, the thread
of GpuDevice will be resume to monitor 'hwc.lock'. Once the file
lock is grabbed. DrmMaster will be set, and stop ignoring vblank.

Change-Id: Ie31937d710c3813d42770d38c8c80de05ae661dd
Tests: Compile sucessful for Android. Work well with KMSCube
Tracked-On: https://jira.devtools.intel.com/browse/OAM-76578

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -29,6 +29,21 @@ GpuDevice::GpuDevice() : HWCThread(-8, "GpuDevice") {
 
 GpuDevice::~GpuDevice() {
   display_manager_.reset(nullptr);
+  HWCThread::Exit();
+
+  if (-1 != lock_fd_) {
+    close(lock_fd_);
+    lock_fd_ = -1;
+  }
+}
+
+void GpuDevice::ResetAllDisplayCommit(bool enable) {
+  if (enable == enable_all_display_)
+    return;
+  enable_all_display_ = enable;
+  size_t size = total_displays_.size();
+  for (size_t i = 0; i < size; i++)
+    total_displays_.at(i)->EnableDRMCommit(enable);
 }
 
 bool GpuDevice::Initialize() {
@@ -57,13 +72,19 @@ bool GpuDevice::Initialize() {
     display_manager_->RemoveUnreservedPlanes();
   }
 
-  lock_fd_ = open("/vendor/hwc.lock", O_RDONLY);
+  // Ignore all updates by default
+  ResetAllDisplayCommit(false);
+  lock_fd_ = open(HWC_LOCK_FILE, O_RDONLY);
   if (-1 != lock_fd_) {
     if (!InitWorker()) {
       ETRACE("Failed to initalize thread for GpuDevice. %s", PRINTERROR());
     }
   } else {
-    ITRACE("Failed to open " LOCK_DIR_PREFIX "/hwc.lock file!");
+    ITRACE("Failed to open %s", HWC_LOCK_FILE);
+    // HWC should become drm master and start to commit.
+    // if hwc.lock is not available
+    display_manager_->setDrmMaster(true);
+    ResetAllDisplayCommit(true);
   }
 
   return true;
@@ -103,11 +124,38 @@ void GpuDevice::GetConnectedPhysicalDisplays(
 }
 
 bool GpuDevice::EnableDRMCommit(bool enable, uint32_t display_id) {
-  // TODO clean all display for commit status
   size_t size = total_displays_.size();
   bool ret = false;
   if (size > display_id)
     ret = total_displays_.at(display_id)->EnableDRMCommit(enable);
+  return ret;
+}
+
+bool GpuDevice::ResetDrmMaster(bool drop_master) {
+  bool ret = true;
+  if (drop_master) {
+    ResetAllDisplayCommit(false);
+    display_manager_->DropDrmMaster();
+    ITRACE("locking %s and monitoring if %s is unlocked.", HWC_LOCK_FILE,
+           HWC_LOCK_FILE);
+    // Resume GpuDevice thread to check hwc.lock and re-apply drm master.
+    lock_fd_ = open(HWC_LOCK_FILE, O_RDONLY);
+    if (-1 != lock_fd_) {
+      // Only resume GpuDevice thread for dropping DRM Master and
+      // the lock file exist.
+      Resume();
+      return !display_manager_->IsDrmMaster();
+    }
+  }
+  // In case of setDrmMaster or the lock file is not exist.
+  // Re-set DRM Master true.
+  display_manager_->setDrmMaster(false);
+  ResetAllDisplayCommit(true);
+  DisableWatch();
+  if (drop_master)
+    ret = !display_manager_->IsDrmMaster();
+  else
+    ret = display_manager_->IsDrmMaster();
   return ret;
 }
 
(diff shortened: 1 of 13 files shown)


Shao-Feng:
2 way to setDrmMaster for HWC.
firstly, kmscube/earlyevs invoke ResetDrmMaster(0) for normal exit process
secondly, kmscube/earlyevs unlock 'hwc.lock' for kmscube/earlyevs crash case.

Whatever way kmscube/earlyevs take, kmscube/earlyevs need to dropDrmMaster for themselves first. otherwise, hwc will not take drm Master.

Shao-Feng:
Updated for add a new thread to keep invoking drmSetMaster if drmSetMaster return fail.
Now, kmscube/earlyevs can crash without addition operation.

common/core/gpudevice.cpp
```diff
@@ -57,13 +69,15 @@ bool GpuDevice::Initialize() {
     display_manager_->RemoveUnreservedPlanes();
   }
 
-  lock_fd_ = open("/vendor/hwc.lock", O_RDONLY);
+  // Ignore all updates by default
+  ResetAllDisplayCommit(false);
+  lock_fd_ = open(HWC_LOCK_FILE, O_RDONLY);
```
js0701: It seems a defect to me that if no lock_fd_ HWC resist to initial
If no lock_fd, HWC should open gpu_fd and become master I think
No hwc.lock should mean the system want no other application to hold drm master

common/core/gpudevice.cpp
```diff
@@ -103,14 +117,36 @@ void GpuDevice::GetConnectedPhysicalDisplays(
...
 
+bool GpuDevice::ResetDrmMaster(bool drop_master) {
+  // clean all display for commit status
+  bool ret = true;
+  if (drop_master) {
+    ResetAllDisplayCommit(false);
+    display_manager_->DropDrmMaster();
+    ITRACE("locking %s and monitoring if %s is unlocked.", HWC_LOCK_FILE,
+           HWC_LOCK_FILE);
+    // launch a thread to check hwc.lock and re-apply drm master.
```
js0701: You don't need a new thread
HWC has already a thread for vsync
In the vsync thread, you can add logic

Check HWC is master or not,
If not:
If( no hwc.lock or it is not locked) {
setSelftoBeDRM master
ForeceRefresh(); //this will force SurfaceFlinger to send Present Call to HWC, so the UI appear immediately
}

common/core/gpudevice.cpp
```diff
@@ -904,22 +938,24 @@ void GpuDevice::HandleRoutine() {
   // TODO: Add splash screen support.
   if (lock_fd_ != -1) {
     display_manager_->IgnoreUpdates();
-    update_ignored = true;
 
     if (flock(lock_fd_, LOCK_EX) != 0) {
-      ETRACE("Failed to wait on the hwc lock.");
+      ITRACE("Fail to grab the hwc lock.");
```
js0701: Why need this drop master logic?

wsi/drm/drmdisplaymanager.cpp
```diff
@@ -442,17 +446,73 @@ void DrmDisplayManager::IgnoreUpdates() {
...
 
+void DrmDisplayManager::DrmMasterThread::NeedDrmMaster(bool need) {
+  need_drm_master_ = need;
+}
+
+void DrmDisplayManager::setDrmMaster() {
+  int ret = drmSetMaster(fd_);
+  if (ret) {
+    ETRACE("Failed to call drmSetMaster : %s", PRINTERROR());
+    drm_master_thread_->NeedDrmMaster(true);
```
js0701: Do we need a new drm_master_thread_?

js0701: The logic here does not make sure setDrmMaster is successful
However, we need to make sure it is successful
Suggest to keep origin logic instead of using a thread

wsi/drm/drmdisplaymanager.cpp
```diff
@@ -45,10 +45,12 @@ DrmDisplayManager::DrmDisplayManager() : HWCThread(-8, "DisplayManager") {
 DrmDisplayManager::~DrmDisplayManager() {
   CTRACE();
   std::vector<std::unique_ptr<DrmDisplay>>().swap(displays_);
+
 #ifndef DISABLE_HOTPLUG_NOTIFICATION
   close(hotplug_fd_);
 #endif
   drmClose(fd_);
+  drm_master_thread_.reset(nullptr);
```
js0701: a unique_ptr does not need this line to be released

common/core/gpudevice.cpp
Shao-Feng: Get it. If  lock_fd_ == -1, HWC will take drmmaster, and forcerefresh

wsi/drm/drmdisplaymanager.cpp
Shao-Feng: The origin logic will block to setDrmMaster and enter a endless loop, once kmscube/earlyevs still hold drmMaster and invoke HWC to setDrmMaster. Need a async thread to setDrmMaster. 
Once the variable need_drm_master still true, the thread will keep running for setDrmMaster

common/core/gpudevice.cpp
Shao-Feng: The GpuDevice Thread is original here for checking hwc.lock. we just resume it here. 
In original implementation, this thread will exit after getting the lock and commiting frame.
Now, the thread will be waiting after getting the lock and commiting frame in booting first run.
and be resumed once this API is invoked by kmscube/earlyevs.

Shao-Feng: I will add the logic you mentioned into vblank thread.

common/display/vblankeventhandler.cpp
```diff
@@ -114,6 +123,8 @@ void VblankEventHandler::HandleRoutine() {
   int ret = drmWaitVBlank(fd, &vblank);
   if (!ret)
     HandlePageFlipEvent(vblank.reply.tval_sec, (int64_t)vblank.reply.tval_usec);
+
+  SetDrmMaster();
```
YuanjunHuang: dont think it is a good idea to check and set drm master in vblank handler coz every display has a vblank handler and the drm master is gpu device-wide related.

Shao-Feng:
Updated for review comments.
VBlank thread is not used.
When set/drop DRM master, will try 10 rounds if drmSetMaster/drmDropMaster return fail.
add 1 parameter "synced" in setDrmMaster for 2 ways to set Drm-Master.
synced = true, will only try 10 rounds if drmSetMaster/drmDropMaster always return fail.
synced = false, will always invoke drmSetMaster with 10ms delay until drmSetMaster return pass.

In initialize and thread (lock is available), use synced = false.
In external API, use synced = true.

Shao-Feng:
Updated for re-set DRM master once the lock file can not be opened.
Change the parameter name of setDrmMaster to "must_set". True for endless loop until drmSetMaster success. False for 10 rounds retry with 10 ms delay if drmSetMaster fail

PR #559, opened 2019-03-12
Should not change ignoreUpdate status in GpuDevice Thread

Shao-Feng:
Already use 'ForceIgnore' to control ignore/update here.
No need to invoke it again. and it will block to show UI
on Mosaic and logical mode.

Change-Id: I3b77628094df3c2971a50abf3ade00949e2f0947
Tracked-On: None
Tests: Work well with Mosaic and logical mode.

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -1022,8 +1022,6 @@ void GpuDevice::HandleRoutine() {
   //    we need to take control.
   // TODO: Add splash screen support.
   if (lock_fd_ != -1) {
-    display_manager_->IgnoreUpdates();
-
     if (flock(lock_fd_, LOCK_EX) != 0) {
       ITRACE("Fail to grab the hwc lock.");
     } else {


common/core/gpudevice.cpp
```diff
@@ -1022,8 +1022,6 @@ void GpuDevice::HandleRoutine() {
   //    we need to take control.
   // TODO: Add splash screen support.
   if (lock_fd_ != -1) {
-    display_manager_->IgnoreUpdates();
```
YuanjunHuang: looks good to me. 
but why double setting FrameStateTracker::kIgnoreUpdates and FrameStateTracker::kForceIgnoreUpdates will cause mosaic and logical mode issue?

Shao-Feng: We just change the status 'forceignore' here. and this flag only work in drm and physical display. it is a bug for ignoring and forcerefreshing on Mosaic and logic display. the flag 'forceignore' didn't work for them. It should be fixed. I close this one. and will prepare another for adding this flag into mosaic and logical too

PR #562, opened 2019-03-21
Using std::Map to manage the VDS create/destroy.

feifanx:
Change-Id: Ic6656ac0b78df4eb2aa433104eaad3e8dc3a0ac8
Tracked-On: None
Test: Android UI normal
Signed-off-by: xue.yifei <yifei.xue@intel.com>

### DIFF
--- a/wsi/drm/drmdisplaymanager.cpp
+++ b/wsi/drm/drmdisplaymanager.cpp
@@ -390,9 +390,8 @@ NativeDisplay *DrmDisplayManager::CreateVirtualDisplay(uint32_t display_index) {
   NativeDisplay *latest_display;
   std::unique_ptr<VirtualDisplay> display(
       new VirtualDisplay(fd_, buffer_handler_.get(), display_index, 0));
-  virtual_displays_.emplace_back(std::move(display));
-  size_t size = virtual_displays_.size();
-  latest_display = virtual_displays_.at(size - 1).get();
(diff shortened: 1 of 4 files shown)


wsi/drm/drmdisplaymanager.cpp
```diff
@@ -585,7 +585,7 @@ NativeDisplay *DrmDisplayManager::CreateVirtualPanoramaDisplay(
   NativeDisplay *latest_display;
   std::unique_ptr<VirtualPanoramaDisplay> display(
       new VirtualPanoramaDisplay(fd_, buffer_handler_.get(), display_index, 0));
-  virtual_displays_.emplace_back(std::move(display));
+  virtual_displays_.emplace(display_index,std::move(display));
```
shuangwan01: The returned VirtualPanoramaDisplay is not the one just created, this will introduce the HWC crash. Actually, there is no need to put new created VirtualPanoramaDisplay into map as it will introduce another crash when surfaceflinger handle the hotplug. You could refer the commit https://github.com/intel/IA-Hardware-Composer/pull/550, just create the the instance and return it accordingly.

feifanx: update it with your patch on acrn stable

Shao-Feng:
This patch is already cherry-picked in 1A, merged in public too

PR #571, opened 2019-04-18
The buffer interlace info should change to frame by frame

yhe39:
So far the buffer interlace is set according to GRALLOC_USAGE_PRIVATE_0.
It will cause play some irregular size video has green line in screen bottom.

Change-Id: I2bff0bf84a056c1e313cb79c2762e788598fd7ac
Tests: Video play without issue.
Tracked-On: https://jira.devtools.intel.com/browse/OAM-79458

### DIFF
--- a/common/core/overlaylayer.cpp
+++ b/common/core/overlaylayer.cpp
@@ -106,6 +106,11 @@ void OverlayLayer::SetBuffer(HWCNativeHandle handle, int32_t acquire_fence,
     }
   } else {
     buffer->SetOriginalHandle(handle);
+    // need to update interlace info since interlace info is update frame by
+    // frame
+    if (buffer->GetUsage() == kLayerVideo)
+      buffer->SetInterlace(
+          resource_manager->GetNativeBufferHandler()->GetInterlace(handle));
   }
 
   buffer->SetDataSpace(dataspace_);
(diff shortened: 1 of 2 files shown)


yhe39:
@js0701 @YuanjunHuang @Shao-Feng  please help to review.
Actually this patch revert some part of https://github.com/intel/IA-Hardware-Composer/commit/6c8cfc455267c881c433d508814c68e4929b0b95 

Deinterlace: change the default settings to auto
interlace infomation should be according to GRALLOC_USAGE_PRIVATE_0
frame by frame tracking is not necessary

The above patch will cause the green line issue.

js0701:
hi Yue,
We need to evaluate the performance.
I remmber that if we check interface frame by frame,
it can't hit 60FPS for some stream

yhe39:
ok, i will check the video fps, the https://github.com/intel/IA-Hardware-Composer/commit/6c8cfc455267c881c433d508814c68e4929b0b95 doesn't link any issue, so not sure the impact. 
i will check:)

js0701:
I mean that if the GetInterlace call is though the IOCTL GEM USERDATA, it will downgrace the performance

yhe39:
i understand, i will trigger one eb and run the video related pnp test to evaluate this patch. Thanks.

yhe39:
Anfeng helps to run PnP test on GP_P and celadon_p, there is no PnP regression. thanks!

PR #576, opened 2019-04-23
Add solid color support in VA

yhe39:
If the layer do not have buffer only contain the color info,
set the output_region to 0, the surface_region will filled with
background color

Change-Id: I67f576aea90d7e77c74eb9404fe9b99ce7950eee
Tracked-On: https://jira.devtools.intel.com/browse/OAM-76646
Tests: video play without issue.


yhe39:
This patch need modify, some logic is not right.
Will update later, so do not need review so far. Thanks.

PR #577, opened 2019-05-07
Assign rendering area with Damage size except rotate and display-diff

Shao-Feng:
SurfaceDamage should be assigned to rendering_area when surface damage
is empty except rotate_mode or the resolution of connected display is
different.

Change-Id: I0c3b1f33c51c3ed2bfaedb5f7d4d73696a00576c
Test: Work well for benchmark T-Rex and Manhattan.
Tracked-On: https://jira.devtools.intel.com/browse/OAM-80208

### DIFF
--- a/public/gpudevice.h
+++ b/public/gpudevice.h
@@ -64,6 +64,12 @@ class GpuDevice : public HWCThread {
 
   void GetConnectedPhysicalDisplays(std::vector<NativeDisplay*>& displays);
 
+  bool IsSameResolutionForAllDisplay();
+
+  bool IsRotateMode() {
+    return rotate_mode_;
+  }
+
   const std::vector<NativeDisplay*>& GetAllDisplays();
 
   void RegisterHotPlugEventCallback(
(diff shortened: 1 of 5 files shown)


Shao-Feng:
@jennycao  Please help to verify for PnP Benchmark T-Rex and Manhattan.
Also the bug OAM-71886, OAM-73581(Celadon)

public/gpudevice.h
```diff
@@ -64,6 +64,12 @@ class GpuDevice : public HWCThread {
 
   void GetConnectedPhysicalDisplays(std::vector<NativeDisplay*>& displays);
 
+  bool IsSameResolutionForAllDisplay();
+
+  bool IsRotateMode() {
+    return rotate_mode_;
```
js0701: What is rotate_mode_ mean?
each display can have different roate, for example, display 0 is not rotated, while display 1 is rotated

js0701: Here it seems you make it golobal?

Shao-Feng: I didn't check the rotate configuring for each display, currently, only check if the configure item ROTATION="false" is true or false.  Yes, I agree with you. we can always enable surface_damage for primary display and un-rotated display even with the exception. I will try to update it

js0701: Why when rotated, surface damage can't be use?

Shao-Feng: SF limitation, the coordinate of surface_damage/displayframe are calculated only with the primary display size. SF did not send the rotated coordinate to hwc when the extend display is rotated. HWC will use incorrect coordinate to set GL scissor. then some layer are not updated correctly

js0701: The coordinate of surface_damage is not related with display size or rotation, that's my knowledge, the coordination of surface_damage is similar with source_corp as I know

PR #578, opened 2019-05-08
Update rendering area when surface_damage is empty

Shao-Feng:
When surface_damage is empty, it mean the layer is not changed.
the rendering area must be set as empty to avoid re-render a
unchanged layer.

Test: Work well for benchmark T-Rex and Manhattan.
Change-Id: I8abee7408b1338617d85e7af6696e42375e7d788
Tracked-On: https://jira.devtools.intel.com/browse/OAM-80208

### DIFF
--- a/common/core/hwclayer.cpp
+++ b/common/core/hwclayer.cpp
@@ -106,21 +106,20 @@ void HwcLayer::SetDisplayFrame(const HwcRect<int>& display_frame,
 void HwcLayer::SetSurfaceDamage(const HwcRegion& surface_damage) {
   uint32_t rects = surface_damage.size();
   state_ |= kLayerContentChanged;
+  state_ |= kSurfaceDamageChanged;
   HwcRect<int> rect;
   ResetRectToRegion(surface_damage, rect);
   if (rects == 1) {
     if ((rect.top == 0) && (rect.bottom == 0) && (rect.left == 0) &&
         (rect.right == 0)) {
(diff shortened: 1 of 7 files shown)


Shao-Feng:
This patch will cause OAM-71886, OAM-73581(Celadon).  Do not merge until we dound the solution for the 2 bugs.

Shao-Feng:
Add a commit for resolving the issue OAM-71886. Please help to review.

Shao-Feng:
2 commits are updated. Please help to review
first one is for resolving PnP regression issue. the root cause of this issue is that the partial refresh is disabled on Q. because all 0 surfacedamage are not set as rendering area. so re-enable 0 surfacedamage for un-changed layer.
Second one is for resolving rotation issue. In rotation, the damage is not calculated correctly. because the transform_ never be sent from SF, and always 0. From SF, the transform is always 0. so we have to use plane_transform_ to calculate the damage location according to display size. this patch for resolving the issue OAM-71886

Shao-Feng:
Issue in playing video with rotation mode is not reproduced by these patch. actually, on ww21 celadon release, video can not be played in rotation mode. will commit another patch for fixing it

Shao-Feng:
With and without these patches, HWC will meet similar issue for supporting "SurfaceFlinger Rotation" (command "adb shell service call window 47 i32 2").
Merge these patch first. and will check what is the issue in supporting "SurfaceFlinger Rotation"

PR #580, opened 2019-05-10
Correct the alpha logic in glProgram

js0701:
the alpha cover should not be necessary

Change-Id: Ieef8282d431d73ff8d5520a1bc4e3b6262721e9b
Tests: Android UI correctly rendered
Tracked-On: None


js0701:
Don't for merge. Experiment Only

js0701:
The composition method is a simpler method that draw a texture each time to the target.
I tested it can work well
Here we just want to see if it can bring any performance improve compare the original composition method
HWC original composition is to draw all textures together once, thus it need to be separated into different rectangles that is with different texture source

js0701:
Do not merge because we don't see performance improvement

tpalli:
> Do not merge because we don't see performance improvement

Was the intention here to use GL_EXT_shader_framebuffer_fetch and gl_LastFragData somehow?

PR #581, opened 2019-05-13
Check if hwc is drm master by default in initialization

Shao-Feng:
Check if hwc is drm master by default in initialization

Use the method drmAuthMagic to check if hwc is already DRM master
or not.

Change-Id: I1931017e1223ae550c52816e6b2a64c74bdf9190
Tests: Work well on celadon with kernel 4.19.19
Tracked-On: None

### DIFF
--- a/wsi/drm/drmdisplaymanager.cpp
+++ b/wsi/drm/drmdisplaymanager.cpp
@@ -444,6 +444,28 @@ void DrmDisplayManager::IgnoreUpdates() {
   }
 }
 
+bool DrmDisplayManager::IsDrmMasterByDefault() {
+  spin_lock_.lock();
+  if (drm_master_) {
+    spin_lock_.unlock();
+    return drm_master_;
+  }
+  drm_magic_t magic = 0;
+  int ret = 0;
+  ret = drmGetMagic(fd_, &magic);
+  if (ret)
+    ETRACE("Failed to call drmGetMagic : %s", PRINTERROR());
+  else {
+    ret = drmAuthMagic(fd_, magic);
+    if (ret)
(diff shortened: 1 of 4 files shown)


wsi/drm/drmdisplaymanager.cpp
```diff
@@ -454,6 +454,10 @@ void DrmDisplayManager::setDrmMaster(bool must_set) {
   uint8_t retry_times = 0;
   do {
     ret = drmSetMaster(fd_);
+
+    // Revert me, A workaround for supporting kernel before 4.19.31
+    if (errno == EACCES)
```
YuanjunHuang: is there any drm API to check if current client is already drm master?

Shao-Feng: No, no API is defined in xf86drm.h for the Master status. Only 2 methods are related with Master.
extern int drmSetMaster(int fd);
extern int drmDropMaster(int fd);

js0701:
What is the return code of kernel before 4.19.31 if another process is already drm master?

Shao-Feng:
the return code is -1 before 4.19.31. and return 0 in kernel 4.19.32

Shao-Feng:
Do not merge, I am thinking if this change will impact our dynamic switching feature. when Early EVS is set as master, maybe 13 is also the errno. Then HWC will fail to set as master any more.

Shao-Feng:
Updated for review.
Use the method drmAuthMagic to check if hwc is already DRM master
or not. we ever think to check the file '/sys/kernel/debug/dri/128/clients'. but this file is only available in eng and userdebug image. so the IOCTL drmAuthMagic  is selected to check if hwc is already DRM master.

Shao-Feng:
work well on GP too

PR #584, opened 2019-05-15
When damage is same with display, keep it unchanged

Shao-Feng:
In some case, the damage is out of source_crop and same with
display_frame. then the damage should not be transformed.

Change-Id: I11a9575e9c9e41fb2e1e9f0af09c18c89856c069
Tests: work well with customer's OpenGL App.
Tracked-On: https://jira.devtools.intel.com/browse/OAM-80358

### DIFF
--- a/common/core/hwclayer.cpp
+++ b/common/core/hwclayer.cpp
@@ -223,6 +223,12 @@ int32_t HwcLayer::GetAcquireFence() {
 
 void HwcLayer::SufaceDamageTransfrom() {
   int ox = 0, oy = 0;
+
+  if (surface_damage_ == display_frame_) {
+    current_rendering_damage_ = surface_damage_;
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


common/core/hwclayer.cpp
```diff
@@ -223,6 +223,12 @@ int32_t HwcLayer::GetAcquireFence() {
 
 void HwcLayer::SufaceDamageTransfrom() {
   int ox = 0, oy = 0;
+
+  if (surface_damage_ == display_frame_) {
```
js0701: I believe this may result in some bugs?
Suppose a source crop is larger than display_frame_ case,
partial damage can result in surface_damage equal to display_frame?
But it still need transform

Shao-Feng: OK, I will add a status in SetSurfaceDamage to avoid transform if no damage is sent by invokeing this method.

PR #587, opened 2019-05-18
Fix compile error in GbmBufferHandler.

YuanjunHuang:
The error is brought in by referencing wrong version of gbm.h API
when implementing RBC in linux platform. It should reference the
gbm.h API from minigbm.

Test: Compile successfully on Linux based on minigbm.
Tracked-On: None

### DIFF
--- a/common/core/overlaylayer.cpp
+++ b/common/core/overlaylayer.cpp
@@ -571,7 +571,7 @@ void OverlayLayer::ValidatePreviousFrameState(OverlayLayer* rhs,
 
   if (!layer->HasVisibleRegionChanged() && !content_changed &&
       surface_damage_.empty() && !layer->HasLayerContentChanged() &&
-      !(state_ & kNeedsReValidation)) {
+      !(state_ & kNeedsReValidation) && !force_content_changed_) {
     state_ &= ~kLayerContentChanged;
   }
 }
--- a/common/core/overlaylayer.h
+++ b/common/core/overlaylayer.h
@@ -123,11 +123,17 @@ struct OverlayLayer {
   }
 
   const HwcRect<int>& GetSurfaceDamage() const {
-    return surface_damage_;
+    if (force_content_changed_)
+      return display_frame_;
+    else
+      return surface_damage_;
   }
 
   HwcRect<int>& GetSurfaceDamage() {
-    return surface_damage_;
+    if (force_content_changed_)
+      return display_frame_;
+    else
+      return surface_damage_;
   }
 
   uint32_t GetSourceCropWidth() const {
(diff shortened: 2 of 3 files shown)


YuanjunHuang:
Hold and will discuss which gbm should be based for Linux platform, minigbm or from Mesa.

js0701:
Both Linux and Android should use Minigbm.
Why we git it compiled on Android but failure on Linux

YuanjunHuang:
> Both Linux and Android should use Minigbm.
> Why we get it compiled on Android but failure on Linux

For Android the buffer handler is based on gralloc rather than gbm API/gbm.h.  And gbm.h in minigbm is outdated. 

Why for Linux/desktop case, we should use Minigbm rather than gbm in Mesa?

js0701:
This is for the consistent between Linux and Android. Android is using gralloc based on Minigbm.

YuanjunHuang:
> This is for the consistent between Linux and Android. Android is using gralloc based on Minigbm.

ok. then we need this patch.

tpalli:
Alternatively you could consider patching minigbm to have identical API for gbm. Reason is that gbm is the 'standard' API while minigbm is Google's own playground.

YuanjunHuang:
> Alternatively you could consider patching minigbm to have identical API for gbm. Reason is that gbm is the 'standard' API while minigbm is Google's own playground. I'm not sure if Google has plans to utilize gbm but for using hwcomposer on standard Linux desktop, a real gbm would be better choice as that one is available on Linux distros.

thanks for your input Tapani. this is not just about patching the gbm API. but the API implementation is in huge gap between in minigbm and in Mesa. this is the same for our minigbm repo and Google upstream minigbm repo.

tpalli:
Yeah unfortunately there's a big gap, my proposal was trying to patch minigbm (case by case) and get google accept those changes, this would bring minigbm closer to gbm. As example in this case at least some of the changes could be simply fixed by changing the function signature. In longer terms this has the benefit that it would be easier to integrate with new features that land to upstream gbm, otherwise minigbm has to duplicate all the work gbm does. Just a suggestion!

YuanjunHuang:
> Yeah unfortunately there's a big gap, my proposal was trying to patch minigbm (case by case) and get google accept those changes, this would bring minigbm closer to gbm. As example in this case at least some of the changes could be simply fixed by changing the function signature. In longer terms this has the benefit that it would be easier to integrate with new features that land to upstream gbm, otherwise minigbm has to duplicate all the work gbm does. Just a suggestion!

Got your point. I am thinking why google calls it 'mini' is that google simplified the implementation architecture we can see from code. but anyway the API should keep the same with GBM from mesa. Suggest we can put this into our task backlog and check in this patch first.

tpalli:
Sure going forward with this patch is fine to me. Maybe should initiate some discussion with google guys about future and plans of minigbm at some point.

PR #588, opened 2019-05-18
Disable all not-in-used plane once DRM Master is reset

Shao-Feng:
A workaround for the plane refresh which is involved by
earlyEvs and AOSP EVS use different plane.

Change-Id: Ia2b28303ea0eaa6deda46e6d3a8d7a04674ea394
Tests: Work well with Kmscube testing APP.
Tracked-On: None

### DIFF
--- a/wsi/drm/drmdisplay.cpp
+++ b/wsi/drm/drmdisplay.cpp
@@ -16,6 +16,7 @@
 
 #include "drmdisplay.h"
 
+#include <sys/time.h>
 #include <cmath>
 #include <limits>
 #include <set>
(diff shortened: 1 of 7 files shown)


Shao-Feng:
Do not review/merge until EarlyEvs confirm the requirement

Shao-Feng:
Need to updated based on first committing trace patch

Shao-Feng:
Fixed linux build error

Shao-Feng:
This patch make HWC robust to work with other DRM App, please help to review

wsi/drm/drmdisplay.cpp
```diff
@@ -510,6 +510,9 @@ bool DrmDisplay::Commit(
     return false;
   }
 
+  // Disable not-in-used plane once DRM master is reset
```
js0701: It is better we add a judge before this ResetPlanes call
instead of hiding the judge in ResetPlanes

Shao-Feng:
Updated for review comments

PR #594, opened 2019-05-29
support Iris driver

tpalli:
Patch forces compositor to disable sync if this was explicitly wanted
during build since setting property 'board.disable.explicit.sync' does
not work as expected.

Jira: None
Test: Compile and run sucessfully with Iris driver on Android.

### DIFF
--- a/Android.common.mk
+++ b/Android.common.mk
@@ -139,9 +139,20 @@ LOCAL_C_INCLUDES += \
 	$(LOCAL_PATH)/common/compositor/vk \
 	$(LOCAL_PATH)/../mesa/include
 else
+
+# iris driver flags
+ifneq ($(filter iris, $(BOARD_GPU_DRIVERS)),)
+$(warning "Iris driver not fully supported, instability or lower performance may occur!")
+LOCAL_CPPFLAGS += \
+	-DUSE_GL \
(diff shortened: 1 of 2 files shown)


js0701:
Hi Tapani,

Why Iris can't support explict sync?
Android framework is using explict fence to sync

tpalli:
> Hi Tapani,
> 
> Why Iris can't support explict sync?
> Android framework is using explict fence to sync

It does not seem to work yet (and is currently disabled in driver by not exposing PIPE_CAP_NATIVE_FENCE_FD capability) , I'm assuming this to be a bug in Iris driver, I've tried to fix things but haven't found a cure so far.

tpalli:
Note I'm fine not merging this yet. I'm just basically trying to reduce the amount of diff I have for making Iris to work so that it would be easier to do testing and development as I have changes in multiple repos.

js0701:
OK. I think best way it to fix in Iris i915 so explicit fence can be supported

tpalli:
> OK. I think best way it to fix in Iris i915 so explicit fence can be supported

OK that is fine. I've pushed small changes to minigbm, libdrm (external) and selinux. There are additional changes required in mesa, mixins and also hwcomposer (these). Anyone wanting to test Iris, please contact me and I can help if having problems.

PR #605, opened 2019-06-12
[WA]Force refresh layer content on extend display

Shao-Feng:
In "Presentation with Media Router" test case, the surface-damage
for the changing layer on extend display is 0. For work around it
HWC will mark "content is changed" for all layers on extend display

Change-Id: I048b56e967a6f06ba73eff5ae77c95a55cbd028c
Tests:         Tested on Android Q
Tracked-On:    None

### DIFF
--- a/common/core/overlaylayer.cpp
+++ b/common/core/overlaylayer.cpp
@@ -571,7 +571,7 @@ void OverlayLayer::ValidatePreviousFrameState(OverlayLayer* rhs,
 
   if (!layer->HasVisibleRegionChanged() && !content_changed &&
       surface_damage_.empty() && !layer->HasLayerContentChanged() &&
-      !(state_ & kNeedsReValidation)) {
+      !(state_ & kNeedsReValidation) && !force_content_changed_) {
     state_ &= ~kLayerContentChanged;
   }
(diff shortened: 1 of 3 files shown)


js0701:
How is surface_damage been processed in this situation?
I see you make the state content changed, but what is the values of surface damage

Shao-Feng:
In the special case, the empty surface_damage will be changed as minus value in top or left. and will not be empty.
The change in the state is for judge if the surfacedamage need to be transformed again for rotation in Overlaylayer. In Overlaylayer, the state will be check for avoiding to rotate surface_damage for empty array case. so the state can not be removed after transformed in hwclayer

Shao-Feng:
No surface_damage is sent from SF after the OpenGL application is running. just 0000 are sent only 1 time when launching the application.
So the surface damage is always 0000 during executing this app.
With this workaround, whether if the surface_damage is empty or not, the layer on extend display will always be refreshed

Shao-Feng:
this workaround only work for the case "No composition is needed" for the layer. I will check if we need to cover the composition case.

Shao-Feng:
Updated for composition

PR #608, opened 2019-06-18
Force commit when surface damage for layer is 0000

Shao-Feng:
The behavior of SF is changed. the empty of damage array are not
sent for sleeping(s3). Add a workaround to avoid flicking

Change-Id: I4c38ebe570be5ffe2c7dc206345d75d95ea45823
Tests: Work well on Android Q
Tracked-On: https://jira.devtools.intel.com/browse/OAM-83039

### DIFF
--- a/common/display/displayqueue.cpp
+++ b/common/display/displayqueue.cpp
@@ -437,7 +437,8 @@ void DisplayQueue::GetCachedLayers(const std::vector<OverlayLayer>& layers,
         reset_composition_regions = true;
 
       last_plane.SetOverlayLayer(layer);
-      if (layer->HasLayerContentChanged()) {
+      if (layer->HasLayerContentChanged() ||
+          layer->GetSurfaceDamage().empty()) {
         ignore_commit = false;
       }
 


common/display/displayqueue.cpp
```diff
@@ -437,7 +437,8 @@ void DisplayQueue::GetCachedLayers(const std::vector<OverlayLayer>& layers,
         reset_composition_regions = true;
 
       last_plane.SetOverlayLayer(layer);
-      if (layer->HasLayerContentChanged()) {
+      if (layer->HasLayerContentChanged() ||
+          layer->GetSurfaceDamage().empty()) {
```
js0701: @Shao-Feng 
Can't we change HasLayerContentChanged() function to add this WA?
It will be better I think

Shao-Feng: I hope we don't need to merge this one. Just commit here for tracing the bug. I hope to find out the root cause why screen is flicking  after the unchanged frame-commit is ignored. The logic should be correct. No need to commit for the un-changed frame before sleeping. Unless the buffer of layer is changed in turning to sleep. but SF did not inform us

Shao-Feng:
Here is the commit in SF for changing SurfaceDamage logic
commit 8366f563a621981be0a83b675cbbc714c6d98cd1
Author: Yichi Chen <yichichen@google.com>
Date:   Mon Mar 25 19:44:06 2019 +0800

    HWC2: Optimize execution by caching all HWC2 calls in setPerFrameData
    
    To avoid redundant HWC2 calls into hwcomposer service, we cache all HWC2
    calls in BufferLayer::setPerFrameData by having visible region, damage
    region and layer buffers cached in HWC2::Layer.
    
    Bug: 119414178
    Test: Display function test on F2.
    Test: b/129317072 Display function tests on M1/S1, M2/S2, B1/C1, B4/S4,
          and C2/F2.
    Change-Id: I1679e755ac55a459c93b42b9f8225f0c270228e6

Shao-Feng:
Not only SurfaceDamage, it cache all HWC2 calls in setPerFrameData

common/display/displayqueue.cpp
YuanjunHuang: Is it possible the issue is in HasLayerContentChanged()? since SF use cache call in new AOSP, after resume from S3, layer cached is cleared, but hwc still think the layer is not changed? just a guess.

Shao-Feng: The flicking happen before powermode is set as kOff. not in resuming.

Shao-Feng:
This issue is fixed in AOSP. this one is not needed

PR #80, opened 2017-06-01
Support for MultiMode Display Config 

pyn1:
Added Test case for MultiMode Display support

Jira:IAHWC-34
Tests: Able to set displaymode as per user choice
Signed-off-by: N, Poornima Y <poornima.y.n@intel.com>

PR #610, opened 2019-06-21
Fix klocwork issues in GFX domain base on #80 daily scan

yuanzhel:
Revert an old fix because it dosen't pass kw scan.
Add more initialization statements and resource free functions.


yugang:
LGTM

PR #613, opened 2019-07-06
Filter EDIP for prefer and performance config

Shao-Feng:
Asked by AOSP, only 2 display config should be return to SF
one is the preferred config. and the other is optional better
performance config.

Change-Id: I966bbdeb819e54b12a5ef37103e593207b95a940
Tests: Work well on Android Q GP
Tracked-On: https://jira.devtools.intel.com/browse/OAM-83748

### DIFF
--- a/wsi/drm/drmdisplay.cpp
+++ b/wsi/drm/drmdisplay.cpp
@@ -409,6 +409,64 @@ bool DrmDisplay::GetDisplayAttribute(uint32_t config /*config*/,
   return status;
 }
 
+uint32_t DrmDisplay::FindPreferedDisplayMode(size_t modes_size) {
+  uint32_t prefer_display_mode = 0;
+  if (modes_size > 1) {
+    SPIN_LOCK(display_lock_);
+    for (size_t i = 0; i < modes_size; i++) {
+      // There is only one preferred mode per connector.
+      if (modes_[i].type & DRM_MODE_TYPE_PREFERRED) {
+        prefer_display_mode = i;
+        IHOTPLUGEVENTTRACE("Preferred display config is found. index: %d", i);
+        break;
+      }
+    }
+    SPIN_UNLOCK(display_lock_);
+  }
+  if (prefer_display_mode_ != prefer_display_mode)
+    prefer_display_mode_ = prefer_display_mode;
+  return prefer_display_mode;
+}
+
+uint32_t DrmDisplay::FindPerformaceDisplayMode(size_t modes_size) {
+  uint32_t perf_display_mode;
+  perf_display_mode = prefer_display_mode_;
+  if (modes_size >= prefer_display_mode_) {
+    int32_t prefer_width = 0, prefer_height = 0, prefer_interval = 0;
+    int32_t perf_width = 0, perf_height = 0, perf_interval = 0,
+            previous_perf_interval = 0;
+    GetDisplayAttribute(prefer_display_mode_, HWCDisplayAttribute::kWidth,
+                        &prefer_width);
+    GetDisplayAttribute(prefer_display_mode_, HWCDisplayAttribute::kHeight,
+                        &prefer_height);
+    GetDisplayAttribute(prefer_display_mode_, HWCDisplayAttribute::kRefreshRate,
+                        &prefer_interval);
+    previous_perf_interval = prefer_interval;
+    IHOTPLUGEVENTTRACE("Preferred width:%d, height:%d, interval:%d",
+                       prefer_width, prefer_height, prefer_interval);
+    for (size_t i = 0; i < modes_size; i++) {
+      if (i != prefer_display_mode_) {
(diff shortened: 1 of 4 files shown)


dyang23:
Review +1

jennycao:
@Shao-Feng with this patch, 
Do we still need the WA patch 
Limits some monitors' modes due to hardware limitation. https://github.com/intel/IA-Hardware-Composer/commit/bee374610533f2766de31870026a557bb0fc6357

Shao-Feng:
In most case, only the prefer display mode will be sent to SF. just in some special case, the performance display mode is different with the preferred one. For the patch, bee3746 it work for the display "TD2230 Series", from the existing mode list of this display, we will also only send the preferred display mode to SF too. but not sure if the display mode list will be changed, and a unsupported EDIP item will be selected as "performance" mode. It should be OK to revert it for current display. but not 100% sure it work in future/with the updated display.

wsi/drm/drmdisplay.cpp
```diff
@@ -409,6 +409,59 @@ bool DrmDisplay::GetDisplayAttribute(uint32_t config /*config*/,
...
+      }
+    }
+    SPIN_UNLOCK(display_lock_);
+  }
+  return prefer_display_mode_;
+}
+
+uint32_t DrmDisplay::GetPerformaceDisplayMode(size_t modes_size,
+                                              uint32_t prefer_display_mode) {
+  perf_display_mode_ = prefer_display_mode;
```
js0701: It is bad coding here...
since perfer_display_mode_ is a member of this class, you don't need to pass and assign it everywhere

wsi/drm/drmdisplay.cpp
```diff
@@ -409,6 +409,59 @@ bool DrmDisplay::GetDisplayAttribute(uint32_t config /*config*/,
...
+  if (modes_size >= prefer_display_mode) {
```
js0701: since it written this way, for else situation, perf_display_mode is not set, right?

js0701: Suppose perf_mode_mode should be set a default value prefer_display_mode

js0701:
Please address comments

wsi/drm/drmdisplay.h
```diff
@@ -171,6 +171,10 @@ class DrmDisplay : public PhysicalDisplay {
 
   void TraceFirstCommit();
 
+  uint32_t GetPreferedDisplayMode(size_t modes_size);
```
js0701: Need to be private?

wsi/drm/drmdisplay.cpp
```diff
@@ -409,6 +409,59 @@ bool DrmDisplay::GetDisplayAttribute(uint32_t config /*config*/,
   return status;
 }
 
+uint32_t DrmDisplay::GetPreferedDisplayMode(size_t modes_size) {
+  if (modes_size > 1) {
+    SPIN_LOCK(display_lock_);
+    for (size_t i = 0; i < modes_size; i++) {
+      // There is only one preferred mode per connector.
+      if (modes_[i].type & DRM_MODE_TYPE_PREFERRED) {
+        prefer_display_mode_ = i;
```
shuangwan01: Suggest don't update internal state in GetPreferedDisplayMode since the method name is more likely a read only method.

wsi/drm/drmdisplay.cpp
```diff
@@ -409,6 +409,59 @@ bool DrmDisplay::GetDisplayAttribute(uint32_t config /*config*/,
...
+        GetDisplayAttribute(i, HWCDisplayAttribute::kWidth, &perf_width);
+        GetDisplayAttribute(i, HWCDisplayAttribute::kHeight, &perf_height);
+        GetDisplayAttribute(i, HWCDisplayAttribute::kRefreshRate,
+                            &perf_interval);
+        IHOTPLUGEVENTTRACE("EDIP item width:%d, height:%d, rate:%d", perf_width,
+                           perf_height, perf_interval);
+        if (prefer_width == perf_width && prefer_height == perf_height &&
+            prefer_interval > perf_interval &&
+            previous_perf_interval > perf_interval) {
+          perf_display_mode_ = i;
```
shuangwan01: Suggest don't update internal state in GetPerformaceDisplayMode since the method name is more likely a read only method.

wsi/drm/drmdisplay.h
Shao-Feng: No need to invoke them out of this class, so set them as private method

wsi/drm/drmdisplay.cpp
Shao-Feng: I will change the name

Shao-Feng:
Updated for reviewing comments

Shao-Feng:
Add a commit to revert the Workaround for ViewSonic display

PR #614, opened 2019-07-08
Add power mode set when connecting display even in connected state.

yugang:
This is WA as APL NUC always triggers hotplug when resume from S3,
which causes the postpone mode set be ingnored and display can't
be resumed from blackscreen.

Change-Id: I49e66552b373f1adde7aa6d04a19c508adeab991
Test: Display can be resumed from S3 successfully.
Tracked-On: OAM-81142
Sighed-off-by: Yugang, Fan <yugang.fan@intel.com>

### DIFF
--- a/wsi/physicaldisplay.cpp
+++ b/wsi/physicaldisplay.cpp
@@ -152,6 +152,10 @@ void PhysicalDisplay::Connect() {
 
   SPIN_LOCK(modeset_lock_);
   if (connection_state_ & kConnected) {
+    IHOTPLUGEVENTTRACE(
+        "PhysicalDisplay::Connect connected already, return with power mode "
+        "update.");
+    UpdatePowerMode();
     SPIN_UNLOCK(modeset_lock_);
     return;
   }


yugang:
for some corner cases, S3 resume will trigger hotplug, then power mode set will postphone the mode on setting. but in display connect calling for this hotplug, it will return early without power mode setting, which causes the blackscreen and can't resume from S3.

by the way, this hotplug is different with other platforms as other platforms' S3 resume has no hotplug, so no mode postphone. we add this as code logic is right, and also can fix the issue as WA. but for long term, still need power management to check why there is a hotplug.

@js0701 @Shao-Feng @YuanjunHuang @shuangwan01 ....pls help review, thanks

wsi/physicaldisplay.cpp
```diff
@@ -152,6 +152,8 @@ void PhysicalDisplay::Connect() {
 
   SPIN_LOCK(modeset_lock_);
   if (connection_state_ & kConnected) {
+    IHOTPLUGEVENTTRACE("PhysicalDisplay::Connect connected already, return with power mode update.");
+    UpdatePowerMode();
```
shuangwan01: LGTM

PR #621, opened 2019-07-12
Use SurfaceFlinger to do client composite if layers more than 3

dyang23:
If layers more than 3, layer1 and layer2 set to Device composite
type, and other layers set to Client composite type.

Change-Id: I6c32a5d5b3cdd31feb113288fba4afc11842b9b5
Tracked-On:

### DIFF
--- a/common/core/hwclayer.cpp
+++ b/common/core/hwclayer.cpp
@@ -422,6 +422,14 @@ bool HwcLayer::IsCursorLayer() const {
   return is_cursor_layer_;
 }
 
+void HwcLayer::MarkAsVideoLayer() {
+  is_video_layer_ = true;
+}
+
+bool HwcLayer::IsVideoLayer() const {
+  return is_video_layer_;
+}
+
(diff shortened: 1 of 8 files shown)


js0701:
Do we consider video need to be a separated plane, that is not client type here?

js0701:
Thanks for this patch
I generally agree with it.
Beyond the comment, just a few things
1) Explain what is the benefit to return client type 
    say, junky frame rate how much reduce
2) Also you can explain the video layer logic, that is we wish to choose original way when we meet video, the HWC will try to separate planes or use VPP for video
3) Add a switch to choose the old way, it will help us to do debug and analysis

dyang23:
If any video layer included, we will keep original logic, that is all the layer will set it as DEVICE type, and how to mapping to plane will handled in present() process.

dyang23:
patch update, please review again.

js0701:
@dyang23 
OK. I will approve. But it just has a not necessary commit? I guess you just need one commit

dyang23:
@js0701 Yes, the not used commit already removed, please review again, thanks.

PR #622, opened 2019-07-24
Initialize SurfaceDamage with display_frame_ by default

Shao-Feng:
From Q, AOSP will not set the surfaceDamage for each frame. and some
times, surface_damage will not be set by SF. So has to set SurfaceDamage
with display_frame_ by default. and always do 'Transform' for
surfaceDamage once it is set.
Another change, when surfaceDamage is empty from SF, surface_damage will
be assigned with source_crop for 'Transform'.

Change-Id: I608c9b985ba8575280170ac886f2c508d181574f
Tests:      Work well on Android Q
Tracked-On: https://jira.devtools.intel.com/browse/OAM-84227

### DIFF
--- a/common/core/hwclayer.cpp
+++ b/common/core/hwclayer.cpp
@@ -118,9 +118,7 @@ void HwcLayer::SetSurfaceDamage(const HwcRegion& surface_damage) {
       return;
     }
   } else if (rects == 0) {
-    rect = display_frame_;
-    // damage is assigned with display_frame, no need to transform
-    state_ &= ~kSurfaceDamageChanged;
+    rect = source_crop_;
   }
 
   if ((surface_damage_.left == rect.left) &&
@@ -280,7 +278,7 @@ void HwcLayer::SufaceDamageTransfrom() {
   } else if (surface_damage_.empty()) {
     current_rendering_damage_ = surface_damage_;
   } else {
-    current_rendering_damage_ = translated_damage;
+    current_rendering_damage_ = display_frame_;
   }
 }
 
@@ -295,6 +293,8 @@ void HwcLayer::Validate() {
     layer_cache_ &= ~kSourceRectChanged;
     if (state_ & kSurfaceDamageChanged) {
       SufaceDamageTransfrom();
+    } else {
+      current_rendering_damage_ = display_frame_;
     }
   }
 
--- a/common/core/overlaylayer.cpp
+++ b/common/core/overlaylayer.cpp
@@ -187,7 +187,7 @@ void OverlayLayer::ValidateTransform(uint32_t transform,
 void OverlayLayer::TransformDamage(HwcLayer* layer, uint32_t max_height,
                                    uint32_t max_width) {
   const HwcRect<int>& surface_damage = layer->GetLayerDamage();
-  if (surface_damage.empty() || !layer->HasSurfaceDamageRegionChanged()) {
+  if (surface_damage.empty()) {
     surface_damage_ = surface_damage;
     return;
   }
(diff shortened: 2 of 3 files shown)


Shao-Feng:
For the case, if the left or top is not 0. On Android Q, the surface-damage is not set correctly from SF.
Below is the dump of SF.
-------------------------------------------------------------------------------------------
 Layer name
           Z |  Window Type |  Comp Type |  Transform |   Disp Frame (LTRB) |          Source Crop (LTRB)
- - - - - - - - - - - - - - - -- - - - - - - - - - - - - - - -- - - - - - - - - - - - - - -
 com.example.android.apis/com.example.android.apis.ApiDemos#0
  rel      0 |            1 |     DEVICE |          0 |    8   76 1913 1024 |    0.0   76.0 1905.0 1024.0
- - - - - - - - - - - - - - - -- - - - - - - - - - - - - - - -- - - - - - - - - - - - - - -
 StatusBar#0
  rel      0 |         2000 |     DEVICE |          0 |    0    0 1920   76 |    0.0    0.0 1920.0   76.0

the source_crop is (0, 76, 1905, 1024), displayframe is (8, 76, 1913,1024)
the surface_damage from SF is below
07-23 07:57:16.823  2841  2841 E iahwcomposer: SetLayerSurfaceDamage: ==========hwc_region[1]
07-23 07:57:16.824  2841  2841 E iahwcomposer: SetLayerSurfaceDamage: ==========hwc_region[0] (LTRB): 0, 0, 1905, 1024

we can see, the surface damage from SF is not correct. 
has to assign display_frame to surface_damage for this case

js0701:
I think we have file bug to AOSP right? if haven't do so, why not

Shao-Feng:
We have a bug for the surfaceDamage is never set by SF. and AOSP provide a diff for it. but this diff is not accepted by Google. so AOSP team hope to revert it. and AOSP team check the behavior on pixel. on pixel, the behavior  with the same SF is different with us. so they hope we can set the surfaceDame with displayframe by default if SF did not set

Shao-Feng:
For the incorrect surfaceDamage size when the left or top is not 0, below comment is included in Android P release
https://github.com/intel/IA-Hardware-Composer/blob/master/common/core/hwclayer.cpp#L231
It seems we already handle it with WA. but now, the surfacedamege could be assigned with Source_Crop, so assign display_frame here for this case

common/core/hwclayer.cpp
```diff
@@ -118,9 +118,7 @@ void HwcLayer::SetSurfaceDamage(const HwcRegion& surface_damage) {
       return;
     }
   } else if (rects == 0) {
-    rect = display_frame_;
-    // damage is assigned with display_frame, no need to transform
-    state_ &= ~kSurfaceDamageChanged;
+    rect = source_crop_;
```
js0701: Does this WA actually disable partial update?

Shao-Feng: No, partial update are only disabled when the left and top of source_crop are not 0.
Partial-refresh still work when the surface-damage is set and source_crop.left and top are 0.

Previously, HWC suppose SF send the 3 attribute, surface_damage, source_crop, and display_frame for each layer in each frame.
From Q, it is not true. SF send a layer only with source_crop and display_frame. surface_damage could not be set by SF.
So use the flag kSurfaceDamageChanged to mark if the surface_damage is set or not, if it is not set, assigned with display_frame. 

Once it is set with a non-empty array, use this surface-damage if source_crop.top and left are 0 (Most case, the source_crop are 0,0, X, X). if source_crop.top and left are not 0, use WA

Once it is set with a empty array, it mean the whole layer need to be updated. assigned with source_crop here for transforming

PR #623, opened 2019-07-25
Skip Client layer without buffer

Shao-Feng:
The Client layer composed by SF could be sent to HWC.
and trigger a error log below
HWC don't support layer without buffer if not in type SolidColor

Change-Id: I76bb79520453c68a0b17fbf69db0a151d993bbb4
Tests: Work on Android Q
Tracked-On: None

### DIFF
--- a/os/android/iahwc2.cpp
+++ b/os/android/iahwc2.cpp
@@ -637,6 +637,14 @@ HWC2::Error IAHWC2::HwcDisplay::PresentDisplay(int32_t *retire_fence) {
       cursor_z_order = l.second.z_order();
       continue;
     }
+#ifndef FORCE_ALL_DEVICE_TYPE
+    if ((l.second.validated_type() == HWC2::Composition::Client) &&
+        (l.second.GetLayer()->GetNativeHandle() == NULL)) {
+      ICOMPOSITORTRACE(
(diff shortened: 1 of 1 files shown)


os/android/iahwc2.cpp
```diff
@@ -637,6 +637,14 @@ HWC2::Error IAHWC2::HwcDisplay::PresentDisplay(int32_t *retire_fence) {
       cursor_z_order = l.second.z_order();
       continue;
     }
+#ifndef FORCE_ALL_DEVICE_TYPE
+    if ((l.second.validated_type() == HWC2::Composition::Client) &&
```
YuanjunHuang: do you know in what case and why this situation will be met?

js0701: it is a MACRO added by Dong. Default should be force

js0701: @dyang23  Please also help to review

Shao-Feng: It is SurfaceFlinger's behavior. Now, we can meet it in Home and App page. I suppose when SF do composition for Client layer, this issue could happen.

Shao-Feng: Till now, this MACRO is not defined. it is not force all layer "Device". I will prepare one patch for it

dyang23:
LGTM

PR #628, opened 2019-08-01
SolidColor need to be handled as Device layer

Shao-Feng:
If the z_order of SolidColor is lower than other Device layer.
and overlap with other Layer. the overlaped Device layer will be
covered by the SolidColor if SolidColor.alpha is not 0.
So need to handle SolidColor as same as Device layer

Change-Id: I653e0e519fdeac161ce86303fbd1668ba840cb16
Tests: Work well on Android Q.
Tracked-On: None


dyang23:
LGTM

PR #634, opened 2019-08-21
Re-invoke SetActiveConfig for Hotplug to initialize ClientLayer

Shao-Feng:
SetActiveConfig is not invoked by SF when hotplug even the display
is not primary. HWC must invoke it for initialize the clientLayer.

Change-Id: I2e4a2321d28f59c718db8c743c9513c469066f68
Tests: Work well for HotPlug on Android Q
Tracked-On: None


Shao-Feng:
@jennycao this patch should fix the missing client layer in hotplug

dyang23:
nice catch, LGTM.

PR #635, opened 2019-08-23
Update ValidateLayers for only 1 plane case

Shao-Feng:
Only 1 plane is available for HWC on Kvm.
Update the logic to
1/ Non-video case and Client is used, all layers are Client. just need
to scan out the Client layer
2/ Non-video case and Client is not used, force to compose all layers
by GPU
3/ Video case, Force to compose all layers in VPP
For video case, still meet surface issue (Fail to allocate surface).
Not sure if it is limitation in layer size of VPP. will fix later.

Change-Id: Iabc1fdb0fa8faaca3f49376177eab631d0c571ca
Tests: PlaneReservation work for reserve 1 plane on Android Q
Tracked-On: https://jira.devtools.intel.com/browse/OAM-84953

### DIFF
--- a/common/Android.mk
+++ b/common/Android.mk
@@ -132,6 +132,8 @@ LOCAL_CPPFLAGS += \
 	-DLOCK_DIR_PREFIX='"/vendor/etc"' \
         -DHWC_DISPLAY_INI_PATH='"/vendor/etc/hwc_display.ini"' \
         -DKVM_HWC_DISPLAY_INI_PATH='"/vendor/etc/hwc_display.kvm.ini"' \
+        -DKVM_HWC_PROPERTY='"ro.graphics.hwcomposer.kvm"' \
+        -DUSE_ANDROID_PROPERTIES \
         -D_GNU_SOURCE=1 -D_FILE_OFFSET_BITS=64 \
         -Wno-unused-parameter \
         -O3
--- a/common/Android.mk
+++ b/common/Android.mk
@@ -131,6 +131,7 @@ LOCAL_CPPFLAGS += \
         -Wcast-qual -Wcast-align \
 	-DLOCK_DIR_PREFIX='"/vendor/etc"' \
         -DHWC_DISPLAY_INI_PATH='"/vendor/etc/hwc_display.ini"' \
+        -DKVM_HWC_DISPLAY_INI_PATH='"/vendor/etc/hwc_display.kvm.ini"' \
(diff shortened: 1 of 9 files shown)


Shao-Feng:
Add a patch for reading KVM property and loading KVM config
1 issue still remain, deployment script for deploying the new configuration file  "hwc_display.kvm.ini" need to be updated in "device/intel/" for all platform

common/Android.mk
```diff
@@ -131,6 +131,8 @@ LOCAL_CPPFLAGS += \
         -Wcast-qual -Wcast-align \
 	-DLOCK_DIR_PREFIX='"/vendor/etc"' \
         -DHWC_DISPLAY_INI_PATH='"/vendor/etc/hwc_display.ini"' \
+        -DKVM_HWC_DISPLAY_INI_PATH='"/vendor/etc/hwc_display.kvm.ini"' \
```
YuanjunHuang: What about still using original hwc_display.ini, after parsing it, then check if KVM platform to adjust the available planes? This will introduce less modification, no need an extra config file. I suppose you are trying to make it general for specific platform/use case, but currently there is no other config requests than plane limitation right?

YuanjunHuang:
what is the behaviour in case: there is video layer + only one plane limitation?
It is a known issue right?

common/Android.mk
Shao-Feng: Yes, I think so too. but After discussing with Di, She is worrying if we need to hard code the plane limitation for KVM. On QEMU, hwc can not use the real hardware planes in current implementation, before committing to the real hardware plane, QEMU need to compose all planes, so She think it will be better to render on only one plane. but in the future version, it is possible that hwc can commit to the real hardware plane directly (maybe share with Linux Host), So she suggest to add a new config for KVM.

Shao-Feng:
For 1plane and video case, all layers will be composed by VPP.
I am still in debugging. it seems it is not the issue in VPP. the current Surface recycling logic have problem to run with only 1 plane if the surface need to be recycled. the surface issue also happen in the case 1Plane + disabling Client layer.

Shao-Feng:
The deploy script is updated on https://android.intel.com/#/c/679215/1
please help to review, @jennycao

Shao-Feng:
surface issue is fixed in https://github.com/intel/IA-Hardware-Composer/pull/636

YuanjunHuang:
LGTM.

PR #651, opened 2019-09-17
Update the alpha blending logic for composing.

Shao-Feng:
When layer alpha become 0xff(blending is Premult), re-validate for
composing or scanout. when the alpha(Premult) become 0, still need
to compose, otherwise, a alpha 255 surface will be shown(for alpha=0)
in scanout.

Change-Id: I97bd1321c5122fc05d7b68cbd23c4fc1692ea7d4
Tests: Work well on Android Q
Tracked-On: https://jira.devtools.intel.com/browse/OAM-85651


Shao-Feng:
This patch also work for the black line at the end of video playing

PR #652, opened 2019-09-17
Fixed the dmabuf leak issue caused by prime FDs are not released.

shuangwan01:
Release prime FDs if the buffer needs to be released.

Change-Id: Id9381a2fbd7681aca71a5e6ab387713f5b644e31
Tracked-On: https://jira.devtools.intel.com/browse/OAM-82969
Tests: HWC virtual memory doesn't inrease abnormally.

### DIFF
--- a/common/display/virtualdisplay.cpp
+++ b/common/display/virtualdisplay.cpp
@@ -27,6 +27,7 @@
 #include "hwctrace.h"
 #include "overlaylayer.h"
 
+#include "gpudevice.h"
 #include "hwcutils.h"
 
 namespace hwcomposer {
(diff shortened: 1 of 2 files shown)


common/display/virtualdisplay.cpp
```diff
@@ -308,6 +309,13 @@ bool VirtualDisplay::Present(std::vector<HwcLayer *> &source_layers,
           }
           mHyperDmaExportedBuffers.erase(search);
         }
+
+        FrameBufferManager *fb_manager_ =
+            GpuDevice::getInstance().GetFrameBufferManager();
+
+        fb_manager_->RemoveFB(handle.handle_->meta_data_.num_planes_,
```
MicheleL-Intel: Suggesting checking fb_manager_ is not NULL before referencing it.

shuangwan01: Yes, make sense. Will update the PR accordingly.

PR #653, opened 2019-09-17
Fixed the HWC crash due to messed virtual display handle.

shuangwan01:
The virtual display handle tracked in map might be messed due to
newly created display instance might overwrite exiting one.

Tracked-On: None
Tests: HWC doesn't crash when create/destroy VDS frequently.

### DIFF
--- a/os/android/iahwc2.cpp
+++ b/os/android/iahwc2.cpp
@@ -230,14 +230,42 @@ static inline void supported(char const *func) {
 HWC2::Error IAHWC2::CreateVirtualDisplay(uint32_t width, uint32_t height,
                                          int32_t *format,
                                          hwc2_display_t *display) {
-  *display = (hwc2_display_t)(virtual_display_index_ + HWC_DISPLAY_VIRTUAL +
-                              VDS_OFFSET);
+  if (NULL == display) {
+    ALOGE("Pointer of display is NULL");
+    return HWC2::Error::BadDisplay;
+  }
+
+  if (NULL == format) {
+    ALOGE("Pointer of format is NULL");
+    return HWC2::Error::BadParameter;
+  }
+
+  uint32_t free_display_index = 0;
+
+  spin_lock_.lock();
+
+  for (uint32_t disp_index = 0; disp_index < virtual_display_index_;
+       disp_index++) {
+    auto it = virtual_displays_.find(disp_index);
+    if (it == virtual_displays_.end()) {
+      free_display_index = disp_index;
+      break;
+    }
+  }
+  if (!free_display_index) {
+    free_display_index = virtual_display_index_;
+    virtual_display_index_++;
+  }
+
+  *display =
+      (hwc2_display_t)(free_display_index + HWC_DISPLAY_VIRTUAL + VDS_OFFSET);
   std::unique_ptr<HwcDisplay> temp(new HwcDisplay());
-  temp->InitVirtualDisplay(device_.CreateVirtualDisplay(virtual_display_index_),
-                           width, height, virtual_display_index_,
+  temp->InitVirtualDisplay(device_.CreateVirtualDisplay(free_display_index),
+                           width, height, free_display_index,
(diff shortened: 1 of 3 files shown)


os/android/iahwc2.cpp
```diff
@@ -230,19 +230,34 @@ static inline void supported(char const *func) {
 HWC2::Error IAHWC2::CreateVirtualDisplay(uint32_t width, uint32_t height,
                                          int32_t *format,
                                          hwc2_display_t *display) {
-  *display = (hwc2_display_t)(virtual_display_index_ + HWC_DISPLAY_VIRTUAL +
-                              VDS_OFFSET);
+  spin_lock_.lock();
```
MicheleL-Intel: Identify the shared resources utilized within CreateVirtualDisplay(), and limit the lock for the duration of when shared resources are being accessed, unless it is necessary to secure the lock in the entire span of the function.

os/android/iahwc2.cpp
```diff
@@ -230,19 +230,34 @@ static inline void supported(char const *func) {
 HWC2::Error IAHWC2::CreateVirtualDisplay(uint32_t width, uint32_t height,
                                          int32_t *format,
                                          hwc2_display_t *display) {
```
MicheleL-Intel: Ensure *format and *display are not NULL.

os/android/iahwc2.cpp
```diff
@@ -230,19 +230,34 @@ static inline void supported(char const *func) {
...
+       disp_index++) {
+    auto it = virtual_displays_.find(disp_index);
+    if (it == virtual_displays_.end()) {
+      free_display_index = disp_index;
+      break;
+    }
+  }
+  if (!free_display_index) {
+    free_display_index = virtual_display_index_;
+    virtual_display_index_++;
```
MicheleL-Intel: I think virtual_display_index_ is meant to keep track of virtual display count created or destroyed by the corresponding helpers :)    Incrementing it here without actually having a virtual display created may result to instability.  See CreateVirtualDisplay() and DestroyVirtualDisplay() where this index is being maintained.

shuangwan01: This is intended, actually decreasing index at virtual display destroy is the root cause of the live virtual display instance overwritten. This leads into HWC crash. The reason of it is VDS creation & deletion might be interleaved. We applied reusing freed display index slots approach that ensures ACRN guests works as expected since it hard coded just using first few display index slots.

os/android/iahwc2.cpp
shuangwan01: Okay, will update.

os/android/iahwc2.cpp
shuangwan01: Thanks reminds on this. Just shrank the critical region little bits in updated PR.

os/android/iahwc2.cpp
MicheleL-Intel: I see what you're saying and this sounds like it is effective in resolving the crash.  However, what happens if the newly incremented index is referenced while it does not yet have a virtual display attached to it?  Is this a possible scenario?

shuangwan01: Yes, this is possible. Per evaluation, this scenario is not harmful :-)

MicheleL-Intel: Okay, keeping fingers crossed :)  It's still a bit risky using a shared flag (virtual_display_index_) for this purpose since there is always a possibility of a race condition between CreateVirtualDisplay and DestroyVirtualDisplay.  When (free_display_index == 0), could you call IAHWC2::CreateVirtualDisplay() so that it will go thru the normal channel of incrementing virtual_display_index_?

shuangwan01: Don't worry :-) HWC's DestroyVirtualDisplay method will just return if the display is invalid. The underlying caller of CreateVirtualDisplay and DestroyVirtualDisplay is SF. SF only remove the VDS created by itself.

PR #658, opened 2019-10-08
Fix the ICL 3 displays issue

yhe39:
On ICL the 3 displays resolution are different, causing the
clone/extend display mode work abnormal.

Change-Id: I6f48f8add254a60993a582ada8b11dec146ef24e
Tests: 3 display work well on icl.
Tracked-On: https://jira.devtools.intel.com/browse/OAM-86688

### DIFF
--- a/common/display/displayplanemanager.cpp
+++ b/common/display/displayplanemanager.cpp
@@ -240,6 +240,13 @@ bool DisplayPlaneManager::ValidateLayers(
           // it.
           bool fall_back = FallbacktoGPU(plane, layer, composition);
           test_commit_done = true;
+          bool force_separate = false;
+          size_t composition_size = composition.size();
+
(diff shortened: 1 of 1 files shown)


yhe39:
hi @Shao-Feng @YuanjunHuang ,please help to review. Actually the 3 display issue on ICL is regression caused by the change: https://github.com/intel/IA-Hardware-Composer/commit/55c7cb6b016381dc16557d22bf48bff7abb06f2a
i add back some deleted code. 
Thanks!

Shao-Feng:
I suppose it is not the root cause

yhe39:
oK, so you could continue to investigate the root cause @Shao-Feng.
Is it  OK to close this PR? @YuanjunHuang 
Thanks~!

PR #659, opened 2019-10-09
Fix klocwork critical issue

yuanzhel:
Add a null pointer checking before dereference.

Change-Id: I178c0fd9c148e353640f3dfeda7a3b2038e88ad2
Test: Boot successful
Tracked-On: No
Signed-off-by: Yuanzhe Liu <yuanzhe.liu@intel.com>

### DIFF
--- a/common/display/displayplanemanager.cpp
+++ b/common/display/displayplanemanager.cpp
@@ -238,7 +238,9 @@ bool DisplayPlaneManager::ValidateLayers(
 
           // If we are able to composite buffer with the given plane, lets use
           // it.
-          bool fall_back = FallbacktoGPU(plane, layer, composition);
+          bool fall_back = false;
+          if(plane) 
+            fall_back = FallbacktoGPU(plane, layer, composition);
           test_commit_done = true;
           if (fall_back) {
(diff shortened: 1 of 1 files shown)


common/display/displayplanemanager.cpp
```diff
@@ -238,16 +238,18 @@ bool DisplayPlaneManager::ValidateLayers(
...
-          if (fall_back) {
-            ISURFACETRACE(
-                "Force GPU rander the plane[%d], for the layer[%d] isVideo: "
-                "%d, isSolidColor: %d, alpha: %d",
-                last_plane.GetDisplayPlane()->id(), layer->GetZorder(),
-                layer->IsVideoLayer(), layer->IsSolidColor(),
-                layer->GetAlpha());
-            last_plane.ForceGPURendering();
+          if(plane != NULL) {
+            bool fall_back = FallbacktoGPU(plane, layer, composition);
```
Shao-Feng: The scan tool only worry about if plane is null in invoking FallbacktoGPU right? if so, only check null before invoking this method. no need to check for the reset

PR #662, opened 2019-10-12
Force to create GL surface for ForceGpuForAllLayers

Shao-Feng:
If the first layer is video, a nativesurface will be created during
ForceGpuForAllLayers. In ForceGpuForAllLayers, all surface should be
GL surface.

Change-Id: I0cd48ee863329e64afab4c6497fd2f3d52485f50
Tests: Work well on LH 8 video case
Tracked-On: None

### DIFF
--- a/common/display/displayplanemanager.cpp
+++ b/common/display/displayplanemanager.cpp
@@ -467,7 +467,8 @@ void DisplayPlaneManager::SetDisplayTransform(uint32_t transform) {
   display_transform_ = transform;
 }
 
-void DisplayPlaneManager::EnsureOffScreenTarget(DisplayPlaneState &plane) {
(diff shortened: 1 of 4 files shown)


Shao-Feng:
If we can use VPP for force-composing, this patch is not needed. Because we will not use ForceGpuForAllLayers when video layer is included

common/display/displayplanemanager.cpp
```diff
@@ -518,7 +519,7 @@ void DisplayPlaneManager::EnsureOffScreenTarget(DisplayPlaneState &plane) {
 
   if (!surface) {
     NativeSurface *new_surface = NULL;
-    if (video_separate) {
+    if (video_separate && !force_normal_surface) {
```
MicheleL-Intel: Set (force_normal_surface=false) here?  Or does this flag remain in the same state?

PR #663, opened 2019-10-17
[HDR & WCG patch1] Fix videolayer render issues

shuangwan01:
In Linux env, the jsontestlayer expects videolayer renderer calls
which wasn't built.

Change-Id: Iab15dddcaf1214e7ec8ee40485a719328f98a17f
Tests: In Linux environment, testlayers works for bgra.
Tracked-On: None

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -46,6 +46,7 @@ testlayers_SOURCES = \
     ./common/layerrenderer.cpp \
     ./common/gllayerrenderer.cpp \
     ./common/glcubelayerrenderer.cpp \
+	./common/videolayerrenderer.cpp \
     ./common/esTransform.cpp \
     ./common/jsonhandlers.cpp \
     ./apps/jsonlayerstest.cpp
--- a/tests/apps/jsonlayerstest.cpp
+++ b/tests/apps/jsonlayerstest.cpp
@@ -702,7 +702,6 @@ static void init_frames(int32_t width, int32_t height) {
           printf("un-recognized layer type!\n");
           exit(-1);
       }
-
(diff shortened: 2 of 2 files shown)


tests/Makefile.am
```diff
@@ -46,6 +46,7 @@ testlayers_SOURCES = \
     ./common/layerrenderer.cpp \
     ./common/gllayerrenderer.cpp \
     ./common/glcubelayerrenderer.cpp \
+	./common/videolayerrenderer.cpp \
```
MicheleL-Intel: Is this change supposed to be part of another pull request?  Suggest combining it with the corresponding commit.

shuangwan01: Yes, there are a couple of patches need to be reviewed in parallel and there are some commit order dependence among them. The list of PRs for these patches have been committed into my own git repo:
https://github.com/shuangwan01/IA-Hardware-Composer/pulls

Not sure if there is an efficient way to push PRs into this project directly and manage the further revisions accordingly.

MicheleL-Intel: Is it better to combine the patches under one commit?  The specific commit order requirement may be difficult to maintain because only the author and reviewers are familiar with it right? :)

shuangwan01: Yes, agree with you that one commit will make the PR maintenance more easier. But this feature is involving 3 authors and feature itself even doesn't have tested under Android. So it's expected large number of revisions according to real tests/debugs and review comments. Breaking down into small code pieces will make this practice more efficient and is also easier for community to conduct code review :-)

MicheleL-Intel: Sounds appropriate :) 👍

PR #664, opened 2019-10-30
replace spin lock with mutex

dyang23:
The spin lock is a dead-loop waiting lock, incase
a realtime thread waiting for it, the cpu resource
won't be released and trigger a dead-lock, so use
mutext to replace it.

Change-Id: I572effc8fa79df1a1df8bc44c9e70bf86e23acd3
Tracked-On: None
Signed-off-by: zhang jun <jun.zhang@intel.com>

### DIFF
--- a/Android.common.mk
+++ b/Android.common.mk
@@ -123,7 +123,8 @@ LOCAL_CPPFLAGS += \
 	-Wno-unused-private-field \
 	-Wno-unused-function \
 	-Wno-unused-parameter \
-	-Wno-unused-variable
+	-Wno-unused-variable \
+	-DUSE_MUTEX
 
 LOCAL_CPPFLAGS += -DVA_SUPPORT_COLOR_RANGE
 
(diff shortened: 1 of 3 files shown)


dyang23:
Please @Shao-Feng  @YuanjunHuang  help to review

tpalli:
Sometimes spin-lock is better option if we know that the loop will not spin for long time. Mutex has the additional CPU cost and might make the thing slower. It would be good to have perf runs before and after this change to see the effect.

dyang23:
> Sometimes spin-lock is better option if we know that the loop will not spin for long time (or if we call this very often). Mutex has the additional CPU cost and might make the thing slower. It would be good to have perf runs before and after this change to see the effect.

This fix is not from performance perspect, but it's trying fix a deadlock.
On android the display service will run as runtime process, which mean it will never be scheduled out if it's status is runnable. Another thread whose priority is normal and got the spin lock is waiting a mutex, the thread who hold the mutex status is runable, but it run on the same CPU as display service, that is why the deadlock happened.

Usually, spin lock used in kernel, and to protect the race-condition between CPUs rather than process.

tpalli:
@dyang23 Thank you for the detailed explanation, it makes sense. I have no objections for integrating this change, I just wanted to point out the possible perf cost. :)

PR #666, opened 2019-11-04
Fix screen flicker in Mosaic mode

yuanzhel:
Add seperate logic for Mosaic mode OverlayLayer Validate.

Change-Id: I4cc3acea6f35d67601a3076dd17de6eb2fd3ff98
Tracked-On: https://jira.devtools.intel.com/browse/OAM-85023
Signed-off-by: Yuanzhe Liu <yuanzhe.liu@intel.com>

### DIFF
--- a/common/core/overlaylayer.cpp
+++ b/common/core/overlaylayer.cpp
@@ -571,7 +571,7 @@ void OverlayLayer::ValidatePreviousFrameState(OverlayLayer* rhs,
 
   if (!layer->HasVisibleRegionChanged() && !content_changed &&
       surface_damage_.empty() && !layer->HasLayerContentChanged() &&
-      !(state_ & kNeedsReValidation)) {
+      !(state_ & kNeedsReValidation) && !layer->GetUseForMosaic()) {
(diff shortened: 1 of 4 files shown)


common/core/overlaylayer.cpp
```diff
@@ -571,7 +571,7 @@ void OverlayLayer::ValidatePreviousFrameState(OverlayLayer* rhs,
 
   if (!layer->HasVisibleRegionChanged() && !content_changed &&
       surface_damage_.empty() && !layer->HasLayerContentChanged() &&
-      !(state_ & kNeedsReValidation)) {
+      !(state_ & kNeedsReValidation) && !layer->GetUseForMosaic()) {
```
YuanjunHuang: Still wondering why this issue is related to mosaic display?

Shao-Feng: The property "total_display_" is only set in mosaic mode from ‘common/core/mosaicdisplay.cpp’. with the original 'if (total_displays_ == 1) {", the layer status on first display will not be set correctly, and the surface-damage will not be transferred.

PR #671, opened 2019-12-03
Set all layer as Client on CIV

Shao-Feng:
The buffer of Antutu6.1.1-3D can not be scaned-out directly.
WA: set all layer as Client on KVMgt

Change-Id: Ib7e4c20a1e05cf9dcc8e690373c20eac0c7ab042
Tests: UI normal
Tracked-On: None


Shao-Feng:
Just a WA for running antutu. will check mesa on host for the scanout buffer

PR #674, opened 2019-12-06
rename lib name to hwcomposer.broxton

swimmingcarp:
Tracked-On: OAM-88753
Signed-off-by: yifang ma <yifangx.ma@intel.com>

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -13,10 +13,10 @@
 // limitations under the License.
 
 // =====================
-// libdrmhwc_utils.a
+// libdrmhwc_utils_pri.a
 // =====================
 cc_library_static {
-    name: "libdrmhwc_utils",
+    name: "libdrmhwc_utils_pri",
 
     srcs: ["worker.cpp"],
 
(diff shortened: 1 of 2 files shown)


tests/Android.bp
```diff
@@ -1,7 +1,7 @@
 
 
 cc_test {
```
feijiang1: why change to a totally different name?

Android.bp
```diff
@@ -13,10 +13,10 @@
 // limitations under the License.
 
 // =====================
-// libdrmhwc_utils.a
+// libdrmhwc_utils_broxton.a
```
feijiang1: libdrmhwc_utils_broxton.a => libdrmhwc_utils_pri.a

tpalli:
Is the plan to only support Broxton platform? If not, then maybe use a more generic name, typically Android HAL modules use $(TARGET_BOARD_PLATFORM) in their name.

feijiang1:
We are planning to change hwc library name to libdrm.intel.so, due to such library can be applied to multiple platforms.

PR #676, opened 2019-12-23
add libva support

activebug202:
Signed-off-by: bozhengx <box.zheng@intel.com>


drmhwcomposer.h
```diff
@@ -129,6 +140,8 @@ enum class DrmHwcBlending : int32_t {
 
 struct DrmHwcLayer {
   buffer_handle_t sf_handle = NULL;
+  buffer_handle_t sf_va_handle = NULL;
```
feijiang1: don't need add

drmhwcomposer.h
```diff
@@ -129,6 +140,8 @@ enum class DrmHwcBlending : int32_t {
...
+  std::map<uint32_t, DrmHwcLayer *, std::greater<int>> va_z_map;
```
feijiang1: not easy to understand

drmhwcomposer.h
```diff
@@ -137,17 +150,41 @@ struct DrmHwcLayer {
...
+  }
+  std::map<uint32_t, DrmHwcLayer *, std::greater<int>> getVaLayerMapData(){
+    return va_z_map;
+  }
+  void SetVaLayerData(buffer_handle_t handle){
+    sf_va_handle = handle;
+  }
+  buffer_handle_t get_valayer_handle() const {
+    return sf_va_handle;
+  }
```
feijiang1: need delete

YuanjunHuang:
What is this PR for? what is the difference of va layer with original video layer?

platformminigbm.cpp
```diff
@@ -67,19 +91,62 @@ int DrmMinigbmImporter::Init() {
   return 0;
 }
 
-int DrmMinigbmImporter::ImportBuffer(buffer_handle_t handle, hwc_drm_bo_t *bo) {
+#ifdef ENABLE_DUMP_YUV_DATA
```
feijiang1: need use property to enable yuv dump

platformminigbm.cpp
```diff
@@ -67,19 +91,62 @@ int DrmMinigbmImporter::Init() {
...
+#define DUMP_DATA_FILE "/data/temp"
```
feijiang1: need add width/height/format info in the dump file

drmhwctwo.cpp
```diff
@@ -727,50 +746,82 @@ HWC2::Error DrmHwcTwo::HwcDisplay::SetVsyncEnabled(int32_t enabled) {
   return HWC2::Error::None;
 }
 
+//check whether the layer with the lease two zorder is the video layer
+bool DrmHwcTwo::HwcDisplay::WhetherVideoLayer(std::map< uint32_t,
```
Shao-Feng: Is better to name as "IsXXXX" or "ContainXXX" for the method which return bool

activebug202: > 
> 
> Is better to name as "IsXXXX" or "ContainXXX" for the method which return bool

Get it.I'll change.

feijiang1:
two commits are submitted, confused, please submit a new one

PR #677, opened 2019-12-25
Need to check if plane is reserved before setting All Cilent for KVM

Shao-Feng:
When "Plane Reserve" is not used. No need to set all layer as Client.

Change-Id: I1e517c73c73cf64823bfd04040e1f27cdc4f0726
Tests: None
Tracked-On: None

### DIFF
--- a/os/android/iahwc2.cpp
+++ b/os/android/iahwc2.cpp
@@ -1024,7 +1024,7 @@ HWC2::Error IAHWC2::HwcDisplay::ValidateDisplay(uint32_t *num_types,
       if (!avail_planes--)
         break;
 #ifdef KVM_HWC_PROPERTY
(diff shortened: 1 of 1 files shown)


os/android/iahwc2.cpp
```diff
@@ -978,6 +978,19 @@ HWC2::Error IAHWC2::HwcDisplay::ValidateDisplay(uint32_t *num_types,
 #ifdef FORCE_ALL_DEVICE_TYPE
   force_all_device_type = true;
 #endif
+
```
YuanjunHuang: pls add some comments here why this w/a code piece is needed otherwise it looks weird.

Shao-Feng: Updated

PR #681, opened 2020-01-09
Add systrace log for func-call

Shao-Feng:


Shao-Feng:
just for debugging, do not merge

PR #683, opened 2020-01-10
rename library name to avoid name conflicts from external

swimmingcarp:
1. rename to xxx.intel in Android.bp
2. support DRM_FORMAT_NV12 in hwc

Signed-off-by: yifang ma <yifangx.ma@intel.com>

### DIFF
--- a/platformminigbm.cpp
+++ b/platformminigbm.cpp
@@ -17,6 +17,7 @@
 #define LOG_TAG "hwc-platform-drm-minigbm"
 
 #include "platformminigbm.h"
+#include "i915_private_types.h"
 #include "drmdevice.h"
 #include "platform.h"
 
@@ -72,6 +73,10 @@ int DrmMinigbmImporter::ImportBuffer(buffer_handle_t handle, hwc_drm_bo_t *bo) {
   if (!gr_handle)
     return -EINVAL;
 
+  bool need_modifier = false;
+  if (gr_handle->format == DRM_FORMAT_NV12_Y_TILED_INTEL)
+    need_modifier = true;
+
   uint32_t gem_handle;
   int ret = drmPrimeFDToHandle(drm_->fd(), gr_handle->fds[0], &gem_handle);
   if (ret) {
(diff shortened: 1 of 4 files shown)


platformminigbm.cpp
```diff
@@ -83,20 +88,42 @@ int DrmMinigbmImporter::ImportBuffer(buffer_handle_t handle, hwc_drm_bo_t *bo) {
   bo->width = gr_handle->width;
   bo->height = gr_handle->height;
   bo->hal_format = gr_handle->droid_format;
-  bo->format = gr_handle->format;
+  if (need_modifier)
+    bo->format = DRM_FORMAT_NV12;
+  else
+    bo->format = gr_handle->format;
```
feijiang1: the code need move down together with 
if (need_modifier == false) {

platformminigbm.cpp
```diff
@@ -83,20 +88,42 @@ int DrmMinigbmImporter::ImportBuffer(buffer_handle_t handle, hwc_drm_bo_t *bo) {
...
+    bo->pitches[i] = gr_handle->strides[i];
+    bo->offsets[i] = gr_handle->offsets[i];
+    bo->gem_handles[i] = gem_handle;
   }
 
+  if (need_modifier == false) {
+    ret = drmModeAddFB2(drm_->fd(), bo->width, bo->height, bo->format,
+                        bo->gem_handles, bo->pitches, bo->offsets, &bo->fb_id, 0);
+  }
+  else {
```
feijiang1: } else need place together

platformminigbm.cpp
```diff
@@ -83,20 +88,42 @@ int DrmMinigbmImporter::ImportBuffer(buffer_handle_t handle, hwc_drm_bo_t *bo) {
...
+    uint32_t flag = DRM_MODE_FB_MODIFIERS;
+    uint64_t modifiers[HWC_DRM_BO_MAX_PLANES];
+    uint32_t numPlanes = gr_handle->base.numFds;
+
+    for (uint32_t i = 0; i < numPlanes; i++)
+      modifiers[i] = I915_FORMAT_MOD_Y_TILED;
```
feijiang1: need add format check, if format is DRM_FORMAT_NV12_Y_TILED_INTEL, then modifiers set to I915_FORMAT_MOD_Y_TILED.
And we only support DRM_FORMAT_NV12_Y_TILED_INTEL, can add the assert

platformminigbm.cpp
swimmingcarp: Done.

platformminigbm.cpp
swimmingcarp: Done.

platformminigbm.cpp
swimmingcarp: Done.

PR #684, opened 2020-01-13
Force to set 4K display as 1080P

Shao-Feng:
For KVM CVTE performance issue on 4K display.
Force to set the display config of 4K as 1080P

Tests: Work on KBL-NUC.
Tracked-On: None

### DIFF
--- a/wsi/drm/drmdisplay.cpp
+++ b/wsi/drm/drmdisplay.cpp
@@ -364,6 +364,24 @@ uint32_t DrmDisplay::FindPreferedDisplayMode(size_t modes_size) {
   return prefer_display_mode;
 }
 
+#ifdef KVM_HWC_PROPERTY
+uint32_t DrmDisplay::Find1080DisplayMode(size_t modes_size) {
+  uint32_t display_mode_1080p = 0;
+  if (modes_size >= prefer_display_mode_) {
+    int32_t p_width = 0, p_height = 0;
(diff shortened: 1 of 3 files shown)


wsi/drm/drmdisplay.cpp
```diff
@@ -364,6 +364,24 @@ uint32_t DrmDisplay::FindPreferedDisplayMode(size_t modes_size) {
...
 
+#ifdef KVM_HWC_PROPERTY
+uint32_t DrmDisplay::Find1080DisplayMode(size_t modes_size) {
+  uint32_t display_mode_1080p = 0;
+  if (modes_size >= prefer_display_mode_) {
+    int32_t p_width = 0, p_height = 0;
+    for (size_t i = 0; i < modes_size; i++) {
+      GetDisplayAttribute(i, HWCDisplayAttribute::kWidth, &p_width);
+      GetDisplayAttribute(i, HWCDisplayAttribute::kHeight, &p_height);
+      if (p_width <= 1920 && p_height <= 1080) {
```
tpalli: Shouldn't this rather be '==' than '<='? Otherwise it seems this loop would choose as example 640x480 happily as display_mode_1080p?

Shao-Feng: Good point. Thanks to review this. I will update soon.

PR #688, opened 2020-03-03
Remove KVM property setting for full planes

Shao-Feng:
Tracked-On: None


Shao-Feng:
Just for testing

PR #697, opened 2020-06-03
Use Kernel Param to judge if running with GVT-g

Shao-Feng:
Use the new kernel param to judge if celadon is running with
GVT-g, or GVT-d/native.
The KVM property is not needed any more. and the related method
is removed too

Change-Id: I935b0d97b8dd94ddc7485b9ef9221069a71eefb4
Tests: Work well on both GVT-d and GVT-g
Tracked-On: None

### DIFF
--- a/common/core/gpudevice.cpp
+++ b/common/core/gpudevice.cpp
@@ -16,6 +16,7 @@
 
 #include <gpudevice.h>
 
+#include <intel/intel_gvt.h>
 #include <sys/file.h>
 
 #include "mosaicdisplay.h"
@@ -67,6 +68,7 @@ bool GpuDevice::Initialize() {
   display_manager_->InitializeDisplayResources();
   display_manager_->StartHotPlugMonitor();
 
+  CheckGvtActive();
   HandleHWCSettings();
 
   if (reserve_plane_) {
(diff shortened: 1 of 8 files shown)


public/gpudevice.h
```diff
@@ -117,6 +119,8 @@ class GpuDevice : public HWCThread {
  private:
   GpuDevice();
 
+  void InitGvtActive();
```
YuanjunHuang, line 122: Suggest to name it to CheckGvtActive()

common/core/gpudevice.cpp
```diff
@@ -824,19 +830,25 @@ void GpuDevice::InitializeFloatDisplay(
   }
 }
 
+void GpuDevice::InitGvtActive() {
+  int fd = GetFD();
+  int gvt_active = 0;
+  drm_intel_get_gvt_active(fd, &gvt_active);
+  if (gvt_active) {
+    gvt_active_ = true;
+    ITRACE("Running on GVT");
```
YuanjunHuang, line 839: suggest change to "Running on vGPU (GVT-g) mode"

public/gpudevice.h
Shao-Feng: updated

common/core/gpudevice.cpp
Shao-Feng: Updated

PR #711, opened 2020-12-26
WA for Allocater & Mapper 4.0

renchenglei:
HWC has some dependence for gralloc1, this changes is
to help add some parameter from latest gralloc4 and one
wa for null pointer on GVT-d

Tracked-On: OAM-95365
Signed-off-by: Ren Chenglei <chenglei.ren@intel.com>

### DIFF
--- a/common/Android.mk
+++ b/common/Android.mk
@@ -81,6 +81,8 @@ LOCAL_SRC_FILES := \
         utils/hwcutils.cpp \
         utils/disjoint_layers.cpp
 
+LOCAL_CPPFLAGS += -DUSE_GRALLOC1
+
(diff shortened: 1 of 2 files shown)


Shao-Feng:
This one will make all VA and composing related features dismissed. let's just keep it for the special customer.

renchenglei:
Thanks @Shao-Feng, I will try to figure out the crash issues, and get it fix from minigbm side.

Shao-Feng:
This MACRO is already defined in Android.common.mk, and only be checked in os/android/gralloc1bufferhandler
I suppose it is already enough. why this macro must be defined in wsi and common service?

PR #712, opened 2021-01-20
Revert "support Iris driver"

renchenglei:
This reverts commit 9cc7cdc490349db1fa7abd83c85223bcfe5e8cc1.

For CIV, we need keep both i965 driver and iris driver.
i965 driver is used for KBL/CML, iris driver is used for
TGL, and the two driver is built into one image. We can't
check "BOARD_GPU_DRIVERS" flag to disable EXPLICIT_SYNC,
as it will cause PnP regression on CML.

Tracked-On: OAM-95784
Signed-off-by: Ren Chenglei <chenglei.ren@intel.com>

### DIFF
--- a/Android.common.mk
+++ b/Android.common.mk
@@ -141,19 +141,8 @@ LOCAL_C_INCLUDES += \
 	$(LOCAL_PATH)/common/compositor/vk \
 	$(LOCAL_PATH)/../mesa/include
 else
-
-# iris driver flags
-ifneq ($(filter iris, $(BOARD_GPU_DRIVERS)),)
-$(warning "Iris driver not fully supported, instability or lower performance may occur!")
-LOCAL_CPPFLAGS += \
-	-DUSE_GL \
-	-DDISABLE_EXPLICIT_SYNC
-else
-# i965 driver flags
 LOCAL_CPPFLAGS += \
 	-DUSE_GL
-endif
-
 endif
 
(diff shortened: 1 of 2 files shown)


renchenglei:
Hi @tpalli, mesa is upgrade to 20.2.3 in CIV, do we still need disable EXPLICIT_SYNC for iris?

renchenglei:
I tried to flash CIV on TGL RVP, Android could boot successfully.
@Shao-Feng & @feijiang1 & @YuanjunHuang 
Could you help take a review of this changes, thanks a lot!

feijiang1 (approved):

Shao-Feng:
Does @tpalli agree with this revert?

feijiang1:
> Does @tpalli agree with this revert?

feijiang1 (approved):

renchenglei:
Seems that Tapani is still on a vacation, could we get this merged, and if iris still has issue to support EXPLICIT_SYNC, we can check with device id, instead of BOARD_GPU_DRIVERS flags.

tpalli:
> Seems that Tapani is still on a vacation, could we get this merged, and if iris still has issue to support EXPLICIT_SYNC, we can check with device id, instead of BOARD_GPU_DRIVERS flags.

I'm not on vacation, sorry I just did not spot this until now :) This is OK with me, iris should be working fine (including explicit sync) on Android now.

PR #714, opened 2021-03-14
Add trace for regisering vsync call

Shao-Feng:
Change-Id: I509ec86a36630ee70f917a9b0f4d7908b330a4d2

### DIFF
--- a/common/display/vblankeventhandler.h
+++ b/common/display/vblankeventhandler.h
@@ -27,8 +27,6 @@
 
 #include "hwcthread.h"
 
-#define VBLANK_THRESHHOLD 2777777
-
 namespace hwcomposer {
 
 class DisplayQueue;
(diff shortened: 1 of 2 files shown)


common/display/vblankeventhandler.h
```diff
@@ -27,8 +27,6 @@
 
 #include "hwcthread.h"
 
-#define VBLANK_THRESHHOLD 2777777
```
YuanjunHuang, line 30: what does this threshold mean here?

Shao-Feng: We ever think only when the different in vperiod is larger than this one, the 2.4 callback should be invoked.
For instance, when the period is changed from 90hz to 60hz

os/android/iahwc2.cpp
```diff
@@ -344,6 +344,7 @@ HWC2::Error IAHWC2::RegisterCallback(int32_t descriptor,
       for (size_t i = 0; i < size; ++i) {
         IAHWC2::HwcDisplay *display = extended_displays_.at(i).get();
         display->RegisterVsyncCallback(data, function);
+        ETRACE("######Register original Vsync callback");
```
YuanjunHuang, line 347: If this aims to check in, do not use etrace for log only info.

Shao-Feng: just for verifying, I will commit another one once this one work

PR #717, opened 2021-04-02
Fixed the fence fd leak in SetLayerBuffer.

shuangwan01:
Fence fd needs to be closed if it's not going to be used in HWC
composition. This is typically for layer types of Client, Sideband,
SolidColor and layer with NULL buffer attached.

Change-Id: Icd5d5061011a16bae9f5b1062e7aadeedfff3efa
Tracked-On: OAM-96402


Shao-Feng (approved):
Nice

